PR #56, opened 2021-03-22
Add Feature in user search old frontend because Support has requested the feature long ago

einhornimmond:
## 🍰 Pullrequest
Feature Request: Possibility to search for users with not activated accounts, order by created date. 
Changes: Add select under search in user-search to choose account state: all or account not activated
add_account_state

Update Community Server and Login Server to use new parameter in search query
Update mithriljs user search modul to able to sort user by name, email, balance, public key and created ASC and DESC. 
Shorten oversize emails and give them the action to show to full size email in modal dialog if clicked on it. 

Seems to be compatible with stage1 and stage0


ulfgebhardt:
@einhornimmond You need to rebase this to get the latests tests. Furthermore do you want to merge it into master only or also in stage0 ?

einhornimmond:
Need only merged into master. 
I don't think it makes sense to merge it into stage0.
I think we can now delete stage0 branch (after tagging), this new code based on stage1 code is now running in production without errors.

ulfgebhardt (approved):

PR #63, opened 2021-03-22
Update doc

einhornimmond:
## 🍰 Pullrequest
add docu for update user data interface.


login_server/src/cpp/JSONInterface/JsonGetLogin.cpp
```diff
@@ -71,11 +71,15 @@ Poco::JSON::Object* JsonGetLogin::handle(Poco::Dynamic::Var params)
 		em->addError(new Error("JsonGetLogin::handle", "generic exception calling userModel->getJson: "));
 		em->sendErrorsAsEmail();
 	}
+	// use both variants for compatibility reasons, but final version is Transactions.pending
+	result->set("Transaction.pending", session->getProcessingTransactionCount());
 	result->set("Transactions.pending", session->getProcessingTransactionCount());
```
ulfgebhardt, line 76: Why duplicate here with `s` and without?

login_server/src/cpp/model/table/EmailOptIn.h
```diff
@@ -48,6 +48,7 @@ namespace model {
 			Poco::Data::Statement _loadFromDB(Poco::Data::Session session, const std::string& fieldName);
 			Poco::Data::Statement _loadIdFromDB(Poco::Data::Session session);
 			Poco::Data::Statement _loadMultipleFromDB(Poco::Data::Session session, const std::string& fieldName);
+			Poco::Data::Statement _loadMultipleFromDB(Poco::Data::Session session, const std::vector<std::string> fieldNames, MysqlConditionType conditionType = MYSQL_CONDITION_AND);
```
ulfgebhardt, line 51: Same question: why have the same function defined twice? Is this overloaded?

login_server/src/cpp/JSONInterface/JsonGetLogin.cpp
einhornimmond: Because I wasn't sure if I have fixed every occurance in community-server, but now I have fixed it

einhornimmond (commented):
Ulf: Same question: why have the same function defined twice? Is this overloaded?
```c++
Poco::Data::Statement _loadMultipleFromDB(Poco::Data::Session session, const std::string& fieldName);
Poco::Data::Statement _loadMultipleFromDB(Poco::Data::Session session, const std::vector<std::string> fieldNames, MysqlConditionType conditionType = MYSQL_CONDITION_AND);
````
Have you looked at the parameters? Of course there overloaded. 
The multiple in the name refers to the fact that this functions are you used for mysql calls if multiple results are expected.
The first one is used if one variable for where is used, the second one for more than one. 
I often use overloaded functions in c++.

ulfgebhardt (approved):

PR #82, opened 2021-03-24
Api password reset

einhornimmond:
Add sendEmail api call for use for reset passwords calls and also useable for admins to resend email verification codes to user


configs/login_server/grd_login.properties
```diff
@@ -22,6 +22,8 @@ loginServer.db.user = root
 loginServer.db.password = 
 loginServer.db.port = 3306
 
+frontend.checkEmailPath = http://localhost/account/checkEmail 
+
 email.disable = true
```
ulfgebhardt, line 27:
```suggestion
loginServer.db.port = 3306

frontend.checkEmailPath = http://localhost/account/checkEmail

email.disable = true

```

ulfgebhardt: I believe those characters are windows lineendings. Common standard is linux line endings on the remote.
image

ulfgebhardt: https://github.com/atom/atom/issues/3105

frontend/src/apis/loginAPI.js
```diff
@@ -7,12 +7,31 @@ const EMAIL_TYPE = {
...
+	
+    if(result.data.state !== 'success'){
+      throw new Error(result.data.msg)
+    }
+    return { success: true, result }
+  } catch(error){
+    return { success: false, result: error}
+  }
+}
+
```
ulfgebhardt, line 25: Please feel free to refactor this as needed. This is just quick and dirty from me.

frontend/src/apis/loginAPI.js
```diff
@@ -7,12 +7,31 @@ const EMAIL_TYPE = {
...
 const apiPost = async (url, payload) => {
   try {
     const result = await axios.post(url, payload);
     if(result.status !== 200){
       throw new Error('HTTP Status Error '+result.status)
     }
+	if(result.data.state === 'warning') {
+		return { success: true, result: error }
+	}
```
ulfgebhardt, line 34:
```suggestion
    const result = await axios.post(url, payload);
    if(result.status !== 200){
      throw new Error('HTTP Status Error '+result.status)
    }
	  if(result.data.state === 'warning') {
		  return { success: true, result: error }
	  }
```

login_server/src/cpp/model/email/Email.h
```diff
@@ -69,6 +70,8 @@ namespace model {
 		EmailType mType;
 
 		std::queue<Poco::Net::StringPartSource*> mAdditionalStringPartSrcs;
+		std::string		mCustomText;
+		std::string     mCustomSubject;
```
ulfgebhardt, line 74:
```suggestion
		std::string     mCustomText;
		std::string     mCustomSubject;
```

ulfgebhardt (changes requested):
Thank you for doing this - especially the API implementation directly in the frontend <3 - this is the next feature we will implement.

This PR contains some character issues which I nazi about :)
- on the frontend we use two-space distance instead of tabs
- there is windows lineendings in the file

I know developers love their editors they got use to, for example @Mogge uses some acient tech already available on analog typewriters. But consider vscode or better the open source variant without Microsoft spyware vscodium. It handles those character settings rather well-

image

ulfgebhardt (changes requested):
There is still Windows lineendings

image

See: https://github.com/gradido/gradido/pull/82/files

ulfgebhardt (approved):

PR #83, opened 2021-03-24
add additional docker container which is running skeema on startup to…

### LINKED ISSUE #76: 🚀 [Feature] Migrate Database (on docker)
## :rocket: Feature

Migrate Database (on docker)

- On docker the database is not updated
- We need to update tables and structures automatically
- Preferable with a rollback options (migration up & down)

einhornimmond:
… update dbs if they have changed, cleanup mariadb/Dockerfile

### Issues
- fixes #76

### Todo
if it should work also only with docker-compose up without build, the skeema folder must be mounted instead of copied to reflect current changes

### DIFF
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -57,6 +57,20 @@ services:
      - external-net
     volumes:
      - /sessions
+     
+  #########################################################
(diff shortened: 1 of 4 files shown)


ulfgebhardt (approved):
Awesome!

PR #84, opened 2021-03-25
Vue with nginx

einhornimmond:
## 🍰 Pullrequest
update vue config that vue reside in subfolder vue so it is accessable via nginx with http(s)://{url}/vue
(I wasn't able to put the community-server in subfolder instead)
I don't know if my change break something other in vue so please check first.


docker-compose.override.yml
```diff
@@ -60,4 +64,4 @@ services:
    
 volumes:
   frontend_node_modules:
-  conan:
\ No newline at end of file
+  conan:
```
ulfgebhardt, line 67:
```suggestion
  conan:
```

docker-compose.yml
```diff
@@ -135,4 +133,4 @@ networks:
     internal: true
 
 volumes:
-  db_vol:      
+  db_vol:
```
ulfgebhardt, line 136:
```suggestion
  db_vol:
```

frontend/vue.config.js
```diff
@@ -15,6 +15,7 @@ module.exports = {
     },
   },
   lintOnSave: true,
+  publicPath: '/vue/',
```
ulfgebhardt, line 18: this should be an env variable since this might change according to setup. In the long run it will be hosted on the main domain. Furthermore a sensitive default for development would be `/`

ulfgebhardt (changes requested):
In general I really like this change. I believe the `./vue/` path should be configurable and we should postpone the merge till after the successful video, since it could interrupt development (frontend accessible on a different domain)

PR #101, opened 2021-03-29
Optimize login server build

einhornimmond:
## 🍰 Pullrequest
Now login-server use protoc and his other dependencies from prebuild and to docker hub uploaded images.

### Issues
- Slow Build on Test
- Now the build-time on github is similar to the build-time from frontend

### DIFF
--- a/login_server/Dockerfile
+++ b/login_server/Dockerfile
@@ -1,19 +1,4 @@
 #########################################################################################################
-# Build protoc 
-#########################################################################################################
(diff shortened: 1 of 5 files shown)


ulfgebhardt:
This is awesome, how much time do we save on this?

Another suggestion: You upload the protoc thing to your private account rater then the gradido orga. Why is that? Should we change that?

ulfgebhardt (approved):

PR #103, opened 2021-03-29
Build on run

### LINKED ISSUE #79: 🔧 [Refactor] Loginserver is not recompiled after update
## :zap: Refactor ticket

Loginserver is not recompiled after update when run in docker
`docker-compose up` is not sufficient to make sure the latest version of the loginserver is compiled. `docker-compose up --build` does whats needed but takes 30mins. So we need to make sure the loginserver is only compiled (fully) when the code changed.

einhornimmond:
## 🍰 Pullrequest
Let login server build on docker-compose up rather than on build
to have it automaticly rebuild some parts if src files have changed by calling docker-compose up

### Issues
- fixes #79


ulfgebhardt (approved):
Lots of changes. I trust that it all works <3

PR #111, opened 2021-03-30
markdown of login server api

einhornimmond:
update mardown of login server api


ulfgebhardt (approved):
I do not get why you keep checking in gn submodule updates, but since we don't use it yet its fine with me ;)

Issue #131, opened 2021-04-01
🔧 [Refactor] Vue path configurable


ulfgebhardt:
## :zap: Refactor ticket

Make Vue path configurable (nginx) so we can easily switch between the old and the new frontend.
How can we make both frontends compatible?

ulfgebhardt:
This is not yet solved

PR #138, opened 2021-04-06
Fix compose bug

einhornimmond:
## 🍰 Pullrequest
in frontend 
depends_on:
  - nginx
  
leads to circular dependencie because nginx depens_on frontent. 
Removed it.

By the way add debug message to Login-Server after he starts visible also in docker. 

Make VUE_PATH default to /vue and copy back vue changes in nginx.conf so that on default composer-build
the new client is reachable with http://localhost/vue

### DIFF
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -43,7 +43,7 @@ services:
       - community-server
       - login-server
     volumes: 
-      - ./logs:/var/log/nginx 
+      - ./logs/nginx:/var/log/nginx 
   
   #########################################################
   ## PHPMYADMIN ###########################################
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -16,8 +16,6 @@ services:
       target: production
     networks:
       - external-net
-    depends_on:
-      - nginx
     ports:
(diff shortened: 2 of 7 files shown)


ulfgebhardt:
I do not understand why nginx depends on frontend. Same applies for the login server. Nginx is the webserver and should run when we register the services. If its working I will not object, but I believe the configuration is off regarding logic and standards.

ulfgebhardt:
Does this solve #131 ?

ulfgebhardt (approved):

ulfgebhardt:
We seem to need this to have the master composing again - therefore I consider to merge it and maybe make a followup issue

einhornimmond:
In nginx config I use 
- proxy_pass          http://login-server:1200;
- proxy_pass          http://login-server:1201;
- proxy_pass          http://frontend:8080;
- fastcgi_pass         community-server:9000;

It seems that before starting nginx check if the services are reachable and if not, don't start. 
At least that was what I experienced as I startet with nginx. 
So it is important that this services running before nginx can start. 
If you find a another way, do it.

PR #147, opened 2021-04-07
fix getBalance API call

ulfgebhardt:
## 🍰 Pullrequest

- fix getBalance API call
- make docker-compose debug setup listen to local changes of community server

This results in an invalid balance calculation:

image

It is unclear if the applied "fix" is causing this or if its an inherent problem somewhere else.


einhornimmond:
I have fixed this and other errors in branch backport_balance_recalculation 
have I forget to merge it?
The right version for line 89 is
```php 
if(!$last_state_user_transaction || $last_state_user_transaction->balance <= 0) {
```
If no state_user_transaction exist they must be recalculated, also $recalculate_state_user_transactions_balance = true
It is like a automatic update to able to use this new version of state balance calculation with state_user_transactions even on datasets where state_user_transactions not exist. And it is also for easy repairing code. If state_user_transactions contain wrong data, simply truncate table and it will be calculated on next call of updateBalances for the user.

einhornimmond:
Oh I have again take look and I'm wrong, you are right. 
if $last_state_user_transaction is null then $recalculate_state_user_transactions_balance = true don't make sense,
because it uses only existing entrys of state_user_transactions.
But if transactions present and no state_user_transactions, http://communit-server/transactions/synchronizeWithStateUserTransactions must be called. That was my update-function.

einhornimmond (approved):

ulfgebhardt:
This still leaves me with an incorrect balance count, even after my fix. Is this addressed in another PR? Is it safe to merge this? @einhornimmond

einhornimmond:
It seems I have forgotten the PR for it, but in current state2 code it is fixed. 
So should I take the time to make a PR for the backport_balance_recalculation and merge it with current master and test if all is working or should I better work proceed on stage2 where this things already fixed?

ulfgebhardt:
How long will it take to finish stage2?

einhornimmond:
I don't know. I must now test that my changes work and fix all bugs which occure.

ulfgebhardt:
Then make a separate PR pls

ulfgebhardt:
Can we close this in favor of https://github.com/gradido/gradido/pull/155 @einhornimmond ?

Issue #125, opened 2021-04-01
🚀 [Feature] Link to Member area


ulfgebhardt:
## :rocket: Feature

Have a link to the member area in elopage like on the current client

PR #152, opened 2021-04-07
fix-125-link-to-community

ogerly:
## 🍰 Pullrequest
in the footer of the app the link to the elopage community has been added

FireShot Capture 236 - Gradido Account - localhost

FireShot Capture 257 - Gradido Account - localhost

### Issues
#125

### DIFF
--- a/frontend/src/components/SidebarPlugin/SideBar.vue
+++ b/frontend/src/components/SidebarPlugin/SideBar.vue
@@ -40,11 +40,11 @@
               <span>Transactions</span>
             </router-link>
             <!--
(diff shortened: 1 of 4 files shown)


ulfgebhardt:
This fixes https://github.com/gradido/gradido/issues/125

Discussion:
- The new implementation has the Member area in the footer. Is that ok?
- The new implementation defines this as "Community" and not "Member Area". Is that ok?

einhornimmond:
I think it will be problem with the current users to find it again which changed name and place. 
I see many Support requests incoming.

ulfgebhardt (approved):

Issue #104, opened 2021-03-29
🚀 [Feature] Better URL structure of the Frontend


ulfgebhardt:
## :rocket: Feature

Better URL structure of the Frontend
- [X] URLs have # in them
- [X] Double Redirect is not caught and produce errors in console.log
- [ ] Figure out how we can properly integrate into nginx (in the future the frontend must be reachable under `/` now it must be reachable under `/vue/`

einhornimmond:
To make frontend reachable under / is easy, the real problem is to make community-server reachable not under /
I have already tryed everything which was suggested here:
https://stackoverflow.com/questions/2119736/cakephp-in-a-subdirectory-using-nginx-rewrite-rules

The only way I see is to load vue-frontend inside a cakephp side which I can set via routing to /
But last time I tryed to explain it to alex he didn't understand me. 
In past I had put vue-client inside community-server and edited package.js and public/index.html so that on build a cakephp suitable side in template/pages/frontend.ctp was generated which was loading the vue-frontend and was accessable via /
But I don't know how to make it that Alex can work with automatic rebuild how he is used to it with that configuration. 
I couldn't find a way for build vue with watch how I do it with mithril-js
(watchify ./src/app.js  -t [ babelify --presets [ @babel/preset-env ] ] -o app.js -v)

ulfgebhardt:
#35

Mogge:
See this: https://router.vuejs.org/guide/essentials/history-mode.html#nginx

ulfgebhardt:
@Mogge Whats with the remaining issues?

PR #167, opened 2021-04-12
fix: Remove Hashtag From Route

Mogge:
## 🍰 Pullrequest
 * setting `mode: 'history'` to avoid # in routes (see Todo)
 * added some missing locales (members area)
 * changed locale of members area
 * removed use of store in Send compoenent

### Issues
- relates #104

### Todo

- [ ] On page load there can be seen a 404 for a short time. After that, everything runs fine. I think it has to do with the nginx config (see https://router.vuejs.org/guide/essentials/history-mode.html#nginx)

### DIFF
--- a/frontend/src/views/Layout/DashboardLayout_gdd.spec.js
+++ b/frontend/src/views/Layout/DashboardLayout_gdd.spec.js
@@ -83,7 +83,7 @@ describe('DashboardLayoutGdd', () => {
       })
 
       it('has first item "send" in navbar', () => {
-        expect(navbar.findAll('ul > li').at(0).text()).toEqual('sent')
+        expect(navbar.findAll('ul > li').at(0).text()).toEqual('send')
       })
 
       it('has first item "send" linked to overview in navbar', () => {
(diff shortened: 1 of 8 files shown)


ulfgebhardt (commented):
I do not really understand how this addresses the hashtag thing? Mind explaining please?

Mogge:
We need to have the `mode: 'history'` in the router setup.

frontend/src/views/Layout/DashboardLayout_gdd.spec.js
```diff
@@ -83,7 +83,7 @@ describe('DashboardLayoutGdd', () => {
       })
 
       it('has first item "send" in navbar', () => {
-        expect(navbar.findAll('ul > li').at(0).text()).toEqual('sent')
+        expect(navbar.findAll('ul > li').at(0).text()).toEqual('send')
```
Mogge, line 86: How did this get into master?

ulfgebhardt: The Test is not yet required, fix it and we can make that happen

ulfgebhardt (approved):

PR #174, opened 2021-04-14
Community 14.04.21

einhornimmond:
## 🍰 Pullrequest

Let mysql fill received value for transactions instead of php
received is part of gradido protocol and with blockchain-type mysql
mysql should decide this things (sequence-number and receiving date)

### correct auto-increment basis for id in transactions
With blockchain type = db the mysql table transactions act mainly as ordering service,
the sequence number will be created by mysql auto-increment id.
But if an error occure in code and mysql rollback was called, the next call to transactions insert deliver
a higher id so a gap is created in blockchain.
To prevent this, I added code which set auto-increment to the count of actually transaction entrys in db after rollback.


einhornimmond:
#8a6acd67f7236e767a7a90ea6bbd497d6967bd6d 
Fixtures a used to automatic filling test db
Now I have added a function (inside a new base class for all fixtures) which translate mysql export format into fixture records format. 
So I can easly export test data from my db, change the ( to [ and ) to ] call the function and finish :)

einhornimmond:
#4690192a71f8b720d463fc7c29b6dd84d8145add
#### config/bootstrap.php:
check if $_SERVER['REQUEST_METHOD'] isset because in tests this is not set.
I have added 
```php 
if ($_SERVER['REQUEST_METHOD'] == 'OPTIONS') {
    exit(0);
}
```
because it was needed for frontend working. Maybe it wasn't the right location, you can move/update it if you find a better solution.

#### config/routes.php:
Whitelist empty clientIP for ajax calls because if test framework is called, clientIP is empty.

einhornimmond:
add my test data set to Fixtures for updateBalance test in next step

einhornimmond:
commited to many files, sorry

community_server/config/routes.php
```diff
@@ -65,7 +65,7 @@
             if($entry == 'ElopageWebhook' || $entry == 'AppRequests') {
               return true;
             }
-            if($request->clientIp() == '127.0.0.1' || $request->clientIp() == 'localhost') {
+            if($request->clientIp() == '127.0.0.1' || $request->clientIp() == 'localhost' || $request->clientIp() == '') {
               return true;
             }
```
ulfgebhardt, line 70: I do not understand this block. Handling localhost differently might cause problems since the test & development environment is on localhost - other setups might not.

ulfgebhardt (changes requested):
Please check my comment. Else it looks ok.

I see you work on test: I'd like to have them run on workflows so we can utilize them to test on every PR/Push.

Furthermore we need a way to seed data. But I will make an issue for that.

community_server/config/routes.php
einhornimmond: The function in which this code is located ($csrf->whitelistCallback)
check if csrf check can be disabled for this call. 
I use this to only allow ajax requests from localhost and from ips in config/app.php in API.allowedCaller array. 
When I running tests via Unit Tests (build-in in cakephp) clientIp return empty string instead of expected localhost or 127.0.0.1 so I have added this.

einhornimmond:
"Furthermore we need a way to seed data. But I will make an issue for that."
For cakephp this is done via the FixtureClasses in tests/Fixture folder.

community_server/config/routes.php
ulfgebhardt: 👍🏻

ulfgebhardt (approved):

PR #194, opened 2021-04-15
refactor: Redirects and Misuse of Children in Router

### LINKED ISSUE #35: 🐛 [Bug] Prevent Double Navigation
## :bug: Bugreport

image

We have a lot of double navigation since its not prevented by our vue router.

### Steps to reproduce the behavior
1. Login with wrong credentials
2. Get redirected to the same page
3. See error in console
4. ...
5. Profit

### Expected behavior

Double navigation should be prevented by the vue router or in the appropriate place

Mogge:
## 🍰 Pullrequest
refactoring of the router:
 * No` redirects anymore
 * no bad children anymore
 
Layout is chosen in App.vue.

### Issues
- fixes #35 
- relates #104


frontend/src/views/Layout/DashboardLayout_gdd.spec.js
```diff
@@ -97,7 +97,9 @@ describe('DashboardLayoutGdd', () => {
 
       // to do: get this working!
       it.skip('has second item "transactions" linked to transactions in navbar', async () => {
+        //console.log(wrapper.html())
         navbar.findAll('ul > li > a').at(1).trigger('click')
+        //console.log(wrapper.html())
```
ulfgebhardt, line 102:
```suggestion
```

frontend/src/views/Layout/DashboardLayout_gdd.spec.js
```diff
@@ -97,7 +97,9 @@ describe('DashboardLayoutGdd', () => {
 
       // to do: get this working!
       it.skip('has second item "transactions" linked to transactions in navbar', async () => {
+        //console.log(wrapper.html())
```
ulfgebhardt, line 100:
```suggestion
```

frontend/src/App.vue
```diff
@@ -17,7 +17,7 @@
     </header>
     <div class="">
       <particles-bg type="custom" :config="config" :bg="true" />
-      <router-view />
+      <component :is="$store.state.session_id ? 'DashboardLayout' : 'AuthLayoutGDD'" />
```
ulfgebhardt, line 20: Is this new structure really an improvement? This limits the layout to two and decouples it from the router.
Can we do something like this?

```
import myLayout1 from [...]

[...]

  {
    path: '/profile',
    component: () => import('../views/Pages/UserProfileCard.vue'),
    layout: myLayout1
  },
```

ulfgebhardt (changes requested):

frontend/src/App.vue
Mogge: This is nuxt. I did not find a way to do this with only vue-router so far. But I agree, this would be the best solution.

Mogge: @ogerly  achieved this by using faked children. But we do not have children.

Mogge: We could use a computed property with a switch to implement more than two layouts

ulfgebhardt: So merge as you suggested or refactor in some way? nuxt?

ulfgebhardt (approved):

PR #197, opened 2021-04-16
update code which produce error which last fix (._.);

einhornimmond:
### Issues

- None

### Todo

- [X] None

### DIFF
--- a/login_server/src/cpp/SingletonManager/PendingTasksManager.cpp
+++ b/login_server/src/cpp/SingletonManager/PendingTasksManager.cpp
@@ -193,7 +193,7 @@ void PendingTasksManager::checkForFinishedTasks(Poco::Timer& timer)
 				bool removeIt = false;
 				if (!json.isNull()) {
 					auto state = json->get("state");
-					if (!state.isEmpty() && json->get("state").toString() == "success") {
+					if (!state.isEmpty() && state.toString() == "success") {
 						removeIt = true;
 					}
 				}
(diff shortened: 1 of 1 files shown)


einhornimmond:
Added back a field in gradido protocol what I had removed but which I need for db version and fo multiple blockchain version.

login_server/src/cpp/SingletonManager/PendingTasksManager.cpp
```diff
@@ -192,7 +192,8 @@ void PendingTasksManager::checkForFinishedTasks(Poco::Timer& timer)
 				auto json = transaction->getModel()->getResultJson();
 				bool removeIt = false;
 				if (!json.isNull()) {
-					if (json->get("state").toString() == "success") {
+					auto state = json->get("state");
+					if (!state.isEmpty() && json->get("state").toString() == "success") {
```
ulfgebhardt, line 196:
```suggestion
					if (!state.isEmpty() && state.toString() == "success") {
```

Why do you convert the json to the state variable just to repeat that?

ulfgebhardt (changes requested):

login_server/src/cpp/SingletonManager/PendingTasksManager.cpp
einhornimmond: you are right

ulfgebhardt (approved):

PR #202, opened 2021-04-19
refactor: Remove Cookies

Mogge:
## 🍰 Pullrequest
Use `vuex-persistedstate` to save the store in a window session storage.
 * Login: API call from component. Store only called to commit changes to state
 * Same for Logout and Register
 * Introduced `meta: { requiresAuth: true, }` to the routes when authentication is required
 * requiresAuth is checked in `router.js` with a navigation guard
 * requiresAuth in routes switches layout in App.js

### Todo
Please test! I noticed some buggy behavior on logout. This seems to be a conflict of the two logout buttons we have. I hope it is fixed when we switch to just one menu.


frontend/src/store/store.js
```diff
@@ -1,18 +1,22 @@
...
+  plugins: [
+    createPersistedState({
+      storage: window.sessionStorage,
+    }),
+  ],
   state: {
     session_id: null,
     email: '',
     language: 'en',
     sizeDE: 'normal',
```
Mogge, line 18: @ogerly What is this for? This is only in the store. Can we remove this?

frontend/src/store/store.js
```diff
@@ -31,9 +35,7 @@ export const store = new Vuex.Store({
   // Syncronous mutation of the state
   mutations: {
     language: (state, language) => {
-      //console.log('mutation: language', language)
       state.language = language
-      $cookies.set('gdd_lang', language)
       if (state.language == 'de') {
         state.sizeDE = 'big'
```
Mogge, line 40: Why?

frontend/src/store/store.js
```diff
@@ -42,87 +44,44 @@ export const store = new Vuex.Store({
...
-        // Register failed, we perform a logout
-        // console.log('action createUser to  logout start')
-        dispatch('logout')
-      }
+      commit('session_id', data.session_id)
+      commit('email', data.email)
     },
     logout: async ({ commit, state }) => {
-      //console.log('action: logout')
       if (state.session_id) {
```
Mogge, line 75: Make this call in the component?

frontend/src/views/KontoOverview.vue
```diff
@@ -40,7 +40,7 @@ export default {
     GddTable,
   },
   created() {
-    this.$store.dispatch('accountBalance', $cookies.get('gdd_session_id'))
+    this.$store.dispatch('accountBalance', this.$store.state.session_id)
```
Mogge, line 43: This should be a data. I will fix this later

frontend/src/views/Layout/DashboardNavbar.vue
```diff
@@ -116,7 +116,7 @@ export default {
       this.activeNotifications = false
     },
     logout() {
-      //console.log("DashboardNavbar.vue user logout() : ")
+      this.$router.push('/login')
```
Mogge, line 119: I am not sure about the order here. Is the dispatch executed after redirect?

frontend/src/views/Pages/Register.vue
```diff
@@ -167,19 +169,28 @@ export default {
...
+        })
+        this.model.email = ''
+        this.model.firstname = ''
+        this.model.lastname = ''
+        this.password = ''
+        this.$router.push('/thx')
+      } else {
+        // todo: Display a proper error message!
+        this.$router.push('/login')
+        this.$store.dispatch('logout')
```
Mogge, line 192: Order?

ulfgebhardt: I would turn it around actually

frontend/src/store/store.js
Mogge: I leave this as it is by now. When we switch to one menu I will move the call to the component

ulfgebhardt (approved):
I'd like to have issues for the remaining issues like the language thing and the comments you made

PR #204, opened 2021-04-20
refactor: Remove Right Menu

### LINKED ISSUE #179: 🔧 [Refactor] Menu
## :zap: Refactor ticket

We had a discussion about the menu. Currently we have two menus. There is two options:
- Have one menu
- Have two menus with distinctively different tasks

The common consensus was: Two menus, one for banking, one for the user. I disagree tho - I believe one menu is sufficient at the given time, its less work, easier to communicate to the user.

We could not reach a final decision. If I have to decide, I would go with 1 menu as stated above and I will outline it here. Please make a suggestion for the two menu option (@Mogge (?))

Furthermore I'd like to have symbols for every menu entry using fontawesome (or something else) as icon library. The corresponding icons are denoted behind the Menu. The English translation has to be found according to the German words.

Menu:
```
- Überweisen (Statt Senden) (money-bill-wave)
- Umsätze (statt Transactions) (wallet)
- Schöpfen (statt Aktivitäten) (hand-holding-medical)
----
- Stammdaten (Statt Mein Profil & Settings) (address-card)
----
- Ausloggen (sign-out-alt)
```

ogerly:
## 🍰 Pullrequest
rechtes menü habe ich entfernt

### Issues
- fixes #179

### DIFF
--- a/frontend/src/routes/routes.js
+++ b/frontend/src/routes/routes.js
@@ -10,6 +10,7 @@ const routes = [
   },
   {
     path: '/profile',
+    name: 'Profile',
     component: () => import('../views/Pages/UserProfileCard.vue'),
   },
   {
--- a/frontend/src/views/Layout/DashboardNavbar.vue
+++ b/frontend/src/views/Layout/DashboardNavbar.vue
@@ -70,10 +70,5 @@ export default {
       this.$router.push('/login')
     },
   },
-  beforeDestroy() {
-    if (this.$sidebar.showSidebar) {
-      this.$sidebar.showSidebar = false
-    }
-  },
 }
 </script>
--- a/frontend/src/views/Layout/DashboardNavbar.vue
+++ b/frontend/src/views/Layout/DashboardNavbar.vue
@@ -4,83 +4,33 @@
     class="navbar-top navbar-expand"
     :class="{ 'navbar-dark': type === 'default' }"
   >
-    <a
-      href="#"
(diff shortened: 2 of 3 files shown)


frontend/src/routes/routes.js
```diff
@@ -10,6 +10,7 @@ const routes = [
   },
   {
     path: '/profile',
+    name: 'Profile',
```
ulfgebhardt, line 13: @Mogge do we use this? @ogerly How is this related to the PR?

frontend/src/views/Layout/DashboardNavbar.vue
```diff
@@ -120,5 +70,10 @@ export default {
       this.$store.dispatch('logout')
     },
   },
+  beforeDestroy() {
+    if (this.$sidebar.showSidebar) {
+      this.$sidebar.showSidebar = false
+    }
```
ulfgebhardt, line 76: Isn't this solved by the two different layouts? Do we need those lines? What do they do?

ulfgebhardt (commented):
Two questions. Also I do not understand why two menus are effected. The one in `Sidebar.vue` and the one in `DashboardNavbar.vue`

frontend/src/routes/routes.js
Mogge: No, so far we do not have named views.

Mogge (approved):

PR #212, opened 2021-04-20
feat: Reset Password

Mogge:
## 🍰 Pullrequest
reset the password on the server.

**PROBLEMS**
* it looks like you can receive a session_id via loginViaEmailVerificationCode only once per code. Is this on purpose?
* We do not know the email at this point in the form.
  - Either the server sends also the email after successful login via verification code (better user experience)
  -  or the user must type in the email again and we need an email field on password reset
  -  or, the server finds the email by the verification code
  -  or, we include the email in the link


frontend/src/views/Pages/ResetPassword.vue
```diff
@@ -139,8 +142,15 @@ export default {
...
-  created() {
-    //console.log('resetpage', this.$route)
+  async created() {
+    const optin = this.$route.params.optin
+    const result = await loginAPI.loginViaEmailVerificationCode(optin)
+    if (result.success) {
+      this.authenticated = true
+      this.session_id = result.result.data.session_id
+    } else {
+      alert(result.result.message)
```
Mogge, line 152: We need a proper error handling here!

frontend/src/apis/loginAPI.js
```diff
@@ -64,14 +64,28 @@ const loginAPI = {
...
   },
+  loginViaEmailVerificationCode: async (optin) => {
+    return apiGet(
+      CONFIG.LOGIN_API_URL + 'loginViaEmailVerificationCode?emailVerificationCode=' + optin,
+    )
+  },
+  changePassword: async (session_id, email, password) => {
+    const payload = {
+      session_id,
+      email,
```
Mogge, line 82: We do not have the email available so far.

configs/login_server/grd_login.properties
```diff
@@ -22,7 +22,7 @@ loginServer.db.user = root
 loginServer.db.password = 
 loginServer.db.port = 3306
 
-frontend.checkEmailPath = http://localhost/account/checkEmail
+frontend.checkEmailPath = http://localhost/reset
```
Mogge, line 25: Better `resetPaaword`?

frontend/src/routes/routes.js
```diff
@@ -59,7 +59,7 @@ const routes = [
     component: () => import('../views/Pages/ForgotPassword.vue'),
   },
   {
-    path: '/reset',
+    path: '/reset/:optin',
```
Mogge, line 62: resetPassword

configs/login_server/grd_login.properties
ulfgebhardt: I can't think of any other reset functionality we would want to implement

frontend/src/views/Pages/ResetPassword.vue
ulfgebhardt: Can we solve this in general for the whole application? (Am not a fan of toasts, since they are so far away from the vision center & non-oersisten)

frontend/src/apis/loginAPI.js
ulfgebhardt: Not a Solution to this problem, but consider retrieving missing info via: https://github.com/gradido/gradido/blob/master/docu/login_server.api.md#retrieve-user-data

Mogge:
How to test it:
 1. Click on 'Forgot Password' on the login view
 2. Enter a already registered email
 3. check the table `email_opt_in` of `gradido_login`. `email_opt_in_type_id` should be 2
 4. copy the verification code given in the table
 5. navigate to: `/reset/<verification_code>`

ulfgebhardt (approved):
The described process worked for me <3

PR #215, opened 2021-04-21
refactor: Remove loginAPI Call from Store

### LINKED ISSUE #211: 🔧 [Refactor] Send Logout to API from Component
## :zap: Refactor ticket
We send the logout to the login server from the store. Once we have only one logout button, the API should be called in the component and not in the store.

Mogge:
## 🍰 Pullrequest
Now all API calls to the login_server are removed from the store. They are all in the correspondent components, so we can do a proper error handling if a call fails. 

### Issues
- fixes #211


ulfgebhardt (approved):
Looks good.

PR #225, opened 2021-04-22
Rollback use prebuild dependencies

einhornimmond:
## 🍰 Pullrequest
Using again prebuild dependencies in docker image, 
but keep mounted /src from login-server and community-server for easyer update on changes


ulfgebhardt (approved):
Looks good

PR #227, opened 2021-04-23
fix empty reply by empty group_id

einhornimmond:
## 🍰 Pullrequest
Since stage2 users belong to a group. 
But until now unsecureLogin cannot handle this state. 
I have changed it now. Now if group_id from user is 0, it will be set to 1
Additional I have added a error message if user state has an unknown value.

### DIFF
--- a/login_server/src/cpp/JSONInterface/JsonUnsecureLogin.cpp
+++ b/login_server/src/cpp/JSONInterface/JsonUnsecureLogin.cpp
@@ -79,6 +79,8 @@ Poco::JSON::Object* JsonUnsecureLogin::handle(Poco::Dynamic::Var params)
 		USER_DISABLED
 	*/
 	auto user_state = session->loadUser(email, password);
+	auto user_model = session->getNewUser()->getModel();
+	Poco::JSON::Array infos;
 	
 	switch (user_state) {
 	case USER_EMPTY:
@@ -98,6 +100,10 @@ Poco::JSON::Object* JsonUnsecureLogin::handle(Poco::Dynamic::Var params)
 		result->set("state", "disabled");
 		result->set("msg", "user is disabled");
 		break;
+	case USER_NO_GROUP: 
+		user_model->setGroupId(1);
+		user_model->updateIntoDB("group_id", 1);
+		infos.add("set user.group_id to default group_id = 1");
 	case USER_NO_PRIVATE_KEY:
 	case USER_COMPLETE:
 	case USER_EMAIL_NOT_ACTIVATED:
@@ -105,7 +111,14 @@ Poco::JSON::Object* JsonUnsecureLogin::handle(Poco::Dynamic::Var params)
 		result->set("user", session->getNewUser()->getJson());
 		result->set("session_id", session->getHandle());
 		session->setClientIp(mClientIP);
+		if(infos.size() > 0) {
+			result->set("info", infos);
+		}
 		return result;
+	default: 
+		result->set("state", "error");
+		result->set("msg", "unknown user state");
+		result->set("details", USER_NO_GROUP);
 	}
 	
 	sm->releaseSession(session);


login_server/src/cpp/JSONInterface/JsonUnsecureLogin.cpp
```diff
@@ -98,14 +100,25 @@ Poco::JSON::Object* JsonUnsecureLogin::handle(Poco::Dynamic::Var params)
 		result->set("state", "disabled");
 		result->set("msg", "user is disabled");
 		break;
+	case USER_NO_GROUP: 
+		user_model->setGroupId(1);
+		user_model->updateIntoDB("group_id", 1);
+		infos.add("set user.group_id to default group_id = 1");
```
ulfgebhardt, line 106: no `break` ?

ulfgebhardt (changes requested):

login_server/src/cpp/JSONInterface/JsonUnsecureLogin.cpp
einhornimmond: that's right, no break
In for most user_state simply a error message will be set, breaked, session will be deleted and returned. 

For USER_NO_PRIVATE_KEY, USER_COMPLETE and USER_EMAIL_NOT_ACTIVATED
a state will be set to success, session_id and user added to return json and info if not empty and then directly return without break. 
For USER_NO_GROUP a default group will be set and saved in db, a info will be added for frontend 
and then go on with set state to success, set user and session_id, set info and return like with user_state USER_NO_PRIVATE_KEY, USER_COMPLETE or USER_EMAIL_NOT_ACTIVATED.

```cpp
switch (user_state) {
	case USER_EMPTY:
	case USER_PASSWORD_INCORRECT:
		result->set("state", "error");
		result->set("msg", "password incorrect");
		break;
	case USER_PASSWORD_ENCRYPTION_IN_PROCESS:
		result->set("state", "processing");
		result->set("msg", "password encryption in process");
		break;
	case USER_KEYS_DONT_MATCH:
		result->set("state", "error");
		result->set("msg", "saved keys mismatch");
		break;
	case USER_DISABLED:
		result->set("state", "disabled");
		result->set("msg", "user is disabled");
		break;
	case USER_NO_GROUP: 
		user_model->setGroupId(1);
		user_model->updateIntoDB("group_id", 1);
		infos.add("set user.group_id to default group_id = 1");
	case USER_NO_PRIVATE_KEY:
	case USER_COMPLETE:
	case USER_EMAIL_NOT_ACTIVATED:
		result->set("state", "success");
		result->set("user", session->getNewUser()->getJson());
		result->set("session_id", session->getHandle());
		session->setClientIp(mClientIP);
		if(infos.size() > 0) {
			result->set("info", infos);
		}
		return result;
	default: 
		result->set("state", "error");
		result->set("msg", "unknown user state");
		result->set("details", USER_NO_GROUP);
	}
	
	sm->releaseSession(session);
	
	return result;
``` 

So for user_state USER_NO_GROUP 
this code will run:
```cpp
user_model->setGroupId(1);
user_model->updateIntoDB("group_id", 1);
infos.add("set user.group_id to default group_id = 1");
result->set("state", "success");
result->set("user", session->getNewUser()->getJson());
result->set("session_id", session->getHandle());
session->setClientIp(mClientIP);
if(infos.size() > 0) {
	result->set("info", infos);
}
return result;
```

ulfgebhardt: I would approve this - tho I believe this way is very error prune. If you forget to add a break above the usergroup thing it will result in a successful return + the group_id code execution.

ulfgebhardt (approved):

login_server/src/cpp/JSONInterface/JsonUnsecureLogin.cpp
einhornimmond: Yes that's right, but the whole user group block is only temporary until the new frontend can handle this.

PR #248, opened 2021-04-27
fix bug with object instead of array

einhornimmond:
### fix error from moriz

POST http://localhost/api/sendCoins/
amount    340000
auto_sign    true
email    "bob@example.org"
memo    "Danke für Deine Hilfe"
session_id    -906396123
response:
message    "Cannot use object of type stdClass as array"
url    "/api/sendCoins/"
code    500
file    "/var/www/cakephp/src/Controller/AppRequestsController.php"
line    101


Mogge:
see #260. This PR is merged there

einhornimmond:
nice

PR #250, opened 2021-04-27
update skeema call to allow also for unsafe db updates

einhornimmond:
### Issues
<!-- Which Issues does this fix, which are related?
- fix error by updating db schema

### DIFF
--- a/skeema/Dockerfile
+++ b/skeema/Dockerfile
@@ -25,4 +25,4 @@ COPY ./mariadb/.skeema.login ./gradido_login/.skeema
 COPY ./community_server/skeema/ . 
 COPY ./mariadb/.skeema.community ./gradido_community/.skeema
 
-CMD skeema push
+CMD skeema push --allow-unsafe


skeema/Dockerfile
```diff
@@ -25,4 +25,4 @@ COPY ./mariadb/.skeema.login ./gradido_login/.skeema
 COPY ./community_server/skeema/ . 
 COPY ./mariadb/.skeema.community ./gradido_community/.skeema
 
-CMD skeema push
\ No newline at end of file
+CMD skeema push --allow-unsafe
```
ulfgebhardt, line 28: Whats the effect? Why?

einhornimmond: To allow db updates which are considered unsafe from skeema like transforming this:
 `received` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP
to this:
 `received` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP
 
or for example removing fields from table. 
 
If I don't do this, every time such a db update takes place, everyone must delete there volume db_vol to get a fresh new db and must put back all his test accounts again. 
Skeema is currenlty only used in debug build so is hasn't any effect on production build.

ulfgebhardt: This is a classical situation a migration should be used.
Reading up on skeema https://www.skeema.io/docs/workflow/ it is unclear if it actually supports migrations. It has a command `skeema diff` but I am unsure if thats a static thing or actually generates a migration???

-> In light of this: Whats the purpose we use skeema for? Does it support migrations?
Mind to have a look aswell @Mogge ?

einhornimmond: With skeema push skeema compare the current table structure with the structure defined in file and create an statement to change to the new structure. 
I use skeema already some time and I had never problems with my existing table entries. 
If a field was added, it's get his default value in every existing entry. 

Nearly every time I use skeema in the past I had already data in my tables.

ulfgebhardt: This is a quite uncommon strategy for database migration I would say and it did not work with the `group_id` case, did it?

ulfgebhardt (approved):

skeema/Dockerfile
einhornimmond: It is only a database schema migration.

PR #254, opened 2021-04-27
Login fix pending task default

einhornimmond:
## 🍰 Pullrequest

fix bug which Moriz has found with missing initial values for two parameters from pending_tasks


Mogge:
see #260. It is merged there

PR #261, opened 2021-04-28
Fix problems with transactions in docker environment

einhornimmond:
### Issues
When testet with Docker login-server couldn't reach community-server because he uses the same url for json-requests at community-server which he uses for link-buildung. In Bare-Metal Setups this is fine. If booth running on the same machine it is localhost or if community-server is running elsewhere he should have url.
But with docker the url for links is localhost and the url for json-requests is nginx.

Since stage2 the community-server urls are taken from groups entry. 

So basically I added a host field to groups table and update the sql scripts (schema and insert script) accordingly. 

I have also found and fix some bugs when building docker in a new folder. 

The insert script only work for new databases so if you use you old database you must update the groups.host field in login-server db manually. 

With SQL command in localhost:8074
update groups set host = "nginx";
you can update all entries add once.


Mogge:
Can we have migrations to alter the schema (not now, but in the future)?

ulfgebhardt:
I thought this is what skeema is for actually?! Make migrations, am I wrong?

login_server/skeema/gradido_login/insert/setup_hedera_group.sql
```diff
@@ -1,13 +1,13 @@
...
+(2, 'dockerStage2', 'docker stage2 gradido group', 'localhost', 'nginx', '/', 'gradido test group for docker and stage2');
 
 INSERT INTO `hedera_ids` (`id`, `shardNum`, `realmNum`, `num`) VALUES
 (1, 0, 0, 3),
 (2, 0, 0, 3327),
 (3, 0, 0, 413151);
 
 INSERT INTO `node_servers` (`id`, `url`, `port`, `group_id`, `server_type`, `node_hedera_id`, `last_live_sign`) VALUES
-(1, 'http://0.testnet.hedera.com', 50211, 0, 4, 1, '2000-01-01 00:00:00');
+(1, 'http://0.testnet.hedera.com', 50211, 0, 4, 2, '2000-01-01 00:00:00');
```
ulfgebhardt, line 10: This should cause duplicate keys exception when applying assuming that `id` is a primary key. MySQL allows `INSERT REPLACE`, but I am unsure wether thats the correct approach.

While on live this data is not present yet, the other installations may run into problems (e.g. dockers build lately, maybe staging)

login_server/src/cpp/controller/Group.cpp
```diff
@@ -17,7 +17,7 @@ namespace controller {
 
 	Poco::AutoPtr<Group> Group::create(const std::string& alias, const std::string& name, const std::string& url, const std::string& home, const std::string& description)
 	{
-		auto db = new model::table::Group(alias, name, url, home, description);
+		auto db = new model::table::Group(alias, name, url,"", home, description);
```
ulfgebhardt, line 20:
```suggestion
		auto host = "";
		auto db = new model::table::Group(alias, name, url, host, home, description);
```

Readability over optimization, that's the task of the compiler

einhornimmond:
Skeema is for altering schema yes, but it cannot alter the content.

login_server/skeema/gradido_login/insert/setup_hedera_group.sql
einhornimmond: Maybe I should move the insert scripts into another folder to prevent confusion. 
The insert scripts cannot be run by skeema. 
They only be run by mariadb container if db_vol volume is empty also if you have a empty db.

login_server/src/cpp/controller/Group.cpp
einhornimmond: Please be caution with code suggestions in c++ with auto type. 
In this case I'm not sure if it works like expected because for the compiler it isn't really clear which type he should use. 
"" could be string or char[].
But as this was provisional I have now make it right.

login_server/skeema/gradido_login/insert/setup_hedera_group.sql
einhornimmond: I have moved the insert scripts.

ulfgebhardt (approved):
Looks good

PR #266, opened 2021-04-29
turn registration on and off

ogerly:
## 🍰 Pullrequest

turn registration on and off
#### ... frontend/Dockerfile
In the Dockerfile, I specify whether registration is allowed. 

     ##Register in App  true || false
     ENV VUE_APP_ALLOW_REGISTER="true"
     

#### ... frontend/src/config/index.js
     const environment = {
         ...
         VUE_APP_ALLOW_REGISTER: process.env.VUE_APP_ALLOW_REGISTER === 'true',
      }

#### ... frontend/src/routes/router.js
     if (CONFIG.VUE_APP_ALLOW_REGISTER) {
        router.addRoute({
           path: '/register',
           component: () => import('../views/Pages/Register.vue'),
         })
     }

### DIFF
--- a/frontend/.env.dist
+++ b/frontend/.env.dist
@@ -1,2 +1,3 @@
 LOGIN_API_URL=http://localhost/login_api/
 COMMUNITY_API_URL=http://localhost/api/
+ALLOW_REGISTER=true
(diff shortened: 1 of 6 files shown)


ulfgebhardt:
Can you rename the variable from `VUE_APP_ALLOW_REGISTER` to `ALLOW_REGISTER` please or whats the reasoning you use the `VUE_APP` prefix?
Furthermore I believe the default value is not calculated correctly in the config. Use something like this:
` INVITE_REGISTRATION: process.env.INVITE_REGISTRATION !== 'false', // default = true `

ulfgebhardt (approved):

PR #270, opened 2021-04-29
refactor: Test and Clean Up Store

Mogge:
## 🍰 Pullrequest
 * GDT balance is no longer stored in store. It is a property of the DashboardLayout (as balance)
 * Testing the store
 * As create user and login do the same, just login is used.
 * Coverage to 15%

### DIFF
--- a/frontend/src/store/store.test.js
+++ b/frontend/src/store/store.test.js
@@ -1,0 +1,81 @@
+import { mutations, actions } from './store'
+
+const { language, email, session_id } = mutations
+const { login, createUser, logout } = actions
+
+describe('Vuex store', () => {
+  describe('mutations', () => {
+    describe('language', () => {
+      it('sets the state of language', () => {
+        const state = { language: 'en' }
+        language(state, 'de')
+        expect(state.language).toEqual('de')
+      })
+    })
+
+    describe('email', () => {
+      it('sets the state of email', () => {
+        const state = { email: 'nobody@knows.tv' }
+        email(state, 'someone@there.is')
+        expect(state.email).toEqual('someone@there.is')
+      })
+    })
+
+    describe('session_id', () => {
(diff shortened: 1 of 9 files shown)


frontend/src/store/store.test.js
```diff
@@ -0,0 +1,81 @@
...
+        expect(commit).toHaveBeenNthCalledWith(1, 'session_id', null)
+      })
+
+      it('commits email', () => {
+        logout({ commit, state })
+        expect(commit).toHaveBeenNthCalledWith(2, 'email', null)
+      })
+
+      // how can I get this working?
+      it.skip('calls sessionStorage.clear()', () => {
```
Mogge, line 74: I spent nearly two hours, But I cannot get this working :(

frontend/src/views/Pages/Register.vue
```diff
@@ -174,7 +174,7 @@ export default {
         this.password,
       )
       if (result.success) {
-        this.$store.dispatch('createUser', {
+        this.$store.dispatch('login', {
```
Mogge, line 177: I think we do not need this. After register, the user is redirected to the thank page. There is a link to login. So way store a session id when the user has to login again and the receives as session id? Does not make sense to me.

frontend/src/views/Layout/DashboardLayout_gdd.vue
```diff
@@ -92,6 +97,14 @@ export default {
         // what to do when loading balance fails?
       }
     },
+    async loadGDTBalance() {
+      const result = await communityAPI.transactions(this.$store.state.session_id)
+      if (result.success) {
+        this.GdtBalance = result.result.data.gdtSum
```
Mogge, line 103: / 10000 ??

ulfgebhardt (approved):
Looks good <3

PR #273, opened 2021-05-03
give balance as float value in GDD (not longer GDD cent)

### LINKED ISSUE #171: 🚀 [Feature] Give Frontend GDD float
## :rocket: Feature
Return all GDDs to frontend as float already divided by 10.000

### LINKED ISSUE #327: 🐛 [Bug] More than 5 Transactions Are Shown In AccountOverview
## 🐛 Bugreport

image

Only five transactions

einhornimmond:
### Issues
- fixes #171
- fixes #327


Mogge:
image

I receive a precision of 3 from the server. I would expect 2. E.g `-7.98 GDD`  instead of `-7.983 GDD`

Please see my questions in the code.

frontend/src/views/KontoOverview/GddSend.vue
```diff
@@ -255,6 +255,7 @@ export default {
       const result = await communityAPI.send(
         this.$store.state.session_id,
         this.ajaxCreateData.email,
+        // better to send the user's input here?
         this.ajaxCreateData.amount * 10000,
```
Mogge, line 259: This is important. I would prefer to send `3.14` instead of `31400`

frontend/src/views/Layout/DashboardLayout_gdd.vue
```diff
@@ -94,15 +94,15 @@ export default {
...
+        this.balance = result.result.data.balance// / 10000
       } else {
         // what to do when loading balance fails?
       }
     },
     async loadGDTBalance() {
       const result = await communityAPI.transactions(this.$store.state.session_id)
       if (result.success) {
-        this.GdtBalance = result.result.data.gdtSum / 10000
+        this.GdtBalance = result.result.data.gdtSum// / 10000
```
Mogge, line 105: Is this the expected behavior of GDT as well? Cannot test...

Mogge (changes requested):
* Please check the precision (2 Nachkommastellen)
 * Same for GDT?
 * What happens with the amount send by an transaction?

community_server/src/Model/Entity/StateBalance.php
```diff
@@ -70,7 +71,7 @@ protected function _getDecay()
     
     protected function _getAmountFloat()
     {
-        return floatval($this->amount) / 10000.0;
+        return Number::format(floatval($this->amount) / 10000.0, ['precision' => 2]);
```
ulfgebhardt, line 74: I assume this uses `round()` which rounds up in cases of `x.xx5 - x.xx9` - this is incorrect since the User would be displayed more GDD then he has.

community_server/src/Model/Table/TransactionsTable.php
```diff
@@ -188,7 +188,7 @@ public function listTransactionsHumanReadable($stateUserTransactions, array $use
                     //echo "prev balance: " . $prev->balance . ", diff_amount: $diff_amount, summe: " . (-intval($prev->balance - $diff_amount)) . "<br>";
                     $final_transactions[] = [ 
                         'type' => 'decay',
-                        'balance' => floatval(-intval($prev->balance - $diff_amount)) / 10000.0,
+                        'balance' => Number::format(floatval(-intval($prev->balance - $diff_amount)) / 10000.0, ['precision' => 2]),
```
ulfgebhardt, line 191: Can you reuse the conversion - make its own function you call from everywhere to have correct values ?

community_server/src/Model/Table/TransactionsTable.php
```diff
@@ -256,7 +256,7 @@ public function listTransactionsHumanReadable($stateUserTransactions, array $use
                 $state_balance->record_date = $su_transaction->balance_date;
                 $final_transactions[] = [
                     'type' => 'decay',
-                    'balance' => floatval(-intval($su_transaction->balance - $state_balance->decay)) / 10000,
+                    'balance' => Number::format(floatval(-intval($su_transaction->balance - $state_balance->decay)) / 10000.0, ['precision' => 2]),
```
ulfgebhardt, line 259: Same here, please do not duplicate code, reuse it

login_server/src/cpp/JSONInterface/JsonCreateTransaction.cpp
```diff
@@ -121,8 +121,12 @@ Poco::JSON::Object* JsonCreateTransaction::transfer(Poco::Dynamic::Var params)
 			result = stateError("user not in group", "receiver user isn't in target group");
 		}
 	}
+	auto sender_user = mSession->getNewUser();
+	if (sender_user->getGradidoKeyPair()->isTheSame(*target_pubkey)) {
+		result = stateError("sender and receiver are the same");
+	}
 	if (!result) {
-		auto transaction = model::gradido::Transaction::createTransfer(mSession->getNewUser(), target_pubkey, mTargetGroup, amount, mMemo, mBlockchainType);
+		auto transaction = model::gradido::Transaction::createTransfer(sender_user, target_pubkey, mTargetGroup, amount, mMemo, mBlockchainType);
```
ulfgebhardt, line 129: This seems to be in the wrong PR? #277

ulfgebhardt (changes requested):

login_server/src/cpp/JSONInterface/JsonCreateTransaction.cpp
einhornimmond: You are right, it's belong in the other one.

community_server/src/Model/Table/TransactionsTable.php
einhornimmond: resolved with commit b10886a

community_server/src/Model/Table/TransactionsTable.php
einhornimmond: resolved with commit b10886a

frontend/src/views/KontoOverview/GddSend.vue
einhornimmond: erledigt

frontend/src/views/Layout/DashboardLayout_gdd.vue
einhornimmond: gdt cames now also as float

community_server/src/Model/Entity/StateBalance.php
einhornimmond: I now use a own function with floor like you have wished for

community_server/src/Controller/AppRequestsController.php
```diff
@@ -337,14 +341,29 @@ public function listTransactions($page = 1, $count = 25, $orderDirection = 'ASC'
...
+            'count' => count($transactions),
+            'gdtSum' => $gdtSum,
+            'timeUsed' => microtime(true) - $startTime
+        ];
+        $now = new FrozenTime();
+        $body['decay_date'] = $now;
+        
+        if(!$state_balance) {
+            $body['balance'] = 0;
+            $body['decay'] = 0;
```
ulfgebhardt, line 360: this is not a float. I assume js will be good with zero, but shouldn't it be

```suggestion
            $body['balance'] = 0.0;
            $body['decay'] = 0.0;
```

community_server/src/Template/AppRequests/list_transactions.ctp
```diff
@@ -0,0 +1,17 @@
...
+ * and open the template in the editor.
+ */
+
+$body['balance'] = $this->element('centToFloat', ['cent' => $body['balance'], 'precision' => 4]);
+$body['decay'] = $this->element('centToFloat', ['cent' => $body['decay'], 'precision' => 4]);
+$body['gdtSum'] = $this->element('centToFloat', ['cent' => $body['gdtSum'], 'precision' => 2]);
+
+foreach($body['transactions'] as $i => $transaction) {
+    $body['transactions'][$i]['balance'] = $this->element('centToFloat', ['cent' => $transaction['balance'], 'precision' => 4]);
+}
```
ulfgebhardt, line 15: this seems like a hack.

Am I right that this comes directly from the database and later on is transformed with this loop?
Or is it that the `$body['transactions']`is build by us in the code?

The ideal way would be:

```
$result = []
while(i can read from db){
   $result[] = transform(db.row)
}
```
That way we would only have 1 loop. This is more about effectiveness then correctness, therefore you can neglect this. If we build the transactions array ourselves with a loop somewhere in the code, I would suggest to move the `centToFloat` call there.

community_server/src/Template/Element/centToFloat.ctp
```diff
@@ -0,0 +1,15 @@
...
+ * To change this template file, choose Tools | Templates
+ * and open the template in the editor.
+ */
+
+$cut_places = $precision - 2;
+$transformAmount = $cent;
+if($cut_places > 0) {
+    $transformAmount = floor($cent / pow(10, $cut_places));
+}
+echo $transformAmount / pow(10, $precision -  $cut_places);
```
ulfgebhardt, line 14: What does this do? `$cut_places` is 0 or negative. I do not understand what happens here.

docu/community-server.api.md
```diff
@@ -65,7 +65,7 @@ Assuming: session is valid
 			"type": "send",
 			"transaction_id": 2,
 			"date": "2021-02-19T13:25:36+00:00",
-			"balance": 1920000,
+			"balance": 192,
```
ulfgebhardt, line 68: this must be a float aswell. So it should be
```suggestion
			"balance": 192.0,
```

frontend/src/views/Pages/AccountOverview/GddSend.vue
```diff
@@ -254,6 +254,7 @@ export default {
       const result = await communityAPI.send(
         this.$store.state.session_id,
         this.ajaxCreateData.email,
+        // better to send the user's input here?
         this.ajaxCreateData.amount * 10000,
```
ulfgebhardt, line 258: Will you adjust the send call in this PR aswell? @einhornimmond

ulfgebhardt (changes requested):
Some more comments and unclear code from my side, but in general this starts to look fantastic!
Most important decision we have to discuss is if we want to adjust the sendGDD call in this PR as well, because this is not present in here yet as far as I can see.

Update: As I understand it sendGDD is part of this PR - it seems to be very good code tho, because the effected lines are very little and I did not recognize it therefore. My apologies

community_server/src/Template/AppRequests/list_transactions.ctp
einhornimmond: CakePHP is a MVC-Framework, you suggestet that I use the transform code only once. 
I can use Component which all Controller can use, I can use a Behaviour which all Model Entitys can use or can make a element which all views can use. The code which build together the transaction list is splittet in Controller and Model Table Classes.
So the best solution from point of view was to use the view. It's belong naturally there because it is only for the user to see. 
So I must use a additional loop in view to transform. 
Yes from performance point of view it is not optimal, from performant point of view my first approach was optimal, but you don't like it.

einhornimmond: It don't came directly from db, it is builded by Model/Table/Transactions::listTransactionsHumanReadable

einhornimmond: Actually I would change code on 4 places when I make the transformation direct in Model/Table/Transactions::listTransactionsHumanReadable in the loop.

einhornimmond: Plus I must make a new transformation function because I cannot call the element inside a model table class.

community_server/src/Template/Element/centToFloat.ctp
einhornimmond: With GDD precision is 4 with GDT precision is 2.
Maybe I have choosen a wrong name here. 
For GDD cut_places is 2 and 134156 GDD cent will be in first step rounded down to
1341 and then to 13,41.
For GDT this extra step isn't needed.

frontend/src/views/Pages/AccountOverview/GddSend.vue
einhornimmond: yes

community_server/src/Template/AppRequests/list_transactions.ctp
ulfgebhardt: Its fine, thanks for clarifying ❤️

community_server/src/Template/Element/centToFloat.ctp
ulfgebhardt: ok, just seems quite complicated - resolving this as well, ty ❤️

frontend/src/views/Pages/AccountOverview/GddSend.vue
ulfgebhardt, line 258:
```suggestion
        this.ajaxCreateData.amount,
```
@Mogge Please check if this is correct and works

ulfgebhardt (changes requested):
From my point of view this is fine now except the frontend - that would be the task of @Mogge to check if stuff works and we can remove that *10⁴ from the sendGDD call.

frontend/src/views/Pages/AccountOverview/GddSend.vue
Mogge: CHECKED

ulfgebhardt (approved):

community_server/src/Template/Element/centToFloat.ctp
einhornimmond: I thought when I make it variable then I can make it direct as variable as possible, so it can be used easy also for similar variants.

PR #279, opened 2021-05-04
feat: Visualize Decay of GDD

Mogge:
## 🍰 Pullrequest
Updates every second the balance of GDD a user has.

### DIFF
--- a/frontend/src/views/Layout/DashboardLayout_gdd.vue
+++ b/frontend/src/views/Layout/DashboardLayout_gdd.vue
@@ -1,7 +1,7 @@
 <template>
   <div class="wrapper">
     <notifications></notifications>
-    <side-bar @logout="logout" :balance="balance">
+    <side-bar @logout="logout" :balance="displayedBalance">
       <template slot="links">
         <b-nav-item href="#!" to="/overview">
           <b-nav-text class="p-0 text-lg text-muted">{{ $t('send') }}</b-nav-text>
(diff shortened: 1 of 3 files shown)


frontend/src/views/Layout/DashboardLayout_gdd.vue
```diff
@@ -72,12 +73,19 @@ export default {
...
   data() {
     return {
       balance: 0,
       GdtBalance: 0,
     }
   },
+  computed: {
+    displayedBalance() {
+      const str = String(this.balance)
+      return Number(str.substring(0, str.indexOf('.') + 3))
```
Mogge, line 86: Is this enough to have always two digits? Can this fail in case of `str = 1.1`?

frontend/src/views/Layout/DashboardLayout_gdd.vue
```diff
@@ -110,6 +118,12 @@ export default {
     updateBalance(ammount) {
       this.balance -= ammount
     },
+    decay() {
+      this.balance *= 0.99999997802044727
```
ulfgebhardt, line 122: Is this correct? Can you comment this why this number please?

frontend/src/views/Layout/DashboardLayout_gdd.vue
ulfgebhardt: Sometimes we just get a full integer, this might change with the float change. So you might wanna do this:
`floor(this.balance,2)` or `floor(this.balance << 2) >> 2`

ulfgebhardt (changes requested):

ogerly (approved):

ulfgebhardt:
@Mogge Do you keep going with this?

ulfgebhardt:
@Mogge ping

Elweyn:
Can be used if we need to trigger events in certain period of time!

PR #304, opened 2021-05-05
error messages are displayed

### LINKED ISSUE #240: 🐛 [Bug] Create user error not caught
## :bug: Bugreport

When creating a user and you filled in all fields the signup button appears (this is not good)
When clicking signup you are for sure redirected to the thanks page(?). If the create user call fails this behaviour is happening anyway. Therefor the user is thanked and gets the impression that everything is alright, while its not.

ogerly:
## 🍰 Pullrequest
error messages are displayed
the error transferred from the backend is displayed 1:1 in the error field. 

### Issues
- fixes #240 
 
FireShot Capture 369 - Gradido Account - localhost

FireShot Capture 372 - Gradido Account - localhost

### DIFF
--- a/frontend/src/views/Pages/Register.vue
+++ b/frontend/src/views/Pages/Register.vue
@@ -175,7 +175,7 @@ export default {
       passwordVisible: false,
       submitted: false,
       registerfail: false,
-      registererror: '',
+      showError: '',
     }
   },
   methods: {
--- a/frontend/src/views/Pages/Register.vue
+++ b/frontend/src/views/Pages/Register.vue
@@ -110,6 +110,20 @@
                       </base-input>
                     </b-col>
                   </b-row>
+                  <b-alert
+                    v-if="registerfail"
+                    show
+                    dismissible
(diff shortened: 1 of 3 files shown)


frontend/src/views/Pages/Register.vue
```diff
@@ -160,6 +174,8 @@ export default {
       checkPassword: '',
       passwordVisible: false,
       submitted: false,
+      registerfail: false,
```
ulfgebhardt, line 177: Can we rename this field into something more meaningful like `showError` or `errorAlertvisible` please?

ulfgebhardt (changes requested):

frontend/src/views/Pages/Register.vue
ogerly: erledigt

ulfgebhardt: Lügen!

einhornimmond (approved):

PR #329, opened 2021-05-05
don't show decays smaller than 0,01 GDD

### LINKED ISSUE #288: 🐛 [Bug] There is -0GDD decay transactions
## 🐛 Bugreport

There is -0GDD decay transactions which seem illogical since no transaction is happening. This is closely bundled with "How do we tell the user about the decay" since currently the interface is quite cluttered with decay transactions

einhornimmond:
### Issues
- fixes #288

### DIFF
--- a/community_server/src/Model/Table/TransactionsTable.php
+++ b/community_server/src/Model/Table/TransactionsTable.php
@@ -191,7 +191,7 @@ class TransactionsTable extends Table
                         //echo "prev balance: " . $prev->balance . ", diff_amount: $diff_amount, summe: " . (-intval($prev->balance - $diff_amount)) . "<br>";
                         $final_transactions[] = [ 
                             'type' => 'decay',
-                            'balance' => -intval($prev->balance - $diff_amount),
+                            'balance' => $balance,
                             'decay_duration' => $interval->format('%a days, %H hours, %I minutes, %S seconds'),
                             'memo' => ''
                         ];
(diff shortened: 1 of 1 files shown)


community_server/src/Model/Table/TransactionsTable.php
```diff
@@ -183,15 +183,19 @@ public function listTransactionsHumanReadable($stateUserTransactions, array $use
...
-                    $final_transactions[] = [ 
-                        'type' => 'decay',
-                        'balance' => -intval($prev->balance - $diff_amount),
-                        'decay_duration' => $interval->format('%a days, %H hours, %I minutes, %S seconds'),
-                        'memo' => ''
-                    ];
+                    $balance = -intval($prev->balance - $diff_amount);
+                    // skip small decays (smaller than 0,00 GDD)
+                    
+                    if(abs($balance) > 100) {
```
ulfgebhardt, line 189:
```suggestion
                    if(abs($balance) >= 100) {
```

We want to see cent changes - i think it should therefore be `greater equal`

ulfgebhardt (changes requested):
Once my suggestion is clarified if its valid I would approve.
But in the future we need to refactor this is some way. Decay must be aligned to an transaction and not be its own independent transaction. This is because we want to keep the frontend as stupid as we can. We do not want to have any enhanced ordering or convert logic in there. It is very impractical for the user to have a decay transaction after each "normal" transaction. Therefore we need to find a solution how to send those two chained together to the client.

einhornimmond:
I can also put the decay transaction into the transaction. 
The decay transactions are virtual, they are calculated and don't exist in this form in db.

ulfgebhardt:
I think this should be done in the future @einhornimmond 
We discussed with @BerndHueckstaedt yesterday that it might be good to communicate this now as we currently do it, since the decay is a new feature. In the future we should see to it so that the transaction list is easily readable - i will make a specific ticket for that.

ulfgebhardt (approved):

PR #331, opened 2021-05-05
change transaction default order

### LINKED ISSUE #328: 🐛 [Bug] Wrong Order of Transactions
## 🐛 Bugreport

Order of transactions is ascendant. Please descendant as default 

The transaction update method is not called from GddTable.

einhornimmond:
- fixes #328

### DIFF
--- a/community_server/src/Controller/AppRequestsController.php
+++ b/community_server/src/Controller/AppRequestsController.php
@@ -292,7 +292,7 @@ class AppRequestsController extends AppController
         ]);
     }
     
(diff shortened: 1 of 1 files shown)


Mogge:
I suggest to do this in the with the default request.

Mogge:
I think  this is the better solution: #332

Mogge (changes requested):
See my comments. I suggest to close this approach.

ulfgebhardt:
Both approaches do not exclude each other. Tho we have to think about if this change effects the old frontend.

PR #344, opened 2021-05-07
eslint rule standard

### LINKED ISSUE #340: 💥 [DevOps] Get the Standard Test Working for Eslint
## 💥 DevOps ticket
We want to use the standard rules of eslint. see `.eslintrc`
```
  extends: [
    'standard',
    'plugin:vue/essential',
    'plugin:prettier/recommended'
  ],
```
But it throws an error:
```
Error: standard:
        Configuration for rule "accessor-pairs" is invalid:
        Value {"setWithoutGet":true,"enforceForClassMembers":true,"getWithoutSet":false} should NOT have additional properties.
```
The standard rules find functions or variables that are not in use. I assume we have a lot of them.

ulfgebhardt:
## 🍰 Pullrequest

@Mogge use `yarn upgrade-interactive --latest`

### Issues

- fixes #340 

### Todo

- [X] Fix all errors


Mogge (approved):
I tested the API calls. Everything works as expected

Issue #350, opened 2021-05-07
🐛 [Bug] Minor bugs & translation errors


ulfgebhardt:
## 🐛 Bugreport

- [x] Footer: Gradido Akademie link needs `target=_blank` #233 
- [x] Menu: Member areas should be opened in a  new tab #370
- [x] Design: No message breaks the layout of send GDD page 2 (Confirmation) #375
- [x] Hide: QR-Code scanner #380
- [x] Hide: QR-Code of the user in the top right corner #380
- [x] Design: Warning on the Send GDD Page is not easily readable - white on red is not good #381 
- [x] Design: Highlighting on email/qr code in top right corner without a function - remove #380
- [x] Design: Account overview - distance below GDD & GDT amount too big #382
- [x] Login: Error message cannot be closed #383 
- [x] Login: Error message must disappear when retrying (for a moment if unsuccessful again)  #383 
- [x] Spelling: Password reset has spelling mistake - `zukünftig` #385
- [x] Spelling: Password reset has `GRADIDO` in CAPS, should not be caps. #385
- [x] Translate: `Sprache - en` must be `Language - en` , same for the German equivalent #385
- [x] Translate: Login heading is not translated should be `Anmeldung` #385
- [?] Translate: Passwort requirements not translated (password reset) https://github.com/gradido/gradido/issues/350#issuecomment-836863727
- [x] Translation: ___Rename "Verfall" to "Vergänglichkeit"___ https://github.com/gradido/gradido/issues/350#issuecomment-836860593
- [x] Design: On the 3. Send GDD page there is two horizontal lines below the "Thanks... [Button]" part. Remove one. #385

This is a collection issue for small tasks. Feel free to make separate PRs for a subset of those issues

ogerly:
Translation: Rename "Verfall" to "Vergänglichkeit" 

das habe ich nur als beispiel in einem json file simuliert. "Verfall" gibt es nirgends im code bis jetzt. den beispiel code habe ich wiede rgelöscht.

ogerly:
Translate: Passwort requirements not translated (password reset)  

keine ahnung was gemeint ist.

ulfgebhardt:
Beim passwort reset sind "min 8 characters" etc nur auf english.

PR #382, opened 2021-05-10
fix: Account overview - distance below GDD & GDT amount too big

ogerly:
## 🍰 Pullrequest
Fix:  Account overview - distance below GDD & GDT amount too big
teil von: #350 

FireShot Capture 400 - Gradido Account - localhost


ulfgebhardt:
I would have made it smaller? Not add more on the bottom? What you guys think @Team?

PR #385, opened 2021-05-10
spelling & translation

ogerly:
## 🍰 Pullrequest

      -  Spelling: Password reset has spelling mistake - zukünftig
      -  Spelling: Password reset has GRADIDO in CAPS, should not be caps.
      -  Translate: Sprache - en must be Language - en , same for the German equivalent
      -  Translate: Login heading is not translated should be Anmeldung
      -  Design: On the 3. Send GDD page there is two horizontal lines below the "Thanks... [Button]" part. Remove one.
 
teil von : #350

### DIFF
--- a/frontend/src/locales/en.json
+++ b/frontend/src/locales/en.json
@@ -13,7 +13,7 @@
   "back":"Back",
   "send":"Send",
   "transactions":"Transactions",
-  "language":"languages",
+  "language":"Language",
   "languages":{
     "de": "Deutsch",
     "en": "English"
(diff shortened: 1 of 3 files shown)


frontend/src/locales/en.json
```diff
@@ -13,7 +13,7 @@
   "back":"Back",
   "send":"Send",
   "transactions":"Transactions",
-  "language":"Sprache",
+  "language":"languages",
```
ulfgebhardt, line 16:
```suggestion
  "language":"Language",
```

ulfgebhardt (changes requested):

Mogge (approved):

PR #396, opened 2021-05-11
feat: Dash is Shown When Balance is Loading

Mogge:
## 🍰 Pullrequest
Shows a dash instead of 0 when balance is loading.


frontend/src/components/SidebarPlugin/SideBar.spec.js
```diff
@@ -0,0 +1,102 @@
...
+      describe('logout', () => {
+        it('has a logout button', () => {
+          expect(wrapper.findAll('li').at(1).text()).toBe('logout')
+        })
+
+        it.skip('emits logout when logout is clicked', async () => {
+          const spy = jest.spyOn(wrapper.vm, 'logout')
+          wrapper.findAll('li').at(1).find('a').trigger('click')
+          await wrapper.vm.$nextTick()
+          expect(spy).toHaveBeenCalled()
```
Mogge, line 97: Why do I get 0 calls? Why does the click not trigger the method call?

frontend/src/views/Layout/DashboardLayout_gdd.vue
```diff
@@ -99,14 +101,17 @@ export default {
...
       const result = await communityAPI.transactions(this.$store.state.sessionId)
       if (result.success) {
         this.GdtBalance = Number(result.result.data.gdtSum)
         this.transactions = result.result.data.transactions
         this.balance = Number(result.result.data.decay)
         this.bookedBalance = Number(result.result.data.balance)
         this.transactionCount = result.result.data.count
+        this.pending = false
       } else {
+        this.pending = false
```
Mogge, line 114: May be better to keep pending true here. So when the API call fails, no balance is shown (instead of 0)

ulfgebhardt: That would be a bit dirty I guess but totally acceptable. Do it!

ulfgebhardt (approved):

PR #422, opened 2021-05-13
use new function for balance overview in old frontend, update balance…

### LINKED ISSUE #397: 🐛 [Bug] Old Frontend is broken for some users
## 🐛 Bugreport

Old Frontend is broken for some users

Specifically the account overview throws a lot of warnings and errors.
https://stage1.gradido.net//state-balances/overview

image

image

### LINKED ISSUE #399: 🐛 [Bug] Balance of old and new frontend are not equal
## 🐛 Bugreport

Balance of old and new frontend are not equal. This seems to be caused by a different calculation method.
If the old frontend exists for a few more months till its completely removed, I suggest to fix this, even tho its a minor bug (wrong by 0.01GDD as far as we know)

einhornimmond:
… in session on every php-request

## 🍰 Pullrequest
- Fix transactions overview on old frontend, now using same function like frontend

### Issues
- fixes #397 
- fixes #399


community_server/src/Controller/StateBalancesController.php
```diff
@@ -84,200 +85,68 @@ public function overview()
...
-        foreach ($transferTransactions as $sendCoins) {
-          //var_dump($sendCoins);
-            if ($sendCoins->state_user_id != $user['id']) {
-                array_push($involvedUserIds, intval($sendCoins->state_user_id));
-            } elseif ($sendCoins->receiver_user_id != $user['id']) {
-                array_push($involvedUserIds, intval($sendCoins->receiver_user_id));
-            }
-        }
+        $gdtSum = 0;        
+        $gdtEntries = $this->JsonRequestClient->sendRequestGDT(['email' => $user['email']], 'GdtEntries' . DS . 'sumPerEmailApi');
```
ulfgebhardt, line 96: Whats this `'GdtEntries' . DS . 'sumPerEmailApi'`? is this call documented? `[...]GdtEntries/sumPerEMailApi`

ulfgebhardt (approved):
You are deleting a ton of code - what happend there?
Furthermore you do use an API call I did not know. Could you make a ticket for its documentation if its not documented? <3

community_server/src/Controller/StateBalancesController.php
einhornimmond: no, it isn't documented, because it is part of the GDT API. 
You cannot call it from frontend because of access destrictions. 
It allows only calls from known domains.

ulfgebhardt (approved):

PR #450, opened 2021-05-17
feat: Save Locale in Database

### LINKED ISSUE #209: 🚀 [Feature] Save Locale in Database
## :rocket: Feature
Save the language the user chooses in the database. Load this language on login and set the locale in the frontend accordingly. 

### User Problem
The user must change the language every time logging in when not using the default language.

### LINKED ISSUE #353: 🚀 [Feature] Standard Language or Browser language
## 🚀 Feature

Detect the browser language and decide between the available ones or have German as default since most of our users are German speaking.

Also the language is not saved on reload (logged out status) - common use case would be password reset. The user is not logged in at that time and the interface language changes from German while requesting the reset to English when following the link. How do we handle this?

Mogge:
## 🍰 Pullrequest
Saves the locale when changed and a session_id is available in the database.

 - fixes #353 
 - fixes #209

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -212,7 +212,7 @@ jobs:
           report_name: Coverage Frontend
           type: lcov
           result_path: ./coverage/lcov.info
-          min_coverage: 15
+          min_coverage: 19
           token: ${{ github.token }}
   
   #test:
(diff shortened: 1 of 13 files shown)


ogerly (approved):
works for me, language change is saved in the database

PR #451, opened 2021-05-17
fix sorting and use total count

### LINKED ISSUE #398: 🐛 [Bug] Transaction list sorting incorrect
## 🐛 Bugreport

The transaction list is sorted incorrectly when it exceeds the pagesize. The latest transaction is not on top even tho the sort order is in that direction. @Mogge suggests that this is a problem of SQL statement order. It seems like the `LIMIT` is done before `ORDER` which is incorrect. It must be `[...] ORDER BY [...] $direction[...] LIMIT $limit`

We currently have a hackfix: We increase the page size to 1000.

einhornimmond:
## 🍰 Pullrequest
fix sorting of transaction list if more transactions exist as requested.
By the way I have see that count return only the with limit retrieved transactions + decay transaction.
I have changed it so that it returns the total count of transactions in which the user is involved. Without decay transaction, because they are generated on the fly and not saved in db. 

### Issues
- fixes #398

### DIFF
--- a/community_server/src/Controller/AppRequestsController.php
+++ b/community_server/src/Controller/AppRequestsController.php
@@ -323,23 +323,33 @@ class AppRequestsController extends AppController
           $this->addAdminError('StateBalancesController', 'overview', $gdtEntries, $user['id'] ? $user['id'] : 0);
         }
 
+        $stateUserTransactions_total = $stateUserTransactionsTable
+                                        ->find()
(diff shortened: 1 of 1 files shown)


Mogge (approved):
Ist auch in dem anderen PR mit reingerutscht. Mal sehen was passiert, wenn wir im Frontend auf paginierung umstellen.

PR #477, opened 2021-06-01
feat: GraphQL Schema

Mogge:
## 🍰 Pullrequest
add a simple GraphQL Schema


einhornimmond (changes requested):
## Group
Host isn't needed. Login-Server need it for json request to community-server running on same server, but in single server setup it isn't needed.
Home isn't needed. Login-Server need it for url construction to forward after login or check email, but in single server setup it isn't needed.

## TransactionType
for what you need decay?
Decay transactions are not saved in db, there are calculated
so it isn't possible to get them by transaction id

ulfgebhardt:
@Mogge Do we still need this? How can we get this merged?

Mogge:
Depricated

PR #483, opened 2021-06-07
move decay between transactions into the transactions

### LINKED ISSUE #318: 💬 [Question] Should decay be part of a transaction
## 💬 Question

Currently we have two transaction per transaction:
1. The transaction itself
2. The decay transaction ontop of that

This seems to be illogical to split it into two transactions and brings alot of problems:
- the transaction list gets very cluttered
- if you want to change that the client needs to transform the 1-dimensional array into a 2-dimensional manually.

Should the decay be part of a Transaction?

### LINKED ISSUE #369: 🚀 [Feature] Decay as part of a transaction
## 🚀 Feature

Decay should be part of a transaction to not clutter the transaction list. On the other hand we need to show the user that he received the correct amount. Consider the following example:

- User has 1000GDD, but has not seen his account in a while, therefore he has like 200GDD decay - effectively he has 800GDD.
- Now he receives 300GDD from a friend
- He logs in and now need to see on his transaction list, that he correctly received 300GDD, but also needs to understand that he lost 200GDD during this transaction.
- How can we communicate this? If decay is part of the transaction we would show a amount of 100GDD received which is balance wise correct but not very easy to understand.

### LINKED ISSUE #479: 🚀 [Feature] Move Decay into transactions
## 🚀 Feature
Decay informations into transactions

### LINKED ISSUE #582: 🚀 [Feature] Remove second decay from creation

einhornimmond:
## 🍰 Pullrequest
- Move decay between transactions into the transactions
- Remove second decay from creation transaction

### Issues
- fixes #479 
- fixes #318
- fixes #369 
- fixes #582

### Todo
- [x] Wait on merging Login-Coverage (#472) to update Community-Server tests for this new feature


ulfgebhardt:
I would like to postpone the merge of this till we find a way to display the decay properly in the frontend once its not its own transaction anymore. If we blindly merge this, we will/might have a communication problem.

ulfgebhardt (approved):
If we don't need to make another bugfix release, we can go ahead with this @einhornimmond

PR #484, opened 2021-06-07
Cross group transaction

### LINKED ISSUE #183: 🚀 [Feature] Cross Community User Identification
## :rocket: Feature

How do we identify user across communities? The technical answer is the public key, but how do we transfer money, if I have a community link on an user?

Consider:

1. UserA gives UserB his link : mycommunity.a/u/usera
2. UserB is fro another community: ourcommunity.b/u/userb
3. How can UserB Send GDD to UserA having his link only

We need a protocol how to transfere the public key

einhornimmond:
## 🍰 Pullrequest
Stuff which will be needed for cross group transactions
- get group alias from community server api
- get public key for username and group_alias without session from login server api

### Issues
- fixes #183

### Todo
- [ ] Tests


community_server/src/Controller/AppRequestsController.php
```diff
@@ -156,9 +169,6 @@ public function sendCoins()
         if($required_fields !== true) {
             return $this->returnJson($required_fields);
         }
-        if(!isset($params['memo']) || strlen($params['memo']) < 5 || strlen($params['memo']) > 150) {
-            return $this->returnJson(['state' => 'error', 'msg' => 'memo is not set or not in expected range [5;150]']);
-        }
```
ulfgebhardt, line 161: What happend to those restrictions?

docu/login_server.api.md
```diff
@@ -268,6 +271,7 @@ In case of success:
 		"pubkeyhex": "131c7f68dd94b2be4c913400ff7ff4cdc03ac2bda99c2d29edcacb3b065c67e6",
 		"first_name": "Max",
 		"last_name": "Musterman",
+		"email": "max.musterman@gmail.de"
```
ulfgebhardt, line 274:
```suggestion
		"email": "max.musterman@gmail.de",
```

ulfgebhardt (changes requested):
This seems to be valid changes, but I fail to understand how it solves #183 since we can still not identify a user across communities with one identifier. Furthermore it would be nice to have this in a human readable form.

How do I identify User B from Community C? Does he have a cross-community-ident-hash? does he have a globally unique username? Do we like combine username & community id into one string? How about anonymity?

einhornimmond:
For #183:
With #484 to make Cross Group Transactions frontend can:
call: GET mycommunity.a/api/getGroupAlias 
to get Group alias from target community, for example "gdd1"
then call
POST mycommunity.a/login-api/getUserInfos with 
```json 
{
	"username": "usera",
	"group_alias": "gdd1",
	"ask": [
		"user.pubkeyhex",
  	]
}
```
to get public key from usera in hex format. 
And now it has all needed data for call to sendCoins

community_server/src/Controller/AppRequestsController.php
einhornimmond: The public key is the one identifier which works across communities.
And I have added functions to get the public key from users from another community if you know the community server url and the username, like in your example.

einhornimmond: But in my plan Login-Server is able to get for the username the correct public key from blockchain of the target community. 
We will see if that is also possible with using Iota.

einhornimmond: Oh, you are right I have make mistake there. 
They moved to parseParameterForCreateTransaction
but I have forget the strlen destrictions.

PR #487, opened 2021-06-07
Login wait on passwords with missing chars

### LINKED ISSUE #135: 🚀 [Feature] Wait also on invalid passwords
Currently by login a cpu-intense hashing-function is used to make brute-force attacks harder. 
If a non-existent email is used, login-server wait the time the hashing function usually use to save cpu time but make it harder for hacker to predict which email exist. 
Additional I could also simulate this by invalid passwords rather than give with the error message hackers to much infos.

einhornimmond:
## 🍰 Pullrequest
Fake Sleep by password wich don't match security criteria

### Issues
- fixes #135

### DIFF
--- a/login_server/src/cpp/JSONInterface/JsonUnsecureLogin.cpp
+++ b/login_server/src/cpp/JSONInterface/JsonUnsecureLogin.cpp
@@ -77,7 +77,7 @@ Poco::JSON::Object* JsonUnsecureLogin::handle(Poco::Dynamic::Var params)
 	Poco::JSON::Object* result = new Poco::JSON::Object;
 
 	if (!password.size() || !sm->checkPwdValidation(password, &pwd_errors, LanguageManager::getInstance()->getFreeCatalog(LANG_EN))) {
-	
+		Poco::Thread::sleep(ServerConfig::g_FakeLoginSleepTime);
 		result->set("state", "error");
 		result->set("msg", pwd_errors.getLastError()->getString(false));
 		if (pwd_errors.errorCount()) {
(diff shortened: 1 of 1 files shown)


login_server/src/cpp/JSONInterface/JsonUnsecureLogin.cpp
```diff
@@ -77,12 +77,10 @@ Poco::JSON::Object* JsonUnsecureLogin::handle(Poco::Dynamic::Var params)
 	Poco::JSON::Object* result = new Poco::JSON::Object;
 
 	if (!password.size() || !sm->checkPwdValidation(password, &pwd_errors, LanguageManager::getInstance()->getFreeCatalog(LANG_EN))) {
-	
+		Poco::Thread::sleep(ServerConfig::g_FakeLoginSleepTime);
```
ulfgebhardt, line 80: Why would we block a thread on our server when a user uses a wrong password? The user can actively block our server resources by entering wrong passwords?
Also as far as I understand it, this sleeps x and then answers the user. This does not make sense. It should be checked on request if the user has to wait and report to the user "hey you have to wait x till you can try again"

ulfgebhardt (changes requested):

login_server/src/cpp/JSONInterface/JsonUnsecureLogin.cpp
einhornimmond: Do you now what the sleep call is doing?
"Suspends the current thread for the specified amount of time. " (Q: https://pocoproject.org/docs/Poco.Thread.html)
While the thread is sleeping it don't run on a cpu so it don't use CPU ressources, it's sleeping.
I have testet it, it really don't use CPU Time while it is sleeping. 

If the password was checked with the normal way, that means calculating a argon2 hash it needs ServerConfig::g_FakeLoginSleepTime. 
On Startup a argon2 hash will be calculated and the duration will be measured and put into ServerConfig::g_FakeLoginSleepTime. 
So to give the user everytime the same experience I make it wait instead of early return. 
Else Hackers could find out which passwords the don't need to try at all. 
So that code has the reason to don't calculate Argon2 if the password is empty or has missing characters and cannot be correct but nevertheless return at the same time as if had calculated Argon2 hash. 
The same is happening if the email wasn't found.

einhornimmond:
Oh I see it wasn't done by usercheck, I have added it.

ulfgebhardt (approved):

PR #518, opened 2021-06-11
fix: Remove Target Date in Send Coins Request

einhornimmond:
## 🍰 Pullrequest
Allow to call sendCoins without target_date


ulfgebhardt:
This needs the corresponding frontend implementation - either make a new issue or solve this in this branch

Background: There is a date displayed on the transaction confirmation page which is not the same of the transaction resulting from it.

Mogge (approved):

PR #519, opened 2021-06-11
Change password require old password

### LINKED ISSUE #489: 🚀 [Feature] Change password requires old password
## 🚀 Feature

Change password requires old password

einhornimmond:
## 🍰 Pullrequest
Changing password with updateUserInfos require additional the old password 
with User.password_old 
Example:
```json 
{
	"session_id": -127182,
	"email": "max.musterman@gmail.de",
	"update": {
		"User.password": "1234",
		"User.password_old": "4321"
  	}
}
```

### Issues
- fixes #489


login_server/src/cpp/JSONInterface/JsonUpdateUserInfos.cpp
```diff
@@ -144,32 +144,75 @@ Poco::JSON::Object* JsonUpdateUserInfos::handle(Poco::Dynamic::Var params)
...
-							// 2 = password changed, only hash stored in db, couldn't load private key for re-encryption
-						case 2: jsonErrorsArray.add("password changed, couldn't load private key for re-encryption");  break;
-							// -1 = stored pubkey and private key didn't match
-						case -1: jsonErrorsArray.add("stored pubkey and private key didn't match"); break;
+						old_password_obj.convert(old_password);
+					}
+
+					bool old_password_valid = false;
+					NotificationList errors;
+					if (old_password.size())
```
ulfgebhardt, line 171: `old_password` is always an empty string since it was never set?!

login_server/src/cpp/JSONInterface/JsonUpdateUserInfos.cpp
```diff
@@ -144,32 +144,75 @@ Poco::JSON::Object* JsonUpdateUserInfos::handle(Poco::Dynamic::Var params)
...
+							jsonErrorsArray.add("User.password isn't valid");
+							jsonErrorsArray.add(errors.getErrorsArray());
+						}
+						else 
+						{
+							auto result_new_password = user->setNewPassword(value.toString());
+
+							switch (result_new_password) {
+								// 0 = new and current passwords are the same
+							case 0: jsonErrorsArray.add("new password is the same as old password"); break;
```
ulfgebhardt, line 204: This condition is not checked on pw reset. Why here?

ulfgebhardt (changes requested):

login_server/src/cpp/JSONInterface/JsonUpdateUserInfos.cpp
einhornimmond: three lines above it was set
```cpp
old_password_obj.convert(old_password);
````
old_password_obj is a json-object which has a method convert(T& varname)  which convert the content of the JSON-Object into the type of the given variable and fill it with that. 
The JSON lib Poco use isn't really intuitive I admit. 
Maybe I should use another one.

File: login_server/src/cpp/JSONInterface/JsonUpdateUserInfos.cpp (hunk 2 above)
Comment (einhornimmond) [reply login_server/src/cpp/JSONInterface/JsonUpdateUserInfos.cpp:186]: I don't know what you mean. 
Which condition?
What did you mean with pw reset?
This will be used on password reset and password change.

**Comment by einhornimmond** [comment]
I have added two more function requests here:
- Username Rules #532
- If updateUserInfos was called with existing data without changes #509

File: login_server/src/cpp/controller/User.cpp [hunk 3]
```diff
@@ -194,28 +194,38 @@ namespace controller {
...
+	}
 
+	int User::login(const std::string& password)
+	{
+		std::unique_lock<std::shared_mutex> _lock(mSharedMutex);
+
+		if (!mPassword.isNull() && mPassword->hasKey()) {
+			return 2;
+		}
+		assert(mPassword.isNull());
```
ulfgebhardt, line 222: This looks very fishy. You cannot assert password to be null since it could be not Null and `hasKey = false`?!
Why do you assert that the users password is null on login?

login_server/src/cpp/test/main.cpp
```diff
@@ -27,7 +30,7 @@ void fillTests()
 	//	gTests.push_back(new LoginTest());
 }
 
-void runMysql(std::string sqlQuery)
+int runMysql(std::string sqlQuery)
```
ulfgebhardt, line 33: bool instead of int as return type? You return exactly two values -1 an 0? Or is there a reason you choose int?

ulfgebhardt (commented):
Looks good so far, just have some questions on two occasions.

login_server/src/cpp/JSONInterface/JsonUpdateUserInfos.cpp
ulfgebhardt: the condition that you set the same password. My test case on reset password always involves the same password and that works without a problem. Here you force for a different password. I don't mind, but should there not be identical policies for passwords?

ulfgebhardt (approved):

PR #523, opened 2021-06-11
compare with last transaction sended

### LINKED ISSUE #495: 🐛 [Bug] Double booking for user
## 🐛 Bugreport

A user reports that she intended to send an amount of GDD. But the amount was sent twice.

> heute in der Montagskonferenz wurde gesagt, dass Überweisungen vom Konto Login neu (Beta) wieder funktionieren. Am 16.05.2021 hatte ich dann vom alten Konto versucht zu überweisen und dann wurde der Betrag zwei Mal abgebucht.

We need to evaluate when this can happen. Furthermore we need to figure out a mechanic how we can prevent this.

Common case of this would be a replay-"attack" where the user itself does the replay, by clicking twice

einhornimmond:
## 🍰 Pullrequest
- store last transaction in session (in memory)
- compare transaction with last sended transaction
- return error if last transaction are the same as new transaction only if new transaction is within 100 seconds of the old transaction

### Issues
- fixes #495

### DIFF
--- a/login_server/src/cpp/model/gradido/Transaction.cpp
+++ b/login_server/src/cpp/model/gradido/Transaction.cpp
@@ -630,6 +630,31 @@ namespace model {
 
 		}
 
+		bool Transaction::isTheSameTransaction(Poco::AutoPtr<Transaction> other)
+		{
+			bool result = false;
+
+			auto other_proto = other->getTransactionBody()->getBody();
+			auto other_created = other_proto->created();
+			auto own_body_bytes = getTransactionBody()->getBodyBytes();
+			auto own_body_updated = new proto::gradido::TransactionBody;
+			own_body_updated->ParseFromString(own_body_bytes);
+			auto own_created = own_body_updated->mutable_created();
+			Poco::Int64 timeDiff = other_created.seconds() - own_created->seconds();
+			*own_created = other_created;
+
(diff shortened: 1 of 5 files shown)


login_server/src/cpp/model/gradido/Transaction.cpp
```diff
@@ -630,6 +630,31 @@ namespace model {
...
+			*own_created = other_created;
+
+			result = own_body_updated->SerializeAsString() == other_proto->SerializeAsString();
+
+			delete own_body_updated;
+
+			// if they are more than 100 seconds between transaction they consider as not the same
+			if (abs(timeDiff) > 100) {
+				return false;
+			}
```
ulfgebhardt, line 653: 100 seconds is 1min 40sec. Make it 10 seconds

ulfgebhardt (changes requested):
This is a hackfix. The actual fix would be to generate a unique number for each transaction request and invalidate it once the transaction is sent. For now it is ok to do this hackfix. Longterm this is not the correct solution.

Please correct the 100seconds. Thats a bit much and can result in unwanted false positives.

login_server/src/cpp/model/gradido/Transaction.cpp
einhornimmond: yes sire

ulfgebhardt (approved):

PR #526, opened 2021-06-14
remove check decays for being at least 100 GDD cent

### LINKED ISSUE #439: 🐛 [Bug] 0,01 GDD missing and no Decay
## 🐛 Bugreport

Transaction list of new user:
- Recieved 10GDD
Its the only transaction, no decay transaction displayed

Balance of the Account: 9,99 GDD

This might be solved by the ceil issue described here: #434

einhornimmond:
- remove check decays for being at least 100 GDD cent
- remove debugging lines and empty lines

## 🍰 Pullrequest
For the reason see  #439

### Issues
- fixes #439

### Todo
- [ ] Test?

### DIFF
--- a/community_server/src/Model/Table/TransactionsTable.php
+++ b/community_server/src/Model/Table/TransactionsTable.php
@@ -178,25 +178,18 @@ class TransactionsTable extends Table
             }
             if($prev && $decay == true) 
             {
-                
                 if($prev->balance > 0) {
-                //    var_dump($stateUserTransactions);
                     $current = $su_transaction;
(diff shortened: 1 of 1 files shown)


ulfgebhardt (approved):
I assume we will still strip the decay values of less then 100 GDDcent from the list, right?

einhornimmond:
No, because that lead to the problem you have found. 
If user has get 100000 GDDcent and wait not much time, 
he has only 99999 GDDcent left, with floor it gives us 9,99 GDD so he expect to see 0,01 GDD Decay. 
But because there a really only 0.0001 GDD decay or 1 GDDcent this decay would be removed by my previous code and the user see no decay at all.

PR #551, opened 2021-06-15
hot fixes shown by test

einhornimmond:
## 🍰 Pullrequest

- show in between decays only if <= 0.01 GDD

### Issues

- None

### Todo

- [X] None


ulfgebhardt (approved):
The balance thing looks good - but you also remove some error reporting here. I assume this could cause the issues we experience. So I truest you that this is correct and approve

PR #556, opened 2021-06-15
feat: Trim Email on Blur

Mogge:
## 🍰 Pullrequest
Trims the email of the GDD send form after loosing focus.

### DIFF
--- a/frontend/src/views/Pages/AccountOverview/GddSend/TransactionForm.vue
+++ b/frontend/src/views/Pages/AccountOverview/GddSend/TransactionForm.vue
@@ -41,7 +41,7 @@
                     id="input-1"
                     v-model="form.email"
                     v-focus="emailFocused"
-                    @focus="emailFocused = !emailFocused"
+                    @focus="emailFocused = true"
                     @blur="normalizeEmail()"
                     type="email"
                     placeholder="E-Mail"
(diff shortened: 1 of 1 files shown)


frontend/src/views/Pages/AccountOverview/GddSend/TransactionForm.vue
```diff
@@ -186,6 +190,10 @@ export default {
       this.form.amountValue = Number(this.form.amount.replace(',', '.'))
       this.form.amount = this.$n(this.form.amountValue, 'decimal')
     },
+    normalizeEmail() {
+      this.emailFocused = !this.emailFocused
```
ulfgebhardt, line 194: This is a bit confusing regarding the naming - since you do not set the focus, just the variable. Is this really safe, cant we not just get the focus status instead of doing `@focus="emailFocused = !emailFocused"`

ulfgebhardt (changes requested):

frontend/src/views/Pages/AccountOverview/GddSend/TransactionForm.vue
Mogge: This is how the plugin works. You track the focus of an input field with a data variable.
https://www.npmjs.com/package/vue-focus#mixin

ulfgebhardt: And you are sure `@focus="emailFocused = !emailFocused"` is a good idea?
The linked package explicitly sets the focus: `template: '<input type="text" v-focus="focused" @focus="focused = true" @blur="focused = false">'`
Don't get me wrong - i approve this if you want - it just looks a bit dangerous.

ulfgebhardt (approved):

PR #561, opened 2021-06-15
fix: Change Password Form

Mogge:
## 🍰 Pullrequest
 * alerts removed
 * logic improved
 ^ remove API call on transaction list after login

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -212,7 +212,7 @@ jobs:
           report_name: Coverage Frontend
           type: lcov
           result_path: ./coverage/lcov.info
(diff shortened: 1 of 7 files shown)


Mogge:
@einhornimmond I cannot catch the error message from the server. So far I received data with a `msg`field which contains the error message. Now I receive an array of error messages. We need one standard solution to communicate errors to the client.

einhornimmond:
Ok, I additional put one error message into msg. But the array I like to use for easy debugging errors. 
After I fix the other things I look at that.

ulfgebhardt (approved):
<3

PR #567, opened 2021-06-16
fix: Thousend Dividers on GDD Send Amount Field

Mogge:
## 🍰 Pullrequest
Do not show the dividers of numbers > 1000. E.g show `12345.00` instead of `12,345.00`


ulfgebhardt (approved):
Not working with the regex as expected?

PR #576, opened 2021-06-21
update mysql because tuple has changed

### LINKED ISSUE #575: 🐛 [Bug] Show all users with not activated account don't work anymore
## 🐛 Bugreport
In Old Backend the function to show all users which don't have activated here Gradido Account don't work anymore since our last release.

einhornimmond:
## 🍰 Pullrequest
Error occure because model::table::UserTuple has changed. Description was added, so I must update the custom mysql query for the search also.

### Issues
- fixes #575

### DIFF
--- a/login_server/src/cpp/JSONInterface/JsonGetUsers.cpp
+++ b/login_server/src/cpp/JSONInterface/JsonGetUsers.cpp
@@ -93,10 +93,10 @@ Poco::JSON::Object* JsonGetUsers::handle(Poco::Dynamic::Var params)
 	result->set("state", "success");
 
 	//Poco::JSON::Object jsonResultObject;
-	Poco::JSON::Array  jsonUsersArray;
+	Poco::JSON::Array::Ptr jsonUsersArray = new Poco::JSON::Array;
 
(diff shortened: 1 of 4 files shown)


login_server/src/cpp/JSONInterface/JsonGetUsers.cpp
```diff
@@ -93,10 +93,10 @@ Poco::JSON::Object* JsonGetUsers::handle(Poco::Dynamic::Var params)
 	result->set("state", "success");
 
 	//Poco::JSON::Object jsonResultObject;
-	Poco::JSON::Array  jsonUsersArray;
+	Poco::JSON::Array::Ptr jsonUsersArray = new Poco::JSON::Array;
```
ulfgebhardt, line 96: Why change the type to Pointer? How is this related to the SQL problem?

ulfgebhardt (approved):

login_server/src/cpp/JSONInterface/JsonGetUsers.cpp
einhornimmond: It is not related to SQL Problems, but I need as a type Pointer for testing.

PR #598, opened 2021-06-28
Community fix

### LINKED ISSUE #569: 🚀 [Feature] Show Decay since last transaction
## 🚀 Feature

Show Decay since last transaction. The rest of the decay transaction will be hidden in the transaction themselfs. To have a valid balance the decay since the last transaction must be part of the list.

Pagination must take this into account.

### LINKED ISSUE #586: 🐛 [Bug] Transactions not shown due to error on Community Server
## 🐛 Bugreport

After the decay-in-transaction-update we have a notice in the result which obviously breaks the json. The frontend therefore does not show any transactions or balance.

`Notice (8): Undefined index: type [APP/Template/AppRequests/list_transactions.ctp, line 15]`

einhornimmond:
## 🍰 Pullrequest
- Fix Problem with decay inside transactions
- Update tests to reflect this new feature

### Issues
- fixes #586 
- fixes #569


community_server/src/Model/Table/TransactionsTable.php
```diff
@@ -194,7 +194,7 @@ public function listTransactionsHumanReadable($stateUserTransactions, array $use
                     
                     if($balance > 100) 
                     {
-                      $final_transactions['decay'] = [
```
ulfgebhardt, line 197: This was the error we have seen in the notice?

ulfgebhardt (commented):
I do not understand why we have error due to variable names - shouldn't we detect this automatically?
This does not solve the 26 elements on first page with a 25 pagination yet, right?
If you want to address this in another issue thats cool with me. Please take a look here: https://github.com/gradido/gradido/issues/569#issuecomment-869766776

community_server/src/Model/Table/TransactionsTable.php
einhornimmond: Yes this was the error.
Instead of putting the decay in the final_transaction the transaction current constructed in the loop, I put it into final_transactions which contain all the transactions
```php 
$final_transactions = [$final_transaction];
````
How should that be detected automatic?
That's php code not C/C++.  In C/C++ that couldn't never happen.

ulfgebhardt (approved):
<3 lets get this rolling

PR #603, opened 2021-07-01
refactor: Change Password

Mogge:
## 🍰 Pullrequest
dito


ulfgebhardt (approved):
You reduced the code alot here, do we have coverage improvement?

frontend/src/components/Inputs/InputPasswordConfirmation.vue
```diff
@@ -0,0 +1,68 @@
...
+  },
+  data() {
+    return {
+      password: '',
+      passwordRepeat: '',
+    }
+  },
+  computed: {
+    passwordObject() {
+      return { password: this.password, passwordRepeat: this.passwordRepeat }
```
ulfgebhardt, line 56: Here i would expect something like
```suggestion
      return { password: this.password, isRepeatCorrect: true/false }
```

ulfgebhardt (approved):

PR #611, opened 2021-07-06
feat: Test AccountOverview

Mogge:
## 🍰 Pullrequest
dito

### DIFF
--- a/frontend/src/views/Pages/AccountOverview.vue
+++ b/frontend/src/views/Pages/AccountOverview.vue
@@ -94,7 +94,7 @@ export default {
   },
   methods: {
     setTransaction(data) {
-      this.transactionData = data
+      this.transactionData = { ...data }
       this.currentTransactionStep = 1
     },
     async sendTransaction() {
--- a/frontend/src/views/Pages/AccountOverview.vue
+++ b/frontend/src/views/Pages/AccountOverview.vue
@@ -69,7 +69,7 @@ export default {
   data() {
     return {
       timestamp: Date.now(),
-      transactionData: EMPTY_TRANSACTION_DATA,
+      transactionData: { ...EMPTY_TRANSACTION_DATA },
       error: false,
       currentTransactionStep: 0,
       loading: false,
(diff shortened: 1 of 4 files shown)


frontend/src/views/Pages/AccountOverview.vue
```diff
@@ -69,7 +69,7 @@ export default {
   data() {
     return {
       timestamp: Date.now(),
-      transactionData: EMPTY_TRANSACTION_DATA,
+      transactionData: { ...EMPTY_TRANSACTION_DATA },
```
Mogge, line 72: @ulfgebhardt Do you remember when you said this would be the same as
```
transactionData: EMPTY_TRANSACTION_DATA,
```
Well, it is NOT!

frontend/src/views/Pages/AccountOverview.vue
```diff
@@ -110,7 +110,7 @@ export default {
       this.loading = false
     },
     onReset() {
-      this.transactionData = EMPTY_TRANSACTION_DATA
+      this.transactionData = { ...EMPTY_TRANSACTION_DATA }
```
Mogge, line 113: I am happy to have found this by testing!

frontend/src/views/Pages/AccountOverview.vue
ulfgebhardt: How so? isn't `...` unwinding the object so you can rewind it again wit `{}`?  Whats the difference?

frontend/src/views/Pages/AccountOverview.vue
ulfgebhardt: Explain, enlighten me aswell pls ❤️

ulfgebhardt (changes requested):
Explanation not sufficient, I feels stupid, can't approve

frontend/src/views/Pages/AccountOverview.vue
Mogge: @ulfgebhardt When you do `a = b` than `b` is the same as `a`. This means, when you change `b` you also change `a`. But when you say `a = { ...b }`, the content of `b` is copied to `a`. So when you change `b` it does not change `a`.

Mogge:
?

Mogge:
```
const a = { foo: 0, bar: 1 }
-> undefined
const b = a
-> undefined
b.foo = 1
-> 1
a
-> Object { foo: 1, bar: 1 }

const c = { ...a }
-> undefined
c.foo = 0
-> 0
a
Object { foo: 1, bar: 1 }
```

ulfgebhardt (approved):
Danke

Issue #616, opened 2021-07-08
🐛 [Bug] Description Cannot Be Empty


Mogge:
## 🐛 Bugreport
When calling `POST http://localhost/login_api/updateUserInfos` with 
```
User.description:  ""
User.first_name:  "Some"
User.last_name: "Name"
```
I get
```
errors:  [ "User.description is empty" ]
0: "User.description is empty"
msg: "User.description is empty"
state: "error"
valid_values: 1
```

**update**
same goes for `first_name` and `last_name`

einhornimmond:
Yes, my code assume if you set the field description you like to change it.
Hm... but if people like to clear the description? I don't consider this case.
Ok so I remove the empty check for description to allow clearing it.

PR #617, opened 2021-07-08
fix: Do Not Pass Userdata Description to Server

Mogge:
## 🍰 Pullrequest
The server does not accept an empty description for the user profile. As we do not use description so far, it is always empty. This means, that the user profile cannot be updated.

### Issues
- relates #616

### Todo
- [ ] Allow empty description on the server


ulfgebhardt (approved):

ulfgebhardt:
Please make a followup issue before this gets merged and barried

PR #618, opened 2021-07-08
Allow empty description

### LINKED ISSUE #616: 🐛 [Bug] Description Cannot Be Empty

einhornimmond:
## 🍰 Pullrequest
Allow empty description for updateUserInfos.
If in db a description entry exist and and empty description field will be transfered,
description in db will be emptied also!

### Issues
- fixes #616

### DIFF
--- a/login_server/src/cpp/JSONInterface/JsonUpdateUserInfos.cpp
+++ b/login_server/src/cpp/JSONInterface/JsonUpdateUserInfos.cpp
@@ -123,13 +123,18 @@ Poco::JSON::Object* JsonUpdateUserInfos::handle(Poco::Dynamic::Var params)
 				}
 			}
 			else if ("User.description" == name) {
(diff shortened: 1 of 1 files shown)


Mogge (changes requested):
Same problem with `first_name` and `last_name`

ulfgebhardt (approved):

einhornimmond:
I don't think it is a good idea todo that. 
If you set first_name as field in request and make it empty, the db entry for first_name will be deleted.
In which case user could want delete there first- or last name?

Mogge (approved):

PR #622, opened 2021-07-08
fix: Fix Infinite Loop in Transaction List

Mogge:
## 🍰 Pullrequest
use computed value instead data to avoid looping.


ogerly (approved):
nice

PR #640, opened 2021-07-15
check for null date

### LINKED ISSUE #639: 🐛 [Bug] if not start decay block is set, an error occure
## 🐛 Bugreport
If no start decay block is set, Model/Table/TransactionsTable.php throw a error by trying to get a Timestamp from a nullptr.

einhornimmond:
## 🍰 Pullrequest
Fix crash occuring if no decay block is set and community server tries to call getTimestamp on a nullptr. 

### Issues
- fixes #639

### DIFF
--- a/community_server/src/Model/Table/TransactionsTable.php
+++ b/community_server/src/Model/Table/TransactionsTable.php
@@ -168,7 +168,11 @@ class TransactionsTable extends Table
         if($decay_start_transaction->count()) {
             $decay_start_transaction_id = $decay_start_transaction->first()->id;
         }
-        $decay_start_time = $stateBalancesTable->getDecayStartDateCached()->getTimestamp();
+        $decay_start_date = $stateBalancesTable->getDecayStartDateCached();
+        $decay_start_time = 0;
+        if($decay_start_date) {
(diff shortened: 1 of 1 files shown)


community_server/src/Model/Table/TransactionsTable.php
```diff
@@ -206,7 +210,7 @@ public function listTransactionsHumanReadable($stateUserTransactions, array $use
                           'decay_start' => $calculated_decay['start_date'],
                           'decay_end' => $calculated_decay['end_date']
                       ]; 
-                      if($prev->transaction_id < $decay_start_transaction_id && 
+                      if($decay_start_time && $prev->transaction_id < $decay_start_transaction_id &&
```
Mogge, line 213: `$decay_start_time > 0` ? Or is `0 == false` in any case?

Mogge (approved):

community_server/src/Model/Table/TransactionsTable.php
einhornimmond: $decay_start_time is initalized with 0 and set at a actual value, if an decay block exist

PR #641, opened 2021-07-15
Login fix crash

einhornimmond:
## 🍰 Pullrequest
Fix bugs which let the login-server crash

- Change Pending Task structure from using virtual objects to combine different classes with the base class into on map, to using only direct the model::gradido::Transaction class. With that change, it isn't neccessary anymore to make dynamic casts of pointers which seems to lead to bugs with the Poco::AutoPtr. The Hedera Transactions are not longer needed so it is relative easy.
- Move all Destructors from mode::table classes from public to protected to prevent misuse of the table classes ignoring the reference counting (needed for Poco::AutoPtr)

### Issues
- depends on / includes  #640

### DIFF
--- a/community_server/src/Model/Table/TransactionsTable.php
+++ b/community_server/src/Model/Table/TransactionsTable.php
@@ -168,7 +168,11 @@ class TransactionsTable extends Table
         if($decay_start_transaction->count()) {
             $decay_start_transaction_id = $decay_start_transaction->first()->id;
         }
-        $decay_start_time = $stateBalancesTable->getDecayStartDateCached()->getTimestamp();
+        $decay_start_date = $stateBalancesTable->getDecayStartDateCached();
+        $decay_start_time = 0;
+        if($decay_start_date) {
+            $decay_start_time = $decay_start_date->getTimestamp();
+        }
         
         foreach($stateUserTransactions as $i => $su_transaction)
         {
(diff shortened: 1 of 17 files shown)


community_server/src/Model/Table/TransactionsTable.php
```diff
@@ -168,7 +168,11 @@ public function listTransactionsHumanReadable($stateUserTransactions, array $use
         if($decay_start_transaction->count()) {
             $decay_start_transaction_id = $decay_start_transaction->first()->id;
         }
-        $decay_start_time = $stateBalancesTable->getDecayStartDateCached()->getTimestamp();
+        $decay_start_date = $stateBalancesTable->getDecayStartDateCached();
+        $decay_start_time = 0;
+        if($decay_start_date) {
+            $decay_start_time = $decay_start_date->getTimestamp();
+        }
```
Mogge, line 175: Once again. This is another PR. Please try to have proper PRs.

community_server/src/Model/Table/TransactionsTable.php
```diff
@@ -206,7 +210,7 @@ public function listTransactionsHumanReadable($stateUserTransactions, array $use
                           'decay_start' => $calculated_decay['start_date'],
                           'decay_end' => $calculated_decay['end_date']
                       ]; 
-                      if($prev->transaction_id < $decay_start_transaction_id && 
+                      if($decay_start_time && $prev->transaction_id < $decay_start_transaction_id &&
```
Mogge, line 213: Other PR

Mogge (approved):

community_server/src/Model/Table/TransactionsTable.php
einhornimmond: I have like yesterday proposed put that in another branch and pr (#640) and merge it with that branch so that I can test my changes properly.

PR #699, opened 2021-08-05
Apollo sendCoins via Login-Server

einhornimmond:
## 🍰 Pullrequest
Use the createTransaction API Call from Login-Server for sendCoins directly rather than using Community-Server for tunneling,


Mogge (approved):

Mogge:
Geht dann alles nich wie vorher?

einhornimmond:
Depends on how the frontend handle the various errors. 
The login server handles missing fields errors slightly differently than the community server.
The format is the same but the text in msg and details are differ. 
As long as the input data are valid, nothing should be change at all.

PR #716, opened 2021-08-11
feat: Add Pagination Buttons to GDT Transaction List

Mogge:
## 🍰 Pullrequest
adds pagination buttons to GDT transaction list

### Todo

- [ ] test with more than 5 GDT transactions

### DIFF
--- a/frontend/src/views/Pages/AccountOverview/GddTransactionList.vue
+++ b/frontend/src/views/Pages/AccountOverview/GddTransactionList.vue
@@ -110,7 +110,6 @@ export default {
   data() {
     return {
       currentPage: 1,
-      startDecay: 0,
     }
   },
   props: {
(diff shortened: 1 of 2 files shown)


frontend/src/views/Pages/AccountOverview/GddTransactionList.vue
```diff
@@ -110,7 +110,6 @@ export default {
   data() {
     return {
       currentPage: 1,
-      startDecay: 0,
```
Elweyn, line 113: This was not used right? Or why is it only deleted here?

Elweyn (approved):

ulfgebhardt (dismissed):
Requires fixed https://github.com/gradido/gradido/issues/717

Issue #713, opened 2021-08-10
🔧 [Refactor] Get Rid of Warnings in Specs of UserProfileTransactionList


Mogge:
## 🔧 Refactor ticket
The test of UserProfileTransactionList throws warnings:
```
 [BootstrapVue warn]: observeDom: Requires MutationObserver support.
```
Find a way to eliminate them!

Elweyn:
@Mogge, @ulfgebhardt,
I will remove the window.alert('success message') and window.alert('error message') and replace it with either $toaster success or error method so we can make a proper test like we would like to have it when we use it.

Issue #553, opened 2021-06-15
💥 [DevOps] Enforce 60% converage (frontend)


ulfgebhardt:
## 💥 DevOps ticket

Enforce 60% coverage

ogerly:
![giphys](https://user-images.githubusercontent.com/1324583/122362309-511f2780-cf58-11eb-80b7-55be5f6a00d0.gif)

Elweyn:
@Mogge, @ulfgebhardt,
We need to transfer this issue to the new epic? 
Or is it done?

PR #737, opened 2021-08-17
feat: Testing Tabs of TransactionLists

Mogge:
## 🍰 Pullrequest
 * Test the Tabs of transaction lists (GDD / GDT)
 * disable warnings from bootstrap vue in test setup
 * test thx page
 * setup tests for GDT transaction list

 - fixes: #553 (Oh yeah!)
 - fixes: #713


frontend/yarn.lock
```diff
@@ -3750,9 +3750,9 @@ caniuse-api@^3.0.0:
     lodash.uniq "^4.5.0"
 
 caniuse-lite@^1.0.0, caniuse-lite@^1.0.30001109, caniuse-lite@^1.0.30001181:
-  version "1.0.30001191"
-  resolved "https://registry.yarnpkg.com/caniuse-lite/-/caniuse-lite-1.0.30001191.tgz#bacb432b6701f690c8c5f7c680166b9a9f0843d9"
-  integrity sha512-xJJqzyd+7GCJXkcoBiQ1GuxEiOBCLQ0aVW9HMekifZsAVGdj5eJ4mFB9fEhSHipq9IOk/QXFJUiIr9lZT+EsGw==
+  version "1.0.30001251"
+  resolved "https://registry.npmjs.org/caniuse-lite/-/caniuse-lite-1.0.30001251.tgz"
+  integrity sha512-HOe1r+9VkU4TFmnU70z+r7OLmtR+/chB1rdcJUeQlAinjEeb0cKL20tlAtOagNZhbrtLnCvV19B4FmF1rgzl6A==
```
Mogge, line 3755: Is this a problem? I was playing around with different packages to get rid of the warnings. Then I was asked to update my browser support for the tests:
```
npx browserslist@latest --update-db
```

frontend/test/testSetup.js
```diff
@@ -33,6 +33,9 @@ loadAllRules(i18nMock)
 
 global.localVue = createLocalVue()
 
+// switch of warnings from bootstrap vue
+global.process.env.BOOTSTRAP_VUE_NO_WARN = true
```
Elweyn, line 37: This is only for test cases!? Or do we need to have this in mind in case we get problems in the future?

frontend/src/views/Pages/AccountOverview/GdtTransactionList.spec.js
```diff
@@ -0,0 +1,115 @@
+import { mount } from '@vue/test-utils'
+import GdtTransactionList from './GdtTransactionList'
+
+const localVue = global.localVue
+
+const apolloMock = jest.fn().mockResolvedValue({
+  data: {
+    listGDTEntries: {
+      count: 4,
+      gdtEntries: [
```
Elweyn, line 10: I would set the gdtEntries in a own variable so we could Test if it is setten on the right variable or other Tests.
What do you think?

frontend/src/views/Pages/Login.vue
```diff
@@ -88,7 +88,7 @@ export default {
             email: this.form.email,
             password: this.form.password,
           },
-          fetchPolicy: 'no-cache',
+          fetchPolicy: 'network-only',
```
Elweyn, line 91: Do I understand it right network-only does update the cache so if the user don't have a internet connection can continue to work with the localy stored data!?

frontend/src/views/Pages/UserProfileTransactionList.spec.js
```diff
@@ -30,8 +20,12 @@ describe('UserProfileTransactionList', () => {
     },
   }
 
+  const stubs = {
```
Elweyn, line 23: Does this make that the underlying component is mocked!?

Elweyn (approved):

frontend/test/testSetup.js
Mogge: This is only for the test cases. We wiill have the same warnings when we use bootstrap vue's modals

frontend/src/views/Pages/AccountOverview/GdtTransactionList.spec.js
Mogge: We have three different types. They are all present and we have to test them all.

frontend/src/views/Pages/UserProfileTransactionList.spec.js
Mogge: It does not load the stubbed component. Instead it places a stub, on which you can emit events.

PR #773, opened 2021-08-23
decay with the value 0 is no longer displayed

### LINKED ISSUE #758: 🐛 [Bug] Decay of -0
## 🐛 Bugreport

Our GDD List shows a decay of -0 when it should say something like "No decay" or similar.

ogerly:
## 🍰 Pullrequest
decay with the value 0 is no longer displayed. 
the button is no longer displayed and there is no additional information about a transaction.

### Issues
- fixes #758
 
FireShot Capture 232 - Gradido Account - localhost

### DIFF
--- a/frontend/src/components/DecayInformation.vue
+++ b/frontend/src/components/DecayInformation.vue
@@ -1,7 +1,12 @@
 <template>
   <div>
     <span v-if="decaytyp === 'short'">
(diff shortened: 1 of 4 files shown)


ulfgebhardt:
What about those, where you have two outgoing or incoming transactions shortly after each other?

ulfgebhardt (approved):

Issue #709, opened 2021-08-07
🚀 [Feature] Shield Middleware


ulfgebhardt:
## 🚀 Feature

Implement Shield Middleware in Apollo Server

PR #777, opened 2021-08-26
feat: JSON Web Token for Authentification

### LINKED ISSUE #13: 🔧 [Refactor] Authentication handled by the backend
## :zap: Refactor ticket

It is part of http standard to set cookies from the backend and handle all authentication via this retransmitted cookie. This is not the case for us as far as I understand it. Therefore it will not be possible to safely remember a user. This might be by design?

### Motive

Use web standards

### LINKED ISSUE #213: 🚀 [Feature] Support Bearer Header & JWT
## :rocket: Feature

We currently have two options to signal our login status:
- via cookie
- via session_id as parameter

While the cookie is a standard way of doing authentication it is limited to one domain. Either load common external sources and use the external domain or use a different mechanic to authenticate. With the login server intended to be shared between different communities each with their own domain, the cookie mechanic does not seem to be a good choice.

The session_id is quite simplistic and pollutes the parameters of a call.

Suggestion: Use of bearer headers. Its the standard, it does not pollute the parameters and is basically invisible for the application. Its easy to write a hook for every call  Furthermore the standard of asymetric encryption would be good - namely JWT (Jason Web Token)

See:
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication
- Especially: https://tools.ietf.org/html/rfc6750
- For JWT: https://jwt.io/introduction

Mogge:
## 🍰 Pullrequest
setup jwt for authorization

fixes #213 
fixes #13 
makes #709 obsolete


Mogge (commented):
??

backend/src/auth/auth.ts
```diff
@@ -0,0 +1,21 @@
...
+    const decoded = decode(context.token)
+    if (decoded.sessionId && decoded.sessionId !== 0) {
+      const result = await apiGet(
+        `${CONFIG.LOGIN_API_URL}checkSessionState?session_id=${decoded.sessionId}`,
+      )
+      context.sessionId = decoded.sessionId
+      return result.success
+    }
+  }
+  return false
```
ulfgebhardt, line 20: I would expect this function to throw an Unauthorized exception instead of returning false or is this done elsewhere?

frontend/src/main.js
```diff
@@ -3,15 +3,29 @@ import DashboardPlugin from './plugins/dashboard-plugin'
...
 
 import router from './routes/router'
 
+const httpLink = new HttpLink({ uri: CONFIG.GRAPHQL_URI })
+
+const authLink = new ApolloLink((operation, forward) => {
+  const token = store.state.token
+  operation.setContext({
+    headers: {
+      Authorization: token && token.length > 0 ? `Bearer ${token}` : '',
```
ulfgebhardt, line 20: Why `token.length > 0` why can this happen? Shouldn't we either have `null` or some token string?

ulfgebhardt (commented):
Great work. I think we are almost there - the only thing is: What about the error case? Session ran out, token is invalid? Do we properly handle this? (Serverside would be sufficient for now, clientside can be pushed to a later release)

backend/src/auth/auth.ts
Mogge: The Not Authenticated Errors is raised by the typegraphql @Authorized decorator

frontend/src/main.js
Mogge: Just to make sure that token can not be an empty string (should be valid or null)

ulfgebhardt (approved):

PR #782, opened 2021-08-31
Fix missing gdt id

### LINKED ISSUE #740: 🐛 [Bug] GDT Transaction ID is missing
## 🐛 Bugreport

GDT Transaction ID is missing

We get null values for GDT transaction ids (or so I understood). This made us switch from a direct GDT-Connection to the Community-Server-Tunnel (why? @Mogge ).
Apparently we get null values for the transaction ids, which are used as unique identifier for a GDT-Transaction analog to the GDD-Transactions

einhornimmond:
## 🍰 Pullrequest
- I have fixed the id in gdt server and add id for gdt in backend and frontend. 
- I fixed also a bug with icons in frontend, which lead to not showing my gdt transactions at all

### Issues
- fixes #740

### DIFF
--- a/backend/src/graphql/models/GdtEntry.ts
+++ b/backend/src/graphql/models/GdtEntry.ts
@@ -15,6 +15,7 @@ export enum GdtEntryType {
 @ObjectType()
 export class GdtEntry {
   constructor(json: any) {
+    this.id = json.id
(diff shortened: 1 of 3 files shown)


ulfgebhardt (approved):
Looks good!

PR #783, opened 2021-08-31
update scripts and doc for login-server setup without docker

einhornimmond:
## 🍰 Pullrequest
Describe how to setup Login-Server on Linux Bare-Metal.
Test and fix scripts necessary for that.


.gitmodules
```diff
@@ -31,3 +31,6 @@
 [submodule "login_server/src/proto"]
         path = login_server/src/proto
         url =  https://github.com/gradido/gradido_protocol.git
+[submodule "login_server/dependencies/protobuf"]
+	path = login_server/dependencies/protobuf
+	url = https://github.com/protocolbuffers/protobuf.git
```
ulfgebhardt, line 36: Didn't we remove this just lately?

login_server/README.md
```diff
@@ -0,0 +1,147 @@
...
+install build essentials 
+
+```bash
+sudo apt install -y gcovr build-essential gettext libcurl4-openssl-dev libssl-dev libsodium-dev libboost-dev
+``` 
+
+## CMake
+I use CMake for build file generation and the Login-Server needs at least version v3.18.2
+You can build and install it from source. 
+The Version in apt is sadly to old.
```
ulfgebhardt, line 12:
```suggestion
CMake is used for build file generation and the Login-Server needs at least version v3.18.2
You can build and install it from source. 
The Version in apt is sadly to old.
```

.gitmodules
einhornimmond: Maybe but without Login-Server don't work.

ulfgebhardt (approved):

PR #796, opened 2021-09-08
different path for checkEmail and resetPassword

einhornimmond:
## 🍰 Pullrequest
- add additional path (url in new frontend) for resetPassword email


Mogge (approved):
I approve. We will have to test the new email link.

PR #811, opened 2021-09-15
webpack update

ulfgebhardt:
## 🍰 Pullrequest

update webpack so it works with node 16

### Issues

- None

### How 2 Test

- [ ] Does this work with your node versions and environments?


einhornimmond (approved):
Is working on staging

PR #831, opened 2021-09-17
Implement getBalance in Apollo

### LINKED ISSUE #820: 🔧 [Refactor] Implement Community Call getBalance on Apollo
## 🔧 Refactor ticket

Implement Community Call getBalance on Apollo

einhornimmond:
## 🍰 Pullrequest
- use typeorm for direct access to saved balance in db
- create typeorm models
- add decay function to apollo
- 
### Issues
- fixes #820

### Todo
- [x] Create user typeorm entity
- [x] Create balance typeorm entity
- [x] Add decay calculation function (maybe in util folder?)
- [x] Update getBalance Resolver code
- [x] merge #881

### Why I need the state_user table for that?
state_balances use the state_user_id as foreign key, so first I must get the user public key from login-server, read from state_users with that and than use id for reading the correct balance entry from db.


ulfgebhardt:
@einhornimmond Wollen wir das nicht in Zukunft so machen? https://stackoverflow.com/questions/56311203/can-i-map-entity-field-names-to-alias-column-names-in-typeorm

einhornimmond:
Yes, great. I have searched for something like that but not found.

backend/src/graphql/resolvers/BalanceResolver.ts
```diff
@@ -5,14 +5,30 @@ import { Resolver, Query, Ctx, Authorized } from 'type-graphql'
...
+import { roundFloorFrom4 } from '../../util/round'
 
 @Resolver()
 export class BalanceResolver {
   @Authorized()
   @Query(() => Balance)
   async balance(@Ctx() context: any): Promise<Balance> {
-    const result = await apiGet(CONFIG.COMMUNITY_API_URL + 'getBalance/' + context.sessionId)
+    // get public key for current logged in user
+    const result = await apiGet(CONFIG.LOGIN_API_URL + 'login?session_id=' + context.sessionId)
```
ulfgebhardt, line 19: Why do we login? The user is logged in already. He has a sessionID?

backend/src/index.ts
```diff
@@ -48,7 +49,11 @@ async function main() {
     throw new Error(`Wrong database version - the backend requires '${DB_VERSION}'`)
   }
 
-  // const connection = await createConnection()
+  const toCon = await typeOrmConnection()
```
ulfgebhardt, line 52: This connects? You have `// await con.connect()` in that function.

backend/src/typeorm/entity/Balance.ts
```diff
@@ -0,0 +1,25 @@
...
+  @Column({ name: 'record_date', type: 'datetime' })
+  recordDate: Date
+
+  @Column({ type: 'bigint' })
+  amount: number
+
+  static findByUser(userId: number): Promise<Balance> {
+    return this.createQueryBuilder('balance')
+      .where('balance.userId = :userId', { userId })
+      .getOneOrFail()
```
ulfgebhardt, line 23: Please do not define queries in the model. I think we should separate queries and models. You can put the query in its own file if you intend to reuse it. eg. `queries/balance/findByUser.ts`

backend/src/graphql/resolvers/BalanceResolver.ts
```diff
@@ -5,14 +5,30 @@ import { Resolver, Query, Ctx, Authorized } from 'type-graphql'
...
     if (!result.success) throw new Error(result.data)
-    return new Balance(result.data)
+
+    // load user and balance
+    const userEntity = await tUser.findByPubkeyHex(result.data.user.public_hex)
```
ulfgebhardt, line 23: Instead of findByHex you could also use the email here. Where is the benefit?

ulfgebhardt (changes requested):
Great - tho the balance call is way less code then expected. This is a great step!

backend/src/graphql/resolvers/BalanceResolver.ts
einhornimmond: The JWT Token store only the session_id so I need to call to login-server to get the public-key from the user.
The login call is for returning login-data, the user and how many transactions for the user are pending or waiting on signing. 
It is used by community-server as well and for that I have written this function long time ago.
In my previous model login via api wasn't allowed so I could use the name here. The Class for that is called JsonGetLogin.

If the JWT token would store the public key of the user, I wouldn't need calling the login-server at all in this function. 
This login function from login-server is fast because it didn't call the db. It getting the data 100% from memory. 
It would be even faster with my replaced json pr #625

backend/src/index.ts
einhornimmond: If I call await con.connect() I get the error that the connection cannot be established because the connection already exist. 
But I have copied the code with that line from doc.

backend/src/typeorm/entity/Balance.ts
einhornimmond: So you rather like to use the Data Mapper Pattern over the Active Record pattern?
https://typeorm.io/#/active-record-data-mapper

backend/src/graphql/resolvers/BalanceResolver.ts
einhornimmond: > Great - tho the balance call is way less code then expected. This is a great step!

Yes but this has to do with ignoring a big peace of code entirely. 
I don't use the auto-error-correction code from community-server. It seems that the decay calculation is now stable so we maybe don't need at any longer. But it could lead to errors in future.

einhornimmond:
> Great - tho the balance call is way less code then expected. This is a great step!

Yes but this has to do with ignoring a big peace of code entirely. 
I don't use the auto-error-correction code from community-server. It seems that the decay calculation is now stable so we maybe don't need at any longer. But it could lead to errors in future.

backend/src/graphql/resolvers/BalanceResolver.ts
einhornimmond: At the beginning sometimes the key-pair from users changed and so multiple users with the same email exist. 
The transactions themself ultimately are based on the public key not on the email. 
So it is more robust that way. 
And it prevent data loss when a key-pair from a user change. 
But it isn't ideal we maybe can optimize it further.

backend/src/typeorm/entity/Balance.ts
einhornimmond: Or do you like it more if I go for your proposal and ignore the typeorm standards?

backend/src/graphql/resolvers/BalanceResolver.ts
ulfgebhardt: But we need the SessionID to authenticate against the Server, the publicKey is not allowing this.
Would it be possible to use `getUserInfo` instead?, since login can have unintended sideeffects + its not obvious what happens or is `getUserInfo` not providing the required data?

backend/src/index.ts
ulfgebhardt: Can you use a singleton pattern here to capsule the whole TypeORM connection mechanic handling the case of double connect? Consider this a suggestion, please do your research if its the right cause of action: 
https://stackoverflow.com/questions/50656921/typeorm-create-connection-globally
https://blog.bitsrc.io/the-singleton-pattern-in-typescript-b906303fda93

backend/src/typeorm/entity/Balance.ts
ulfgebhardt: I tried to find this, but couldnt when i scrolled over the docu of TypeORM, thanks ❤️ !

I think its a thing we could discuss with @Mogge and @Elweyn .
The documentation itself gives us a hint aswell - maintainability vs simplicity. We just have to decide what we want. It might be wise to decide this now to prevent larger refactors later.

image

And ofc we will use the standard, I just could not find it in the docu!

backend/src/graphql/resolvers/BalanceResolver.ts
einhornimmond: For `getUserInfo` I need email but this isn't saved in JWT token either. 
The call /login which could be also translated as getDetailsForSession deliver the user object and how many transactions are pending. This call is also used from community server and therefore is good testet and hasn't any side effects. 
But I agree that the name is not quite accurate. Should I change it?

ulfgebhardt: No don't change it. I would have requested to remove the email requirement from getUserInfo anyways - we will refactor this later. Keep it as it is!

backend/src/index.ts
einhornimmond: Does that mean the main function is called more than once?
I though if I put the creation connection call in the main function it will be called only once at startup.

ulfgebhardt: That is true, but it might be easier to collect the connection object where its needed by typing
```
import connection from .../connection

const con = connection.getInstance()
con.queryStuff()
con.queryMoreStuff()
```

instead of managing the connection somewhere in the main function. Or does TypeORM that for us when utilizing their models / query builders?

einhornimmond: typeorm use intern a connection pool and you can get a connection with getConnection.
Entities based on BaseEntity have there own direct access to connections.
For example I can call find functions direct on the entities and typeorm handle the rest.

ulfgebhardt: Awesome, then my point is obsolete. Just keep doing <3

ulfgebhardt (approved):

PR #864, opened 2021-09-20
docker skeema: mount sql files folder dynamic

einhornimmond:
## 🍰 Pullrequest
Mount sql files folder which skeema is checking for db schema updates dynamic instead of copy it in build. 
So it isn't any longer necessary to rebuild skeema after login-server db was changed. 

The .skeema files are config files necessary for skeema to work.

### DIFF
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -140,7 +140,7 @@ services:
     build:
       context: .
       dockerfile: ./skeema/Dockerfile
-      target: skeema_run
+      target: skeema_dev_run
     depends_on:
       - mariadb
     networks: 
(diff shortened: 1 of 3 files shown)


ulfgebhardt (commented):
This looks good. Be aware that this is only in test environment. Did you try Production aswell?
`docker-compose -f docker-compose.yml up --build`
In production you do not bind the volume correctly. My question is: Does the docker file still work with that case?

einhornimmond:
Oh no you are right, I have tried production docker

ulfgebhardt (approved):
This looks perfect now <3

PR #865, opened 2021-09-20
disable auto-resend of not finished transactions on login-server restart

einhornimmond:
### Problem
Login-Server try to resend not finished transaction on restart again.
The reason from the start was that no transactions get's lost by server crash. 
But in production this lead to creation transaction which are invalid because the user has already get enough Gradidos,
resend every time if login-server is startet new. And therefore send again the error message confusing the creator.

### Solution
Make it optional with additional option in config with default false.

### Additional
Removed double code which seems to resulting from some merge or rebase in past

### DIFF
--- a/login_server/src/cpp/ServerConfig.cpp
+++ b/login_server/src/cpp/ServerConfig.cpp
@@ -57,6 +57,7 @@ namespace ServerConfig {
 	int         g_FakeLoginSleepTime = 820;
 	std::string g_versionString = "";
 	bool		g_disableEmail = false;
+	bool	    g_resendTransactionOnStart = false;
 	ServerSetupType g_ServerSetupType = SERVER_TYPE_PRODUCTION;
 	std::string g_devDefaultGroup = "";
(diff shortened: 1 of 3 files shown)


login_server/src/cpp/ServerConfig.cpp
```diff
@@ -57,6 +57,7 @@ namespace ServerConfig {
 	int         g_FakeLoginSleepTime = 820;
 	std::string g_versionString = "";
 	bool		g_disableEmail = false;
+	bool	    g_resendTransactionOnStart = false;
```
ulfgebhardt, line 60: I suggest to have the name equal to the config value just to avoid confusion.

```suggestion
	bool	    g_resendUnfinishedTransactionOnStart = false;
```

einhornimmond: Ich wollte nicht so einen langen Variablennamen haben, aber du hast recht.

ulfgebhardt (approved):

PR #871, opened 2021-09-21
fix  database not reading config

### LINKED ISSUE #824: 🐛 [Bug] Database Migrations do not recognize .env values
## 🐛 Bugreport

The `.env` values are not correctly read on the migration tool. We had to insert password, username and database name in the code to get this working.

We encountered this on production where the default values root+nopw are not applicable.

We encountered this in production mode (js not ts) using the command `npm run up`

ulfgebhardt:
## 🍰 Pullrequest

Fixed database tool to read .env config when built.

Problem was the change of working directory when running the built version. Therefore the .env was searched for in `build/` instead of the root directory of the project

### Issues

- fixes #824 

### Todo

- [X] None

### DIFF
--- a/database/.env.dist
+++ b/database/.env.dist
@@ -4,4 +4,7 @@ DB_USER=root
 DB_PASSWORD=
 DB_DATABASE=gradido_community
 MIGRATIONS_TABLE=migrations
-MIGRATIONS_DIRECTORY=./migrations/
+// This value is set to the default for the built version.
+// Therefore you need to overwrite this when using the dev version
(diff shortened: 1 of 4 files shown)


ulfgebhardt:
@einhornimmond THis patch has the ownside that the `.env` has to have different values for production and dev version. I think this could be a potential pitfall for developers. Would you agree to have a separate .env file in the `build/ ` folder instead? This would fix the issue aswell, but the `.env` files could be the same in dev and production.

einhornimmond (approved):

ulfgebhardt:
I merge this

PR #878, opened 2021-09-22
[WIP] Registration in different communities

### LINKED ISSUE #803: 🚀 [Feature] community selection page as 1st page of registration process
## 🚀 Feature

1. new page to select a community
2. shows a single selectable list with the names of all communities
3. the names of all communities and their registration URL has to be requested per service invocation at the community instanz ( see community creation process)
4. shows an Abbruch and Weiter Button
5. Weiter Button is enabled as soon a community is selected
6. optional: a given PID parameter check per request invocation against the GDT-Server, returns a valid PID - the given one or the default PID of the Förderkreis
7. pressing the Weiter button invokes the registration process on the selected community incuding the mandatory PID

ogerly:
## 🍰 Pullrequest
Registration in different communities. 

there are currently two newly created files in the app. 
1. an overview page for the current community with redirection to the current registration page.
2. an overview page to select another community. 

existing files that have changed. 
3. login.vue
4. register.vue
5. ../locales/en.json
6. ../locales/en.json

the stimulation of the federation 
7. json file : frontend/public/json-example/communities.json

### Issues
- fixes #803 
 
 frontend/src/views/Pages/Login.vue
FireShot Capture 538 - Gradido Account - localhost
 
frontend/src/views/Pages/RegisterCommunity.vue
FireShot Capture 539 - Gradido Account - localhost

frontend/src/views/Pages/RegisterSelectCommunity.vue
FireShot Capture 540 - Gradido Account - localhost

frontend/src/views/Pages/Register.vue
FireShot Capture 541 - Gradido Account - localhost

### DIFF
--- a/frontend/public/json-example/communitys.json
+++ b/frontend/public/json-example/communitys.json
@@ -1,0 +1,7 @@
+{
+  "community": [
+    {"id": 1, "name": "Community 1", "location": "Ort 1", "description":"description 1 description 1 description 1 description 1 description 1 description 1 description 1 ", "url": "http://localhost:3000/vue/"},
+    {"id": 2, "name": "Community 2", "location": "Ort 2", "description":"description 2 description 2 description 2 description 2 description 2 description 2 description 2 ", "url": "https://stage1.gradido.net/vue/"},
+    {"id": 3, "name": "Community 3", "location": "Ort 3", "description":"description 3 description 3 description 3 description 3 description 3 description 3 description 3 ", "url": "https://gdd1.gradido.net/vue/"}
+  ]
+}
--- a/frontend/src/store/store.js
+++ b/frontend/src/store/store.js
@@ -68,6 +68,14 @@ export const store = new Vuex.Store({
     description: '',
     token: null,
     newsletterState: null,
+    community_name: 'Community 1',
+    community: {
+      name: 'Community 1',
+      name2: 'Community 1',
+      location: 'Ort 1',
+      description:
+        'description 1 description 1 description 1 description 1 description 1 description 1',
+    },
   },
   getters: {},
   // Syncronous mutation of the state
--- a/frontend/src/views/Pages/Login.vue
+++ b/frontend/src/views/Pages/Login.vue
@@ -21,11 +21,11 @@
               <div class="text-center text-muted mb-4">
                 <small>{{ $t('login') }}</small>
               </div>
-                <div class="text-center text-muted mb-4">
-                 {{ $store.state.community_name }}
+              <div class="text-center text-muted mb-4">
+                {{ $store.state.community_name }}
               </div>
               <div class="text-center text-muted mb-4">
-                 {{  $store.state.community.description }}
+                {{ $store.state.community.description }}
               </div>
               <validation-observer ref="observer" v-slot="{ handleSubmit }">
                 <b-form @submit.stop.prevent="handleSubmit(onSubmit)">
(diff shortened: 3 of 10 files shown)


frontend/public/json-example/communitys.json
```diff
@@ -0,0 +1,7 @@
+{
+  "community": [
+    {"id": 1, "name": "Community 1", "location": "Ort 1", "description":"description 1 description 1 description 1 description 1 description 1 description 1 description 1 ", "url": "http://localhost:3000/vue/"},
```
Elweyn, line 3: I asked myself if we want to put this in an apollo call so we have apollo responsible for this.
What do you think?

frontend/src/store/store.js
```diff
@@ -68,6 +68,14 @@ export const store = new Vuex.Store({
     description: '',
     token: null,
     newsletterState: null,
+    community_name: 'Community 1',
+    community: {
```
Elweyn, line 72: We want this from the .env or I am mistaken here?

frontend/src/views/Pages/Login.vue
```diff
@@ -20,6 +20,12 @@
             <b-card-body class="p-4">
               <div class="text-center text-muted mb-4">
                 <small>{{ $t('login') }}</small>
+              </div>
+                <div class="text-center text-muted mb-4">
+                 {{ $store.state.community_name }}
```
Elweyn, line 25: Why do we define this so? The name would also be in $store.state.community.name. 
I don't really like names like community_name

frontend/src/views/Pages/RegisterSelectCommunity.vue
```diff
@@ -0,0 +1,55 @@
+<template>
+  <div id="register-select-community">
+    <b-container class="justify-content-center text-center">
+      <div class="mb-4">{{ $t('community.current-community') }}</div>
+
+      <div v-for="community in communitys.community" :key="community.id" class="pb-3">
+        <b-card
+          v-show="community.name === $store.state.community_name"
```
Elweyn, line 8: Or have you instored $store.state.community_name for this scenario?

frontend/src/store/store.js
ogerly: yes, this should probably be in the env file, but since i don't have any exact details and this was only touched on in the conversation yesterday ... i'm waiting for an exact specification. how and where?

frontend/public/json-example/communitys.json
ogerly: that should be decided by @ulfgebhardt

frontend/src/views/Pages/Login.vue
ogerly: There were serious problems with me. For some inexplicable reason, it was not possible to retrieve the name, or it was incorrect. 
Since it will come from the ENV or CONFG anyway, there will no longer be any variables at this point, will there?

frontend/src/views/Pages/RegisterSelectCommunity.vue
ogerly: no, as already written, the name and the other community data information will come from the ENV or CONFIG.

frontend/public/json-example/communitys.json
ulfgebhardt: We can put this into an apollo call, but since we dont have any other data to feed to the client - just static stuff i thought it would simplify things if we just keep it client till we have the infrastructure to serve an actual list from a p2p-net (or blockchain for that matter).
Tho you are right. In the original Issue of claus-peter he defines that we should implement a server call + a database table for this data.
I guess you decide or we do an refinement

ogerly:
new PR https://github.com/gradido/gradido/pull/918

PR #889, opened 2021-09-23
feat: Redirect to Login when JWT has expired

### LINKED ISSUE #827: 🚀 [Feature] Handle JWT Error Cases
## 🚀 Feature

Currently we still get an error Displayed when the session expires. The whole UI is still rendered unfunctional.

We need to handle an expired JWT in the frontend. We should toast a red message "Your session ran out and is no longer valid" and redirect to the login page.

Consider the following: The Server does the validation of the JWT Token. In case an error occurs
- JWT Expired
- JWT is undecodable
- any other reason
the Server sends a status 403 - not authorized header instead of 200.

On the client side we catch a 403 status and trigger the mechanic described above.

This way the client does not need to decode the token. The whole token logic is still on the server.

Mogge:
## 🍰 Pullrequest
Server raises error with message: `'403.13 - Client certificate revoked'` when JWT has expired.
Frontend checks each response for this error. When it is present, redirect to login. The message is translated and toasted by the apollo call.

### Issues
- fixes #827

### DIFF
--- a/frontend/src/main.js
+++ b/frontend/src/main.js
@@ -21,6 +21,12 @@ const authLink = new ApolloLink((operation, forward) => {
     },
   })
   return forward(operation).map((response) => {
+    if (response.errors && response.errors[0].message === '403.13 - Client certificate revoked') {
+      response.errors[0].message = i18n.t('error.session-expired')
+      store.dispatch('logout', null)
+      if (router.currentRoute.path !== '/login') router.push('/login')
+      return response
+    }
     const newToken = operation.getContext().response.headers.get('token')
     if (newToken) store.commit('token', newToken)
     return response
--- a/backend/src/jwt/decode.ts
+++ b/backend/src/jwt/decode.ts
@@ -5,7 +5,7 @@ import jwt from 'jsonwebtoken'
 import CONFIG from '../config/'
 
 export default (token: string): any => {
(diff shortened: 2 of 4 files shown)


frontend/src/main.js
```diff
@@ -21,6 +21,12 @@ const authLink = new ApolloLink((operation, forward) => {
     },
   })
   return forward(operation).map((response) => {
+    if (response.errors && response.errors[0].message === '403.13 - Client certificate revoked') {
+      response.errors[0].message = i18n.t('error.session-expired')
+      store.dispatch('logout', null)
+      if (router.currentRoute.path !== '/login') router.push('/login')
+      return response
+    }
```
ulfgebhardt, line 29: Whats with `if (!token) return new Error('401 Unauthorized')` ?

Mogge: This one is catched by type-graphql:
` "message": "Access denied! You need to be authorized to perform this action!"`
Do you want to catch this as well to localize it?

ulfgebhardt: Its more about you not redirecting in that case.
If the client navigates to a protected page with a broken token, what happens then? Will he be redirected to login already cause of reasons?

ulfgebhardt (approved):

PR #891, opened 2021-09-23
Apollo list gdt transactions

### LINKED ISSUE #821: 🔧 [Refactor] Implement Community Call listGDTTransactions on Apollo
## 🔧 Refactor ticket

Implement Community Call listGDTTransactions on Apollo

einhornimmond:
## 🍰 Pullrequest
On top of apollo get balance (#831) so better merge that before
Simply get email address from db and then call api call direct from gdt server, the community-server call is only a proxy to.

### Issues
- fixes #821

### DIFF
--- a/backend/src/graphql/resolvers/GdtResolver.ts
+++ b/backend/src/graphql/resolvers/GdtResolver.ts
@@ -6,6 +6,7 @@ import CONFIG from '../../config'
 import { GdtEntryList } from '../models/GdtEntryList'
 import { GdtTransactionSessionIdInput } from '../inputs/GdtInputs'
 import { apiGet } from '../../apis/HttpRequest'
+import { User as tUser } from '../../typeorm/entity/User'
 
 @Resolver()
 export class GdtResolver {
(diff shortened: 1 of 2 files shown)


backend/src/graphql/resolvers/GdtResolver.ts
```diff
@@ -6,6 +6,7 @@ import CONFIG from '../../config'
 import { GdtEntryList } from '../models/GdtEntryList'
 import { GdtTransactionSessionIdInput } from '../inputs/GdtInputs'
 import { apiGet } from '../../apis/HttpRequest'
+import { User as tUser } from '../../typeorm/entity/User'
```
ulfgebhardt, line 9: @Mogge and I discussed that we would like the convention `dbModel`. What do you think about that? Instead of `tUser` `dbUser` to make it clear when reading?

ulfgebhardt (commented):
Besides one convention we might wanna change things look great!

backend/src/graphql/resolvers/GdtResolver.ts
einhornimmond: yes make sense, better that t from typeorm, short and clear :)

ulfgebhardt (approved):

PR #894, opened 2021-09-24
Db check with typeorm

### LINKED ISSUE #874: 🔧 [Refactor] Refactor database check in backend to TypeORM model
## 🔧 Refactor ticket

Refactor database check in backend to TypeORM model. Here is to consider if we can actually map the table correctly since it does not define a primary key (make issue on repo + also make issue for .ts/js fileending).

This way we can remove the plain SQL queries and the mysql2 package. Things will become more readable aswell

ulfgebhardt:
## 🍰 Pullrequest

- convert db check from plain sql to typeORM
- clean & refactor `index.ts`

### Issues

- fixes #874

### Todo

- [X] None


backend/src/typeorm/getDBVersion.ts
```diff
@@ -0,0 +1,15 @@
...
+
+const getDBVersion = async (): Promise<string | null> => {
+  const connection = getConnection()
+  const migrations = connection.getRepository(Migration)
+  try {
+    const dbVersion = await migrations.findOne({ order: { version: 'DESC' } })
+    return dbVersion ? dbVersion.fileName : null
+  } catch (error) {
+    return null
+  }
```
ulfgebhardt, line 12: @einhornimmond @Mogge Here you can see how the DataMapper Patters is accessing Data. The Queries are defined externally and not in the Model. For this case another way would be counter intuitive for me anyways, since this is a special case of "getLastColumnOrderedByDBVersionDesc" - doesn't sound like i'd like to have this function on the model to me.

einhornimmond (approved):

backend/src/typeorm/getDBVersion.ts
einhornimmond: As far as I understand you don't use the DataMapper Pattern. 
You use the repository directly in your function. If this case is used only at one place, nothing is to say against it.

einhornimmond (approved):

Issue #98, opened 2021-03-29
🚀 Multiple Addresses 


einhornimmond:
## :rocket: Feature
Multiple Gradido Addresses for every user. 
Only at one Address per user creation is possible, therefore we need a way to check on blockchain-level for which address creation is allowed. 

### Needed for
- Equalization and Environmental Fund
- Public Budget
- GDT 
- multiple possible E-Mails and Usernames per user as sender and receiver for transactions

### User Problem
- Users can separate private from business.
- More privacy by using different addresses.

### Implementation
Maybe with HD-Wallet technique from bitcoin.
TODO: Think about it

### Design & Layout
- Have a overview over own addresses and the possibilities to add more in frontend and api
- Choose sender address by sending coins

einhornimmond:
With AddMemberTransaction save master public key extended (+ chain code) of user to blockchain.
The only Address of this user which can receive Creation Transactions. 

With AddAddressTransaction save from master public key derived public key together with child index, username or email hash and signed with parent private key. So it is possible to caluclate that this public key belongs to the parent public key.
child index 0 is reserved for GDT. 

For that I need implementation of Public child key derivation working with ed25519.

einhornimmond:
For derivation in the extended privkey the 3rd highest bit must be clear,
In this paper: https://raw.githubusercontent.com/input-output-hk/adrestia/master/user-guide/static/Ed25519_BIP.pdf
the reason can be find. Seems to have todo with security. 

So for ca. 50% of our addresses it isn't valid.
But that is not so bad. 
With setting the bit, the public key change, but not the seed.
If we save with add member simply both public keys and use the second only to verifiy that a child address belongs to user, we don't need to change existing code.

Or I fork the code and simply remove the check. But if it is a security measurement, it isn't such a good idea.

PR #899, opened 2021-09-24
Backend multiple addresses

ulfgebhardt:
## 🍰 Pullrequest

stale branch, whats with this @einhornimmond ?

### Issues
- relates #608

### Todo

- [X] None


einhornimmond:
It has todo with #608, but what exactly I don't know any longer without further Inspectation.

einhornimmond:
You don't like it, so you don't merge it and now I can make it again. But I like to keep it as reference that I am faster if allow to work on it at last.

einhornimmond:
It was a experiment for using ed25519 and BIP32 together from a rust modul. 
First I have testet it with Apollo, then I was able to use it on login-server.
More Details on Multiple Adresses and BIP32: #98

PR #915, opened 2021-09-28
Add User on createUser to community server db

### LINKED ISSUE #913: 🐛 [Bug] User Table in Community-Server not filled with create json via vue
## 🐛 Bugreport
If a new user was created with vue frontend he exist only on Login-Server. 
Normally user would be copied over from Community Server on first login. 
But without visiting old frontend no user will be copied.
That leads to errors, if the apollo backend try to read a user like in the new listTransactions implementation.

einhornimmond:
## 🍰 Pullrequest
- on create user call, adding new user to community server db
- return user object in login server after call create user

### Issues
- fixes #913


backend/src/graphql/resolvers/UserResolver.ts
```diff
@@ -84,6 +87,18 @@ export class UserResolver {
...
+    user.pubkey = Buffer.from(fromHex(qluser.pubkey))
+    user.email = qluser.email
+    user.firstName = qluser.firstName
+    user.lastName = qluser.lastName
+    user.username = qluser.username
+
+    const repository = getRepository(DbUser)
+    repository.save(user).catch(() => {
+      throw new Error('error saving user')
+    })
```
ulfgebhardt, line 101: What about the case that the user already exists? Does this never happen?

Also I suggest the following shorthand which should work:

```suggestion
    user.save().catch(() => {
      throw new Error('error saving user')
    })
```

einhornimmond: Ja diese Version läuft solange wir das Active Record Pattern benutzen
"Using the Active Record approach, you define all your query methods inside the model itself, and you save, remove, and load objects using model methods. "
Aber mir scheint das du und moritz sich darüber einig seit, das ihr lieber das DataMapper Pattern benutzen wollt 
"Using the Data Mapper approach, you define all your query methods in separate classes called "repositories", and you save, remove, and load objects using repositories."
und daher dachte kann ich ja neue Sachen gleich mit dem DataMapper Pattern definieren, damit wir das später nicht wieder umbauen müssen. 

Oder wollen wir das vermischen?

einhornimmond: In this case no.
Because if the user already exist

> s? Does this never happen?



> What about the case that the user already exists? Does this never happen?
In this case not. Because if the user already exist login-server will be return an error and the function will exit before calling this.

.gitignore
```diff
@@ -6,3 +6,4 @@ messages.pot
 nbproject
 .metadata
 /.env
+package-lock.json
```
Mogge, line 9: ???

backend/src/graphql/resolvers/TransactionResolver.ts
```diff
@@ -216,13 +216,9 @@ export class TransactionResolver {
...
     @Ctx() context: any,
   ): Promise<TransactionList> {
-    // get public key for current logged in user
-    const result = await apiGet(CONFIG.LOGIN_API_URL + 'login?session_id=' + context.sessionId)
-    if (!result.success) throw new Error(result.data)
-
     // load user
     const userRepository = getCustomRepository(UserRepository)
-    const userEntity = await userRepository.findByPubkeyHex(result.data.user.public_hex)
+    const userEntity = await userRepository.findByPubkeyHex(context.pubKey)
```
Mogge, line 221: Are you up to date to have the pubKey in the context object?

backend/src/graphql/resolvers/UserResolver.ts
```diff
@@ -86,6 +88,17 @@ export class UserResolver {
     if (!result.success) {
       throw new Error(result.data)
     }
+    const qluser = new User(result.data.user)
```
Mogge, line 91: I suggest `user` here as it is imported.

backend/src/graphql/resolvers/UserResolver.ts
```diff
@@ -86,6 +88,17 @@ export class UserResolver {
...
+    const user = new DbUser()
```
Mogge, line 92: `dbUser`

backend/src/graphql/resolvers/UserResolver.ts
```diff
@@ -2,12 +2,14 @@
 /* eslint-disable @typescript-eslint/explicit-module-boundary-types */
 
 import { Resolver, Query, Args, Arg, Authorized, Ctx, UseMiddleware, Mutation } from 'type-graphql'
+import { from_hex as fromHex } from 'libsodium-wrappers'
```
Mogge, line 5: Why is this needed? Please explain.

backend/src/graphql/resolvers/UserResolver.ts
```diff
@@ -86,6 +88,17 @@ export class UserResolver {
     if (!result.success) {
       throw new Error(result.data)
     }
+    const qluser = new User(result.data.user)
+    const user = new DbUser()
+    user.pubkey = Buffer.from(fromHex(qluser.pubkey))
```
Mogge, line 93: Why do we need fromHex here? The pubkey is a buffer.

Mogge (changes requested):
Why is the yarn.lock deleted in the backend folder? What does this new package?

.gitignore
einhornimmond: > 
> 
> Why is the yarn.lock deleted in the backend folder? What does this new package?

Sorry, I have forget to restart the container to get a rebuilded yarn.lock.

einhornimmond:
> 
> 
> Why is the yarn.lock deleted in the backend folder? What does this new package?

Sorry, I have forget to restart the container to get a rebuilded yarn.lock.

.gitignore
einhornimmond: I use npm because I have problems with yarn and to prevent problems with yarn for the other, I have added my package-lock.json to gitignore

backend/src/graphql/resolvers/TransactionResolver.ts
einhornimmond: Yes, I have merged your changes with the jwt token.

backend/src/graphql/resolvers/UserResolver.ts
einhornimmond: Okay, I exchange user with dbuser and qluser with user.

backend/src/graphql/resolvers/UserResolver.ts
einhornimmond: For translating the pubkey from hex format into binary format.
I choose libsodium because I also need it later for check signatures and translate base64 to binary.
Login-Server use libsodium for every crypto-stuff and also for base64 and hex encoding. In the past I had difficulties with native base64 encoding on php so I used also libsodium there.

backend/src/graphql/resolvers/UserResolver.ts
einhornimmond: The Community Server db save the public key in binary format, the graphQL user in hex format.
Binary needs half the space of the hex version, but the hex version is easier for transport via http and debug.

Issue #924, opened 2021-10-01
🔧 [Refactor] Define the decay start block as fixed value in code


ulfgebhardt:
## 🔧 Refactor ticket

Define the decay start block as fixed value in code. Currently the decay start block is read from database which triggers alot of queries. We should define it as fixed value in the code since we don't need the ability to turn it off again.

PR #927, opened 2021-10-01
Clean models2

### LINKED ISSUE #926: 🔧 [Refactor] Fix models to use Datamapper instead of ActiveRecord approach
## 🔧 Refactor ticket

Fix models to use Datamapper instead of ActiveRecord approach

ulfgebhardt:
## 🍰 Pullrequest

I cleaned all models and used Repositories. Furthermore I removed a query not used and moved the decay start block query into the transaction Repository. I removed the virtual static field from the decay model. This could impact performance, but should be solved in #924

#decay

### Issues

- fixes #926

### Todo

- [X] None


einhornimmond (approved):
Sieht soweit gut aus.

PR #943, opened 2021-10-06
feat: Global Catch of Publisher ID to Store

### LINKED ISSUE #941: 🚀 [Feature] Handle PID in Frontend
## 🚀 Feature

We need to handle a PID parameter in the frontend.

- The PID is the publisher ID, a referral, from EloPage.
- This PID is transmitted via Parameter to our application: `ourapplication.com/path?pid=123`
- The PID is a number(!) of a specific length(?)

We want to:
- [X] Save the PID in our application
- [ ] Remove the parameter from URL (?)
- [X] Allow the parameter on every route and handle it properly according to our rules
- [ ] Transmit the PID with the register process
- [ ] Transmit the PID with the login process
- [X] Allow a revisit of the page without losing a previous PID (?)(cookie)
- [ ] Handle different PIDs and have priorities defined
- [ ] (Backend) Have a default pid if none present (not really this issue)

It is ok to implement only parts of the mechanic (e.g. do not save it across sessions), but tickets for the missing parts need to be made.

Minimum requirement:
- Properly handle the PID on the register process
- Allow multiple entry points to the application with a pid and have it available in register.

Mogge:
## 🍰 Pullrequest
 * checks all routes for the presence of `pid` in the query
 * commits the pid to the store.
 * store does not delete pid on logout

### Issues
- fixes #941

### DIFF
--- a/frontend/src/main.js
+++ b/frontend/src/main.js
@@ -50,6 +50,11 @@ Vue.config.productionTip = false
 loadAllRules(i18n)
 
 router.beforeEach((to, from, next) => {
+  const publisherId = to.query.pid
+  if (publisherId) {
+    store.commit('publisherId', publisherId)
+    to.query.pid = undefined
+  }
   if (to.meta.requiresAuth && !store.state.token) {
     next({ path: '/login' })
   } else {
--- a/frontend/src/store/store.js
+++ b/frontend/src/store/store.js
@@ -29,6 +29,9 @@ export const mutations = {
   newsletterState: (state, newsletterState) => {
     state.newsletterState = newsletterState
   },
+  publisherId: (state, publisherId) => {
+    state.publisherId = publisherId
+  },
 }
 
 export const actions = {
--- a/frontend/src/store/store.test.js
+++ b/frontend/src/store/store.test.js
@@ -9,6 +9,7 @@ const {
   lastName,
   description,
   newsletterState,
+  publisherId,
 } = mutations
 const { login, logout } = actions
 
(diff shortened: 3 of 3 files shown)


frontend/src/routes/guards.js
```diff
@@ -0,0 +1,16 @@
+const addNavigationGuards = (router, store) => {
+  router.beforeEach((to, from, next) => {
+    const publisherId = to.query.pid
+    if (publisherId) {
+      store.commit('publisherId', publisherId)
+      delete to.query.pid
+    }
```
ulfgebhardt, line 7:
```suggestion
    // handle publisherId
    const publisherId = to.query.pid
    if (publisherId) {
      store.commit('publisherId', publisherId)
      delete to.query.pid
    }
```

frontend/src/routes/guards.js
```diff
@@ -0,0 +1,16 @@
...
+    if (to.meta.requiresAuth && !store.state.token) {
+      next({ path: '/login' })
+    } else {
+      next()
+    }
```
ulfgebhardt, line 12:
```suggestion
    // handle authentication
    if (to.meta.requiresAuth && !store.state.token) {
      next({ path: '/login' })
    } else {
      next()
    }
```

frontend/src/routes/guards.js
```diff
@@ -0,0 +1,16 @@
+const addNavigationGuards = (router, store) => {
+  router.beforeEach((to, from, next) => {
+    const publisherId = to.query.pid
+    if (publisherId) {
+      store.commit('publisherId', publisherId)
+      delete to.query.pid
```
ulfgebhardt, line 6: This deletes the query parameter on next navigation?
Also this could be optimized and remove the `publisherId` constant. But I agree on this implementation. Our task is maintainability & readability, not optimization. Thats task of the compiler/transpiler.

ulfgebhardt:
I consider it part of the task to clarify what followup tickets we need to handle things properly. This requires decisions regarding how to handle the pid in our application & how our elopage register process will look like. If this would be a strong blocker due to decision time, i'd recommend to keep the issue open and merge this.

PR #949, opened 2021-10-07
check for elopage account

einhornimmond:
## 🍰 Pullrequest
Add Call to Login-Server to check with the saved Elopage Transactions if an User already has an Elopage Account. 
In Local Tests should always return false unless an entry is manuelly (z.B: with phpmyadmin) created in gradido_login.elopage_buys with the appropriate email.

### DIFF
--- a/login_server/src/cpp/model/table/ElopageBuy.cpp
+++ b/login_server/src/cpp/model/table/ElopageBuy.cpp
@@ -113,7 +113,7 @@ namespace model {
 		int UserHasElopageTask::run()
 		{
 			auto elopage_buy = Poco::AutoPtr<model::table::ElopageBuy>(new model::table::ElopageBuy);
-			bool hasElopage = elopage_buy->isExistInDB("email", mEmail);
+			bool hasElopage = elopage_buy->isExistInDB("payer_email", mEmail);
 			return 0;
 		}
 	}


backend/src/graphql/resolver/UserResolver.ts
```diff
@@ -31,9 +31,16 @@ import { UserRepository } from '../../typeorm/repository/User'
...
+  async login(
+    @Args() { email, password, hasElopage }: UnsecureLoginArgs,
+    @Ctx() context: any,
+  ): Promise<User> {
     email = email.trim().toLowerCase()
-    const result = await apiPost(CONFIG.LOGIN_API_URL + 'unsecureLogin', { email, password })
+    const result = await apiPost(CONFIG.LOGIN_API_URL + 'unsecureLogin', {
+      email,
+      password,
+      hasElopage,
```
Elweyn, line 42: I don't understand why we add this parameter to the unsecureLogin.
I think what @ulfgebhardt meant was to add hasElopage to the response of the unsecureLogin call, or have I understood it wrongly?

backend/src/graphql/arg/UnsecureLoginArgs.ts
```diff
@@ -7,4 +7,7 @@ export default class UnsecureLoginArgs {
 
   @Field(() => String)
   password: string
+
+  @Field(() => Boolean)
+  hasElopage?: boolean
```
Elweyn, line 12: Same comment as in UserResolver, I think that this should be a new response parameter.

backend/src/graphql/resolver/UserResolver.ts
ulfgebhardt: No you are correct. What I was talking about was a return value not a argument to the login.

`Whenever an user logs in, he can retrieve his eloPage status as well`

The same applies to register I guess - `Whenever a new user registers, he can retrieve his eloPage status upon success` (this scenario is under the assumption that after a register, you are immediately logged in - i am not sure that is the case, tho)

einhornimmond: It seems Apollo is bad calling services with grapqhl similar functionality? :)

Elweyn (approved):

Issue #947, opened 2021-10-07
🐛 [Bug] Newsletter tests seem to be cheaty, not working


ulfgebhardt:
## 🐛 Bugreport

Newsletter tests seem to be cheaty, not working.
We discovered a lot of errors and inaccuracies when applying the same mechanic for Coin-Animation tests in #936 .
Please investigate whats happening there and if its all cool. Furthermore compare to the Coin-Animation tests.

PR #952, opened 2021-10-07
refactor: Clean Up Frontend Links, Tests and Logic

Mogge:
## 🍰 Pullrequest
 * fix deprecated warning on button links (@ogerly please use router-link with to attribute and b-button inside in the future)
 * remove logic from template of select community
 * test all logic of select community
 * poper test of checkboxes in newsletter and coinanimation

 - fixes: #947

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -344,7 +344,7 @@ jobs:
           report_name: Coverage Frontend
           type: lcov
           result_path: ./coverage/lcov.info
-          min_coverage: 75
+          min_coverage: 76
           token: ${{ github.token }}
   
   ##############################################################################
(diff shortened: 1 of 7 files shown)


frontend/src/views/Pages/UserProfile/UserCard_CoinAnimation.spec.js
```diff
@@ -37,6 +37,7 @@ describe('UserCard_CoinAnimation', () => {
 
   describe('mount', () => {
     beforeEach(() => {
+      jest.clearAllMocks()
```
Mogge, line 40: As long as you not clear the mocks, the tests for store commit and toaster are always true, because there is an call before with the other values @ulfgebhardt

ulfgebhardt (approved):
Thank you for looking into those tests <3

PR #958, opened 2021-10-10
check pid on gdt server

### LINKED ISSUE #805: 🚀 [Feature] member and PID collection from elopage on GDT-Server
## 🚀 Feature

- cyclic request all registered Elopage members inkluding the PID
- store all necessary data of an Elopage member including the PID with the GDT-account
- new service request for a PID-check
  * input parameter PID
  * search this PID in the collected Elopage members
  * if found return this PID
  * if not found return the PID of "Förderkreis"

einhornimmond:
## 🍰 Pullrequest
- add call for backend for checking if pid exist and if not, return pid from "Förderkreis"
- added call on gdt server already

### Issues
- fixes #805

### DIFF
--- a/backend/src/graphql/resolver/GdtResolver.ts
+++ b/backend/src/graphql/resolver/GdtResolver.ts
@@ -36,7 +36,7 @@ export class GdtResolver {
   @Authorized()
   @Query(() => Number)
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  async checkPid(@Arg('pid') pid: number): Promise<number> {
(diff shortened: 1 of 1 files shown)


einhornimmond:
This function does what clauspeter has suggested: 

- input parameter PID
- search this PID in the collected Elopage members
- if found return this PID
- if not found return the PID of "Förderkreis"

einhornimmond:
getPidForUser would therefore be totally misleading. 
What would be possible, however, would be checkPidForExistence.

PR #968, opened 2021-10-12
refactor: In Production only Show one Community

### LINKED ISSUE #967: 🔧 [Refactor] Production Community List only the Gradido-Akademy
## 🔧 Refactor ticket

On production we cannot mix production with staging and development environments. Therefore the mocked community lists msut be reduced to Gradido-Akademy only on production.

Mogge:
## 🍰 Pullrequest
 * introduce PRODUCTION in environment
 * check if production then only return Gradido Production

### Issues
- fixes #967


backend/.env.dist
```diff
@@ -1,3 +1,4 @@
+PRODUCTION=
```
ulfgebhardt, line 1: can we use `NODE_ENV` for this or does this make no sense?

ulfgebhardt (approved):

PR #1026, opened 2021-10-20
refactor: Transaction Component

Mogge:
## 🍰 Pullrequest
 * use gdt transaction id as identifier for collapse
 * use computed instead of multiple method calls
 * tests

### DIFF
--- a/frontend/src/components/Transaction.spec.js
+++ b/frontend/src/components/Transaction.spec.js
@@ -2,15 +2,9 @@ import { mount } from '@vue/test-utils'
 import Transaction from './Transaction'
 import Vue from 'vue'
 
-// disable throwing Errors on warnings to catch the warning
-Vue.config.warnHandler = (w) => {}
-
 const localVue = global.localVue
 
 const consoleErrorMock = jest.fn()
-
-// eslint-disable-next-line no-console
-console.error = consoleErrorMock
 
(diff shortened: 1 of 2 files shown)


frontend/src/components/Transaction.spec.js
```diff
@@ -1,8 +1,17 @@
...
+
+// disable throwing Errors on warnings to catch the warning
+Vue.config.warnHandler = (w) => {}
 
 const localVue = global.localVue
 
+const consoleErrorMock = jest.fn()
+
+// eslint-disable-next-line no-console
+console.error = consoleErrorMock
```
ulfgebhardt, line 13: Whats this? Why?

frontend/src/components/TransactionCollapse.spec.js
```diff
@@ -2,8 +2,18 @@ import { mount } from '@vue/test-utils'
...
+
+// disable throwing Errors on warnings to catch the warning
+Vue.config.warnHandler = (w) => {}
+
 const localVue = global.localVue
 
+const consoleErrorMock = jest.fn()
+
+// eslint-disable-next-line no-console
+console.error = consoleErrorMock
```
ulfgebhardt, line 15: 👀 mmm. Is this because of the `throw new Error...` or did you forget this when debugging?

PR #1045, opened 2021-10-26
feat: Setup Admin Interface

### LINKED ISSUE #1034: 🚀 [Feature] Setup Admin Interface
## 🚀 Feature

Setup Admin Interface

We need an admin interface where we can manage users, create coins and everything else that is already present in the old admin interface. So we need a Set-Up for this so that we can work on the vue after it.

- [ ] vue, no ts
- [ ] Graphql backend connection
- [ ] Token handling(login)
- [ ] Handle No Token/Unauthorized
- [ ] Tests
- [ ] Lint
- [ ] Workflows
- [ ] Bootstrap
- [ ] nginx route integration
- [ ] docker
- [ ] docker-compose

Mogge:
## 🍰 Pullrequest
Setting up the admin interface

### Issues
- fixes #1034 

### Todo
- [X] Setup Docker and Docker Compose
- [X] NGINX (which port?, 8080?)
- [X] setup tests as in frontend
- [X] creare connection to graphql API
- [X] setup store
- [X] setup router
- [ ] Handle JWT (to be done when the backend can handle roles)
- [X] Setup i18n an locales
- [X] CI Workflow


admin/src/router/routes.js
```diff
@@ -0,0 +1,13 @@
+import NotFound from '@/components/NotFoundPage.vue'
+
+const routes = [
+  {
+    path: '/',
+    meta: {
+      requiresAuth: true,
+    },
```
ulfgebhardt, line 8: is this possible?

docker-compose.yml
```diff
@@ -30,6 +30,30 @@ services:
...
+  ########################################################
+  # ADMIN INTERFACE ######################################
+  ########################################################
+  admin:
+    image: gradido/admin:latest
+    build:
+      context: ./admin
+      target: production
+    networks:
+      - internal-net
```
ulfgebhardt, line 42: this requires `external-net`also in production, since its a frontend, correct?

```suggestion
      - internal-net
      - external-net
```

docker-compose.override.yml
```diff
@@ -20,6 +20,25 @@ services:
...
 
+  ########################################################
+  # ADMIN INTERFACE ######################################
+  ########################################################
+  admin:
+    image: gradido/admin:development
+    build:
+      target: development
+    networks:
+      - external-net
```
ulfgebhardt, line 31: In relation to the comment on the `docker-compose.yml`. Please read there.

```suggestion
```

admin/src/store/store.js
```diff
@@ -0,0 +1,19 @@
...
+    state.token = token
+  },
+}
+
+const store = new Vuex.Store({
+  mutations,
+  state: {
+    token: null,
+  },
+})
```
ulfgebhardt, line 17: If this store reads the same as our frontend, since they share a tld, we need to set a localStorage name for the "cookie". We should expect that both systems can not directly interact with eachother via the users browser.

Please also make a followup ticket for the protection of our localstorage data.

Asume the following:
1. User has his sensitive data in the local storage on `our.tld/wallet`
2 Now a rogue element has access to our server and can place a page at `our.tld/hack`

- Can we make sure that hack cannot access the local storage of wallet?
- What about subdomains?

In the scope of this PR its just required to have two distinctive "cookie" names for admin and wallet.

admin/src/main.js
```diff
@@ -0,0 +1,63 @@
...
+import VueApollo from 'vue-apollo'
+
+import CONFIG from './config'
+
+import { BootstrapVue } from 'bootstrap-vue'
+
+const httpLink = new HttpLink({ uri: CONFIG.GRAPHQL_URI })
+
+const authLink = new ApolloLink((operation, forward) => {
+  const token = '' // store.state.token
```
ulfgebhardt, line 24:
```suggestion
  const token = store.state.token
```

Please only disable this in debug mode. E.g. have a `NO_AUTH_FOR_DEVELOPMENT` .env setting. You can default it to `true` in the env (not in the `config/index.js`) at this given moment.

admin/src/router/guards.js
```diff
@@ -0,0 +1,12 @@
+const addNavigationGuards = (router, store) => {
+  router.beforeEach((to, from, next) => {
+    // handle authentication
```
ulfgebhardt, line 3: i suggest:
```suggestion
    // handle authentication
    if( to === '/authenticate' && req.post.token){
       store.commit('token',req.post.token)
    }
```
On the `/authenticate` I suggest to check for a valid token in the store and redirect again according to this finding.

admin/src/main.js
```diff
@@ -0,0 +1,63 @@
...
+    if (response.errors && response.errors[0].message === '403.13 - Client certificate revoked') {
+      response.errors[0].message = i18n.t('error.session-expired')
+      store.dispatch('logout', null)
+      if (router.currentRoute.path !== '/login') router.push('/login')
+      return response
+    }
+    const newToken = operation.getContext().response.headers.get('token')
+    if (newToken) store.commit('token', newToken)
+    return response
+  }) */
```
ulfgebhardt, line 41:
```suggestion
   .map((response) => {
    if (response.errors && response.errors[0].message === '403.13 - Client certificate revoked') {
      response.errors[0].message = i18n.t('error.session-expired')
      store.dispatch('logout', null)
      if (router.currentRoute.path !== '/login') router.push('/login')
      return response
    }
    const newToken = operation.getContext().response.headers.get('token')
    if (newToken) store.commit('token', newToken)
    return response
  })
```

I think we need this - consider to disable this with `NO_AUTH_FOR_DEVELOPMENT` for now if its a problem

ulfgebhardt, line 41:
```suggestion
   .map((response) => {
    if (response.errors && response.errors[0].message === '403.13 - Client certificate revoked') {
      response.errors[0].message = i18n.t('error.session-expired')
      store.dispatch('logout', null)
      if (router.currentRoute.path !== '/login') router.push('/login')
      return response
    }
    const newToken = operation.getContext().response.headers.get('token')
    if (newToken) store.commit('token', newToken)
    return response
  })
```

I think we need this - consider to disable this with `NO_AUTH_FOR_DEVELOPMENT` for now if its a problem

admin/src/App.vue
```diff
@@ -0,0 +1,19 @@
...
+  methods: {
+    async readToken() {
+      if (window.localStorage && window.localStorage.getItem('vuex')) {
+        this.$store.commit('token', JSON.parse(window.localStorage.vuex).token)
+      }
+    },
+  },
+  created() {
+    this.readToken()
+  },
```
ulfgebhardt, line 17: see `router/guards.js` for a suggestion. We also must assume we cannot access the coresponding localStorage entry (see other comment of mine) and we actually should ensure nobody but our domain can.

```suggestion
```

nginx/nginx.conf
```diff
@@ -69,7 +69,19 @@ server {
...
-	location /sockjs-node {
+	location /admin {
+		proxy_http_version 1.1;
+		proxy_set_header   Upgrade $http_upgrade;
+		proxy_set_header   Connection 'upgrade';
+		proxy_set_header   X-Forwarded-For $remote_addr;
+		proxy_set_header   X-Real-IP  $remote_addr;
+		proxy_set_header   Host $host;
+		
+		proxy_pass         http://admin:8080;
```
ulfgebhardt, line 80:
```suggestion
		proxy_pass         http://admin:5000;
```

This is just a suggestion, but shouldn't we port all our applications like 3000, 4000, 5000 ?

ulfgebhardt (changes requested):

ulfgebhardt (approved):

PR #1075, opened 2021-11-03
Login hotfixes 1.5.1

einhornimmond:
## 🍰 Pullrequest
- fix error in user-search (came with publisher_id adding)
- fix broken cancel button in checkTransactions 
- move cancel button after skip button in checkTransactions


Elweyn (approved):
Seems to only change the logic when a transaction to sign can be deleted or canceled.

PR #1079, opened 2021-11-05
fix: Await Resolved Promises in Backend Unit Tests

Mogge:
## 🍰 Pullrequest
dito. Make sure to `await` the resolved promise.


ulfgebhardt (approved):
we need a rebase - coverage is incorrect and needs to be rechecked - good work 🚀

PR #1044, opened 2021-10-25
Apollo create transactions without signation

Elweyn:
## 🍰 Pullrequest
Create transactions on apollo need the premise of protobuf (blockchain), in the test on the local maschine the protobuf was not recognized so this branch is the same code without "Blockchain" 

### Issues
- #823

PR #1081, opened 2021-11-05
fix sign error

einhornimmond:
in #1044 a error has shown with reversed sign, this fix it
With a call to http://localhost/state-balances/autoHealAllStateUserTransactions
all values in db will be fixed.


ulfgebhardt:
can we close this @einhornimmond ?

Issue #1088, opened 2021-11-07
🐛 [Bug] Make sure user confirms EMail before login


ulfgebhardt:
## 🐛 Bugreport

Make sure user confirms EMail before login. Currently the user can login without confirming his EMail and he is even advised to do so after register.

1. Register
2. "Hey we sent you an EMail - make sure to confirm it"-Page
3. Login
4. Redirect to "Hey you did not confirm your EMail yet - want a new Email?"-Page, not logged in
5. Click on Link in Email -> "Hey nice, lets go and login"
6. Login
7. Cool

Elweyn:
## TODOs
- [ ] Login when not Activated > Redirect to "Hey you did not confirm your EMail yet - want a new Email?"-Page, not logged in
- [ ] Click on Link in Email -> "Hey nice, lets go and login"

PR #1099, opened 2021-11-12
Adding a check that the user is activated before letting them login.

### LINKED ISSUE #1097: 🐛 [Bug] EloPage Registration no publisher for user
## 🐛 Bugreport
User has registrated without pid and is now registrated in EloPage without publisher. 
The expected result should be that he has a default publisher.

## TODOs:
- [x] Check if login_users publisherId = null => gives back url with default pid.
- [x] If it is the case we need to check if EloPage store the sended pid in a Cookie
- [x] If User withdraw the pid argument from the EloPage link => Stores the User without publisher

Elweyn:
## 🍰 Pullrequest
bugfix on the master branch tag 1.5.1 we have a problem with the registration that makes that the user is not registrated on KlickTipp. Fix: Change the process so that the user needs to validate his email before he can login.

### Issues
- Fixes: #1088 

### TODOs

- [X] Check if user is activated
- [x] Change the register.vue so that we say to the user that he needs to validate his email.
- [x] Change the error message for not activated user?
- [ ] Login when not Activated > Redirect to "Hey you did not confirm your EMail yet - want a new Email?"-Page, not logged in
- [ ] Click on Link in Email -> "Hey nice, lets go and login"


Elweyn:
@einhornimmond, @ulfgebhardt, @ogerly,
could you guys please review this it would be great!!!! 😄

Elweyn:
## TODOs
- [ ] Login when not Activated > Redirect to "Hey you did not confirm your EMail yet - want a new Email?"-Page, not logged in
- [ ] Click on Link in Email -> "Hey nice, lets go and login"

frontend/src/views/Pages/thx.vue
```diff
@@ -17,25 +19,34 @@
...
   checkEmail: {
+    headline: 'site.thx.title',
     subtitle: 'site.thx.checkEmail',
     button: 'login',
     linkTo: '/login',
   },
+  login: {
+    headline: 'site.thx.errorTitle',
+    subtitle: 'site.thx.activateEmail',
+    button: 'Send Activation Link',
```
ulfgebhardt, line 48: Why not use a string identifier here?

login_server/src/cpp/JSONInterface/JsonUnsecureLogin.cpp
```diff
@@ -158,6 +157,10 @@ Poco::JSON::Object* JsonUnsecureLogin::handle(Poco::Dynamic::Var params)
 		AWAIT(hasElopageTask)
 		result->set("hasElopage", hasElopageTask->hasElopage());
 		return result;
+	case USER_EMAIL_NOT_ACTIVATED:
```
ulfgebhardt, line 160: The error case must be before the success case to prevent potential problems.
Please move this block above the `... case: USER_COMPLETE` block

ulfgebhardt (commented):
Looks very clean! Two minor things tho.
Thanks for doing this so quickly!

login_server/src/cpp/JSONInterface/JsonUnsecureLogin.cpp
einhornimmond: Technically, I don't see a problem, because only one case can exist at a time. It should therefore not matter in terms of function in which order the cases are checked.
Maybe it is nicer for the eye.
As long as all cases have an return or break at the end. 
USER_NO_GROUP is a special case. Because it is intentional that it hasn't a break after it the success block with USER_NO_PRIVATE_KEY and USER_COMPLETE must follow.

einhornimmond (approved):
Looks like a very smart and simple solution for me.

einhornimmond:
What just occurred to me, because you changed the login-server code I must merge my other hotfixes into this branch. 
Luckly I have build the hotfixes on the 1.5 version and not on the master.

frontend/src/views/Pages/thx.vue
Elweyn: Oh good catch!

login_server/src/cpp/JSONInterface/JsonUnsecureLogin.cpp
Elweyn: For lisibility I agree on the error block before success block!

frontend/src/locales/de.json
```diff
@@ -152,6 +153,10 @@
       "ooops": "Ooops!",
       "text": "Seite nicht gefunden. Aber keine Sorge, wir haben noch viele andere Seiten zum Erkunden"
     },
+    "activateEmail": {
+      "headline": "",
+      "subtitle": ""
+    },
```
Elweyn, line 159:
```suggestion
```

frontend/src/locales/en.json
```diff
@@ -152,6 +153,10 @@
       "ooops": "Ooops!",
       "text": "Page not found. Do not worry though, we have plenty of other pages to explore"
     },
+    "activateEmail": {
+      "headline": "",
+      "subtitle": ""
+    },
```
Elweyn, line 159:
```suggestion
```

frontend/src/views/Pages/thx.vue
Elweyn:
```suggestion
```

ulfgebhardt (approved):

PR #1101, opened 2021-11-16
allows db names like gradido_stage2-mysql

einhornimmond:
## 🍰 Pullrequest
My db name :  gradido_stage2-mysql produce an error so I changed the syntax to get it working on my test setup.

### DIFF
--- a/database/src/prepare.ts
+++ b/database/src/prepare.ts
@@ -20,10 +20,10 @@ export default async (): Promise<void> => {
   await con.connect()
 
   // Create Database `gradido_community`
-  await con.query(`
-    CREATE DATABASE IF NOT EXISTS ${CONFIG.DB_DATABASE} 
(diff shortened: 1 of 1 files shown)


database/src/prepare.ts
```diff
@@ -20,10 +20,10 @@ export default async (): Promise<void> => {
   await con.connect()
 
   // Create Database `gradido_community`
-  await con.query(`
-    CREATE DATABASE IF NOT EXISTS ${CONFIG.DB_DATABASE} 
-      DEFAULT CHARACTER SET utf8mb4
-      DEFAULT COLLATE utf8mb4_unicode_ci;`)
+  await con.query('\
+    CREATE DATABASE IF NOT EXISTS `${CONFIG.DB_DATABASE}`\
```
ulfgebhardt, line 24: Wasn't this a problem with the previous skeema setup aswell? I copied the query from there?!

PR #1112, opened 2021-11-17
Login hook elopage

### LINKED ISSUE #1093: 🚀 [Feature] Implement the EloPage Webhooks
## 🚀 Feature

We need to fill the Elopage table `login_elopage_buys` which is filled via a webhook from elopage

### LINKED ISSUE #1108: 🔧 [Refactor] Implement Login Webhook for Elopage on Apollo
## 🔧 Refactor ticket

The `login_server` accepts a webhook from elopage which fills the table `login_elopage_buys`. This have to be implemented
on Apollo. Furthermore this webhook should support a secret in oder to prevent fraudulent actors to cheat on donation values or disrupt our service.

ulfgebhardt:
### Issues

- fixes #1093 
- fixes #1108 

### Todo

- [ ] test
- [x] clarify how the user sets his password when creating an user via elopage hook @einhornimmond 
- [x] clarify how the database field `elopage_user_id` is filled - how is the production data on this? @einhornimmond 
- [x] clarify if there is webhooks on the community server doing similar things? @einhornimmond 
- [x] one field is multiplied with `100.0f` - I assume its the amount donated. Why is it multiplied? Or is it a different field? @einhornimmond


backend/src/webhook/elopage.ts
```diff
@@ -0,0 +1,370 @@
...
+        firstName = val
+        break
+      case 'payer[last_name]':
+        lastName = val
+        break
+      default:
+        // eslint-disable-next-line no-console
+        console.log(`Unknown Elopage Value '${entry}'`)
+    }
+    return null // we write things into the loginElopgaeBuy object, no return value needed
```
Mogge, line 93: Please remove this line. It is confusing. We do not need a return value for the map function, as you say. As it is, it looks like the return statements in the following if clauses, which is not the same.

backend/src/webhook/elopage.ts
```diff
@@ -0,0 +1,370 @@
...
+    if ((await LoginUser.count({ email })) !== 0) {
+      // eslint-disable-next-line no-console
+      console.log(`Did not create User - already exists with email: ${email}`)
+      return
+    }
+
+    // generate a random password - 8 random bytes, the email, special char, capital & small letter, number and another set of 8 random bytes
+    // TODO: The user will be forced to reset his password - how was this done before?
+    const password =
+      randomBytes(8).toString('hex') + email + '!aA1' + randomBytes(8).toString('hex')
```
Mogge, line 145: Good solution to ensure a valid password!

backend/src/webhook/elopage.ts
```diff
@@ -0,0 +1,370 @@
...
+
+    const userResolver = new UserResolver()
+    try {
+      await userResolver.createUser({
+        email,
+        firstName,
+        lastName,
+        password,
+        language: 'default',
```
Mogge, line 154: Frontend should fallback to default language with this. I guess it works.

backend/src/webhook/elopage.ts
```diff
@@ -0,0 +1,370 @@
...
+        publisherId: loginElopgaeBuy.publisherId,
+      })
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.log(`Could not create User for ${email}. Following Error occured:`, error)
+    }
+  }
+}
+
+/*
```
Mogge, line 164: Do you want to keep this dump?

Mogge (changes requested):

einhornimmond:
`clarify how the user sets his password when creating an user via elopage hook `
Via Email Verification Code. 
User get's email, click on link and checkEmail will automatic forward user to update password page if no password was set. 
After the User has choosen a password, his key pair will be generated and saved encrypt with his password and email and the passphrase should be saved encrypted with a server admin public key, but at the moment it is saved unencrypted. 
In the Past that was also the place to show the User his Passphrase so he can store it away save. But it leads to much confusion. So we have disabled this feature. 

`clarify how the database field elopage_user_id is filled - how is the production data on this?`
It isn't filled at all. 
On production there all filled with 0.
By designing the table I thought the publisher_id is the elopage_user_id so I would fill them later. But I never finished it.

`clarify if there is webhooks on the community server doing similar things?`
No, the community server don't listen at the elopage webhooks. 

`one field is multiplied with 100.0f - I assume its the amount donated. Why is it multiplied? Or is it a different field?`
I save the product_price as int so I need to multiple the decimal value I get from elopage to get euro cent. 
It is for preventing accuracy loss which sometimes happen with specific numbers saved in decimal format.

backend/src/webhook/elopage.ts
ulfgebhardt: A map **must** return a value

ulfgebhardt: use apply?

backend/src/webhook/elopage.ts
ulfgebhardt: This was part of `createUser`:

```
    // Validate Language (no throw)
    if (!isLanguage(language)) {
      language = DEFAULT_LANGUAGE
    }
```

by setting language to something thats not a language (currently only `en` and `de`) the `DEFAULT_LANGUAGE` would be applied. This is currently `de`. This mechanic was chosen in order to not assume `de` will be the default language.

Tho I take this into account and make `language` for `createUser` an optional parameter:
```
    // Validate Language (no throw)
    if (!language || !isLanguage(language)) {
      language = DEFAULT_LANGUAGE
    }
```

Especially since we figured out that there is more changes required for the `createUser` function, I consider it fine to do thing there.

backend/src/webhook/elopage.ts
ulfgebhardt: Not good enough tho. We need to have a special case for the users coming from elopage where they have to set their password after email activation

backend/src/webhook/elopage.ts
ulfgebhardt: no

backend/src/webhook/elopage.ts
ulfgebhardt: We should think about changing the whole process. See #1123

backend/src/webhook/elopage.ts
```diff
@@ -0,0 +1,154 @@
...
+
```
einhornimmond, line 21: Have you asked Theodora if you can make here address public?

einhornimmond (approved):
The elopage part look ok, if you haven't a problem with the many log entries which will produced over time with that.

Mogge (approved):

backend/src/webhook/elopage.ts
ulfgebhardt: Email and Address are both available on the internet freely when searching for that name - therefore I believe its ok that we keep this data in for now - a history rewrite is not required

PR #1119, opened 2021-11-18
Error on upgrade database after downgrade

Elweyn:
## 🍰 Pullrequest
Create tables on an already exsting table breaks the script, so fix with a create table if not exists. Insert statement breaks if their is a duplicate key that is imported change so that it does not break.

### Issues
- None

### Todo
- [X] None


Mogge (approved):
Good catch

Mogge:
What happens when the gradido login database is not present?

Elweyn:
Good question! I did not check that!

ulfgebhardt (commented):
I explicitly decided against this when I wrote this, since I expect the tables to not exist and the data not to be present.
In our case you actually had a broken database state. You did a `migrate up` for 2 patches and later did not revert them correctly since you switched branches.

We can decide to do it as you suggest, but this is actually an expected behaviour and ment to fail since you have an unclean database.

Opinions @Elweyn @Mogge ?

Elweyn:
OK I understand your point, so what would have been the right path of action so that I don't get a broken db?

I did:
Work on something near master branch, then I had to go on the tag 1.5.1 where the db lays 2 patches behind.
Now I did go in the migrations table and deleted the 2 patches /!\ This should have broke the db /!\
After I did go back on master and started the db again and it could not upgrade the db, what is logical since the tables were present.
What should I have done instead? Is their an action to downgrade the db? @ulfgebhardt

PR #1120, opened 2021-11-18
Error for removing coin animation

Elweyn:
## 🍰 Pullrequest
When clicking on the coin animation in the user profile we want to desactivate the coinanimation and by clicking it again it should be activated again, and it should be stored in the database.
Set Coinanimation on true => it worked
Set Coinanimation on false => no update was started on the settings

### Issues
- None

### Todo
- [X] None


backend/src/graphql/resolver/UserResolver.ts
```diff
@@ -537,7 +537,7 @@ export class UserResolver {
     await queryRunner.startTransaction('READ UNCOMMITTED')
 
     try {
-      if (coinanimation) {
+      if (coinanimation !== undefined) {
```
Mogge, line 540: Why?

Mogge:
```
null !== undefined

-> true
```

Mogge:
```
if (undefined)
 -> false
```

Mogge:
```
if (null)
 -> false
```

Elweyn: Yeah cause of: 
```
if (false)
  -> store false
```

Elweyn: Now it does not store false, so each time you log back in you get the coin animation, the db update is not triggered on coinanimation === false!

backend/src/typeorm/repository/UserSettingRepository.ts
```diff
@@ -18,7 +18,7 @@ export class UserSettingRepository extends Repository<UserSetting> {
     let entity = await this.findOne({ userId: userId, key: key })
 
     if (!entity) {
-      entity = new UserSetting()
+      entity = this.create()
```
Elweyn, line 21:
```suggestion
      entity = new UserSetting()
```

backend/src/graphql/resolver/UserResolver.ts
Elweyn:
```suggestion
      if (coinanimation !== null && coinanimation !== undefined) {
```

ulfgebhardt (approved):

PR #1125, opened 2021-11-19
Login admin interface

ulfgebhardt:
## 🍰 Pullrequest

Implements the login/authenticate mechanism for the new admin interface. Furthermore the wayback is implemented and a convenience mode to disabled auth for development

### Issues

- includes #1116 
- includes #1045 

### Todo

- [x] Convenience mode
- [x] `isAdmin` from backend
- [x] logout
- [x] back2wallet
- [x] verify login (requires #1095 )
- [x] frontend tests

PR #1121, opened 2021-11-18
feat: Setup Data Seeding

Mogge:
## 🍰 Pullrequest
Setup seeding of data.

### Issues
- relates #774

PR #1126, opened 2021-11-20
backend_rights

ulfgebhardt:
## 🍰 Pullrequest

Implements a roles & rights based system for authentication & authorization.

### Issues

- based on #1125 

### Todo

- [x] Admin role check (requires #1121 )
- [ ] Clarify if we want to encode role in jwt to save a constant database query @Mogge 
- [x] test
- [x] set `isAdmin` based on role


backend/src/graphql/resolver/UserResolver.ts
```diff
@@ -439,6 +475,7 @@ export class UserResolver {
     return new SendPasswordResetEmailResponse(response.data)
   }
 
+  @Authorized([RIGHTS.RESET_PASSWORD])
```
Mogge, line 478: No token?

backend/src/graphql/resolver/UserResolver.ts
```diff
@@ -565,6 +602,7 @@ export class UserResolver {
     return true
   }
 
+  @Authorized([RIGHTS.CHECK_USERNAME])
```
Mogge, line 605: No token?

backend/src/graphql/resolver/UserResolver.ts
```diff
@@ -588,6 +626,7 @@ export class UserResolver {
     return true
   }
 
+  @Authorized([RIGHTS.CHECK_EMAIL])
```
Mogge, line 629: No token

backend/src/graphql/resolver/UserResolver.ts
```diff
@@ -194,6 +195,37 @@ const SecretKeyCryptographyDecrypt = (encryptedMessage: Buffer, encryptionKey: B
...
+      .catch((error) => {
+        throw new Error(error)
+      })
+    user.coinanimation = coinanimation
+    user.isAdmin = true // TODO implement
+    return user
+  }
+  */
+
+  @Authorized([RIGHTS.LOGIN])
```
Mogge, line 228: No token

Mogge (changes requested):
This is not going to work like this. we have lots of calls that never will have a token, but by setting `@Authorized(SOME_ROLE)` you must have a token, as it is designed by now.

backend/src/auth/JWT.ts
```diff
@@ -0,0 +1,19 @@
+import jwt from 'jsonwebtoken'
+import CONFIG from '../config/'
+import { CustomJwtPayload } from './CustomJwtPayload'
+
+export const decode = (token: string): CustomJwtPayload | null => {
```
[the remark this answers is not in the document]
Mogge: Stupid question. From the context, of course :)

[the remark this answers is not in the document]
ulfgebhardt: The JWT is called token in this context I believe.

ulfgebhardt:
@Mogge Please have a look at this - especially the `isAuthorized` - we make a database query on every call to the server to figure out the role - alternatively we can save the role identifier or whatever in JWT and prevent an additional query.

Mogge (approved):

PR #1134, opened 2021-11-24
proto_signatures

ulfgebhardt:
## 🍰 Pullrequest

Implements the protobuff signatures:
- transactions.tx_hash
- transaction_signatures
- login_pending_tasks

This is done to replicate the old database in its entity with all data present before. We decided to do this as experiment since the `login_pending_tasks` does not hold the data in unencrypted form - just as protobuff data.

### Issues

- None

#### login_admin_pending_tasks
peter@lustig.de schöpft 1000 für Dezember für bob@baumeister.de no memo

before:
```
insert into login_pending_tasks(id,user_id,request,created,finished,result_json,param_json,task_type_id,child_pending_task_id,parent_pending_task_id) values('2','4','
f
d
 r���2X����s�1�Q�w�\�8,c��xN�p@�F��+s�]���µA�
C�����Ȉn�V�͜�^-o�"�k'ʇ�|�sg��Q�P`;��ύ:1
'
 �	١F7O�u�g}���J���ީm�@S�m��~���	����','2021-12-10 23:27:14','2000-01-01 00:00:00','','{"blockchain_type":1}','10','0','0');
```

after:
```
insert into 
login_pending_tasks(id,user_id,request,created,finished,result_json,param_json,task_type_id,child_pending_task_id,parent_pending_task_id) values('4','4','
f
d
 r���2X����s�1�Q�w�\�8,c��xN�p@�!g���F]���ֵoUY���
|_��f�]n��/O�9����՚�"�4�+X�K�
�*Wu��=
��ύ:1
'
 �	١F7O�u�g}���J���ީm�@S�m��~���	��ύ','2021-12-10 23:27:14','2000-01-01 00:00:00','','{"blockchain_type":1}','10','0','0');
```

The difference can be explained by the timestamp seconds when the block was created which is encoded into to protobuff. Remaining date values are the same. The signature resulting from the block will obviously differ and is also included.

I consider this close enough.

### Todo

- [x] login_pending_tasks
- [x] refactor
- [x] decay start block
- [ ] verify that `transactions.txHash` has the correct values
- [ ] verify that `transaction_signatures` has the same entries
- [x] verify that `login_admin_pending_tasks` has the same entries


backend/src/graphql/resolver/TransactionResolver.ts
```diff
@@ -464,6 +470,80 @@ async function getPublicKey(email: string): Promise<string | null> {
   return loginUser.pubKey.toString('hex')
 }
 
+const protoTransaction = (
```
Mogge, line 473: Why do we have this function twice? What is the difference to the one in the AdminResolver?

backend/src/proto/gradido.proto.d.ts
```diff
@@ -0,0 +1,1943 @@
...
+            public static toObject(message: proto.gradido.TransactionBody, options?: $protobuf.IConversionOptions): { [k: string]: any };
+
+            /**
+             * Converts this TransactionBody to JSON.
+             * @returns JSON object
+             */
+            public toJSON(): { [k: string]: any };
+        }
+    }
+}
```
Mogge, line 1943: What the hell is this file doing? Nearly 2000 lines of code what for?

backend/src/proto/gradido.proto.js
```diff
@@ -0,0 +1,4805 @@
...
+            return TransactionBody;
+        })();
+
+        return gradido;
+    })();
+
+    return proto;
+})();
+
+module.exports = $root;
```
Mogge, line 4805: WTF

Mogge (commented):
This is a hell of a PR

ulfgebhardt:
> This is a hell of a PR

In a good or bad way? xD

backend/src/proto/gradido.proto.js
ulfgebhardt: This whole file is generated by the proto library. I checked in the generated file since I do not want to require everyone to build this stuff everytime - just have it statically linked till we remove it again.

backend/src/proto/gradido.proto.d.ts
ulfgebhardt: This whole file is generated by the proto library. I checked in the generated file since I do not want to require everyone to build this stuff everytime - just have it statically linked till we remove it again.

backend/src/graphql/resolver/TransactionResolver.ts
ulfgebhardt: You are perfectly right - this still has to be unified. The main difference is what is signed into the tx hash as I understand it. In every manual transaction we have `sodium.crypto_generichash_update(state, 'start decay')` and in those generated by the system for the transactions and creations we don't update the generic crypto hash with a string value like that. But I am not 100% sure yet. I implemented this in two different versions for now since it is implemented twice (or more times) in the login_server and some more implementations on the community_server. I expect those processes to varry alot and to have to be adapted based on the findings when testing for 100% value reproduction. Later on I will ofc unify those.

backend/src/graphql/resolver/AdminResolver.ts
```diff
@@ -41,13 +232,39 @@ export class AdminResolver {
...
+        loginUser.pubKey,
+        loginAdminUser.pubKey,
+        loginAdminUser.privKey,
+      )
+      loginPendingTask.created = creationDateObj
+      loginPendingTask.resultJson = ''
+      loginPendingTask.paramJson = '{"blockchain_type":1}'
+      loginPendingTask.taskTypeId = 10
+
+      await loginPendingTaskRepository.save(loginPendingTask)
```
Elweyn, line 265: This works and creates the entry in the login_pending_task tabelle!
But we can not confirm this creation on the old backend... http://localhost/account/checkTransactions

Maybe we are still missing something

Elweyn (approved):
It seems to work even if the old admin interface can't query the creations that were created in the new admin.
Could create on new & old backend in the same session

ulfgebhardt:
Closing this - not needed

PR #1141, opened 2021-11-26
Admin delete pending creation query

Elweyn:
### Issues

- None

### Todo

- [X] None


Mogge (approved):
After delete a pending creation, the pending creation still exists in the DB

PR #1151, opened 2021-12-02
fix: Catch Expired Session when Coming from Admin Interface

Mogge:
## 🍰 Pullrequest
Catch server error when session is expired after coming back to the wallet from the admin interface.

### DIFF
--- a/frontend/src/routes/guards.js
+++ b/frontend/src/routes/guards.js
@@ -15,12 +15,19 @@ const addNavigationGuards = (router, store, apollo) => {
   router.beforeEach(async (to, from, next) => {
     if (to.path === '/authenticate' && to.query.token) {
       store.commit('token', to.query.token)
-      const result = await apollo.query({
-        query: verifyLogin,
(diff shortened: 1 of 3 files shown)


frontend/src/routes/guards.js
```diff
@@ -15,12 +15,19 @@ const addNavigationGuards = (router, store, apollo) => {
...
+        .query({
+          query: verifyLogin,
+          fetchPolicy: 'network-only',
+        })
+        .then((result) => {
+          store.dispatch('login', result.data.verifyLogin)
+          next({ path: '/overview' })
+        })
+        .catch(() => {
+          store.dispatch('logout')
```
ulfgebhardt, line 28: we could "hide" stuff here by doing:

```suggestion
          store.dispatch('not-found')
```

ulfgebhardt (approved):

frontend/src/routes/guards.js
Mogge: ?

PR #1175, opened 2021-12-09
We need to store 1000 and not 10000000 in the db

Elweyn:
so we divide it by 10000.

### Issues

- None

### Todo

- [X] None


Mogge (changes requested):
No. Please close

ulfgebhardt (changes requested):
This is incorrect according to my research

Elweyn:
https://github.com/gradido/gradido/pull/1181

PR #1179, opened 2021-12-09
wrong_login_is_not_password_reset

ulfgebhardt:
## 🍰 Pullrequest

an incorrect login should always result in an error not in a request to reset the password

### Issues

- None

### Todo

- [X] None


frontend/src/views/Pages/Login.vue
```diff
@@ -104,13 +104,8 @@ export default {
...
         })
-        .catch((error) => {
-          if (error.message.includes('No user with this credentials')) {
-            this.$toasted.global.error(this.$t('error.no-account'))
-          } else {
-            // : this.$t('error.no-email-verify')
-            this.$router.push('/reset/login')
-          }
+        .catch(() => {
+          this.$toasted.global.error(this.$t('error.no-account'))
```
Elweyn, line 108: We need the push to /reset/login in case the user has not activated his account yet.
```suggestion
        .catch((error) => {
          this.$toasted.global.error(this.$t('error.no-account'))
          if (error.message.uncludes('User email not validated')) {
            this.$router.push('/reset/login')
          }
```

Elweyn (changes requested):
My logic was not right to only check on No user with this credentials but we need to send the user to /reset/login in a specific case.

frontend/src/views/Pages/Login.vue
ogerly: it may be better to check for locales ... if the text should ever change ... for whatever reason :)  
      ```
       if (error.message.uncludes(this.$t('error.no-account'))) {
            this.$router.push('/reset/login')
       }
       ```

Elweyn (approved):

Mogge (changes requested):
Test

Mogge (approved):

Elweyn:
If the user has not activate his email he gets redirected to:

Bildschirmfoto 2021-12-13 um 17 03 34

PR #1180, opened 2021-12-10
fix_i18n

ulfgebhardt:
## 🍰 Pullrequest

fix the i18n to load the keys again - with the bable transform this no longer happend

### Issues

- None

### Todo

- [X] None


Elweyn (approved):
I can not judge this really well.

PR #1182, opened 2021-12-10
state-balance balanceDate can't get the date of the last transaction

Elweyn:
We need to store the creationDate date.

### Issues

- None

### Todo

- [X] None


backend/src/graphql/resolver/AdminResolver.ts
```diff
@@ -174,7 +175,11 @@ export class AdminResolver {
     if (!lastUserTransaction) {
       newBalance = 0
     } else {
-      newBalance = lastUserTransaction.balance
+      newBalance = await calculateDecay(
+        lastUserTransaction.balance,
+        lastUserTransaction.balanceDate,
+        new Date(),
```
Mogge, line 181: Should this date be the same as in line 204?

Elweyn: Change it store ``new Date()`` in a ``variable`` and use it everywhere where we used ``new Date()``

ulfgebhardt (commented):
I do not really understand this change - you align the dates? Is this because this is more accurate or because else things don't work properly?

ulfgebhardt (approved):

Issue #1165, opened 2021-12-08
🔧 [Refactor] Admin issues regarding code quality & style


ulfgebhardt:
## 🔧 Refactor ticket

The new admin interface still have some problems with its code quality & style. There is still parts of the logic in the frontend where there should be only display related stuff on the frontend.

The following Problems are currently known:
- [x] There is integers used instead of floats for the gdd values.
- [x] There is still multiplications and divisions by 10000 - this has to be done in the backend
- [x] There is `array.reverse` calls to change order of arrays. This has to be done on the backend - even better on the database.
- [x] There is a GraphQL-Error occurring in the admin frontend in specific conditions
- [x] Reduce database calls #1164 
- [ ] Pagination required?

```Float instead of Int, 1000 auf Edit muss in Backend passieren, Reverse call auf array in Frontend, Fail to find Entity repository.findOneOrFail()```

PR #1185, opened 2021-12-13
Moved reverse of the getPendingCreations to the AdminResolver instead…

Elweyn:
… to do it in the CreationConfirm page.

### Issues

- #1165 

### Todo

- [X] Remove 10000 of Frontend & Admin
- [X] Move array.reverse to backend
- [x] Use floats instead of integer 
- [x] Remove GraphQL error

### DIFF
--- a/admin/src/graphql/createPendingCreation.js
+++ b/admin/src/graphql/createPendingCreation.js
@@ -3,7 +3,7 @@ import gql from 'graphql-tag'
 export const createPendingCreation = gql`
   mutation (
     $email: String!
-    $amount: Int!
+    $amount: Float!
     $memo: String!
     $creationDate: String!
     $moderator: Int!
--- a/admin/src/pages/CreationConfirm.vue
+++ b/admin/src/pages/CreationConfirm.vue
@@ -53,27 +53,32 @@ export default {
     removeConfirmResult(e, event) {
       let index = 0
       const findArr = this.confirmResult.find((arr) => arr.id === e.id)
-      if (event === 'remove') {
-        this.$apollo
-          .mutate({
-            mutation: deletePendingCreation,
-            variables: {
-              id: findArr.id,
-            },
-          })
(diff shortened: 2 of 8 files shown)


ulfgebhardt (approved):

admin/src/graphql/createPendingCreation.js
```diff
@@ -3,7 +3,7 @@ import gql from 'graphql-tag'
 export const createPendingCreation = gql`
   mutation (
     $email: String!
-    $amount: Int!
+    $amount: Float!
```
Mogge, line 6: I do not like this at all. `amount` is definitely an integer. Did you try `parseInt(amount)` when setting the variables for the mutation?

admin/src/pages/CreationConfirm.vue
```diff
@@ -70,6 +70,10 @@ export default {
           .catch((error) => {
             this.$toasted.error(error.message)
           })
+      } else {
```
Mogge, line 73: I guess this is the case `confirmed`. Can we have a `switch` here that throws an error when no case matches?

Mogge (changes requested):

admin/src/graphql/createPendingCreation.js
Elweyn: I though that it was something that we had to change!?
We want to create 1000,50 GDD or do we want to fix it to integers?

admin/src/pages/CreationConfirm.vue
Elweyn: Yeah this would be a good alternativ

admin/src/graphql/createPendingCreation.js
Elweyn: Was the point "Use float instead of integer" What did we mean with this?

Mogge (approved):

PR #1187, opened 2021-12-13
feat: Catch No Cookies Allowed

### LINKED ISSUE #1063: 🐛 [Bug] No cookies - no page
## 🐛 Bugreport

If you have cookies on localhost (or any other domain running the wallet) disabled, the page stays empty. The user is left without a hint what to do to fix it.

We should advice the user in this case to enable cookies and that our page is (for now) without function for him if he does not do so.

### To reproduce:
1. Navigate to http://localhost/vue
2. Delete LocalStorage
3. Forbid cookies on localhost
4. (shift) Reload Page

### Workaround:
Enable cookies

### Related:
With an empty storage there is several error appearing on the console:

```
vue.runtime.esm.js?2b0e:619 [Vue warn]: Error in render: "TypeError: Cannot read properties of null (reading 'name')"

found in

---> <Login> at src/views/Pages/Login.vue
       <AuthLayoutGDD> at src/views/Layout/AuthLayout_gdd.vue
         <App> at src/App.vue
           <Root>
```

```
vue.runtime.esm.js?2b0e:1888 TypeError: Cannot read properties of null (reading 'name')
    at Proxy.render (Login.vue?d29d:79)
    at VueComponent.Vue._render (vue.runtime.esm.js?2b0e:3548)
    at VueComponent.updateComponent (vue.runtime.esm.js?2b0e:4066)
    at Watcher.get (vue.runtime.esm.js?2b0e:4479)
    at new Watcher (vue.runtime.esm.js?2b0e:4468)
    at mountComponent (vue.runtime.esm.js?2b0e:4073)
    at VueComponent.Vue.$mount (vue.runtime.esm.js?2b0e:8415)
    at init (vue.runtime.esm.js?2b0e:3118)
    at merged (vue.runtime.esm.js?2b0e:3301)
    at createComponent (vue.runtime.esm.js?2b0e:5978)
```

Furthermore:
- [ ] Remove reference from Readme.md

Mogge:
## 🍰 Pullrequest
Dirty hack to catch error when store can not be created because cookies are not allowed. We need a URL outside the application that explains the problem to the user.

### Issues
- fixes #1063 

### Todo
- [ ] Supply a support page (e.g. on gradido.net) that explains the user what just happened and how to fix it.


frontend/src/main.js
```diff
@@ -30,6 +30,11 @@ loadAllRules(i18n)
 
 addNavigationGuards(router, store, apolloProvider.defaultClient)
```
Elweyn, line 31: What happens if store is null or undefined here? Or is this not the behavior when blocking everything?

frontend/src/main.js
```diff
@@ -30,6 +30,11 @@ loadAllRules(i18n)
...
 
+if (!store) {
+  // Please supply a support page
+  window.location.assign('https://gradido.net/')
```
Elweyn, line 35: Like @ulfgebhardt said yesterday, would be prettier if we could have a page in our application

frontend/src/store/store.js
```diff
@@ -78,32 +78,38 @@ export const actions = {
...
-  actions,
-})
+    getters: {},
+    // Syncronous mutation of the state
+    mutations,
+    actions,
+  })
+} catch (error) {}
+
+export { store }
```
Elweyn, line 115: If we go in catch store == null?

frontend/src/store/store.js
```diff
@@ -78,32 +78,38 @@ export const actions = {
...
-  // Syncronous mutation of the state
-  mutations,
-  actions,
-})
+    getters: {},
+    // Syncronous mutation of the state
+    mutations,
+    actions,
+  })
+} catch (error) {}
```
ulfgebhardt, line 113: lets not do this - its better to have the error on the console in case things go wrong?! e.g. esoteric browsers like edge behave completely irrational. Or does then to redirect not trigger because the js dies? if so please log the error to the console here in order to allow debugging

ulfgebhardt (approved):

PR #1193, opened 2021-12-16
[ Tutorial ] decay start block seed after first transactions

ogerly:
## 🍰 Pullrequest
This PR does not need to be memorised, it describes how to get to the decay start block in the developer environment. 

- important is that the database/src/index.ts  
  Line await runSeeder(DecayStartBlockSeed)
after the runSeeder() functions of the example user.

 - and that the date in the database/src/seeds/decay-start-block.seed.ts line 13 is set to one month later. 
 
 
 

Bildschirmfoto von 2021-12-16 07-08-24

Bildschirmfoto von 2021-12-16 08-16-40

1. cd database 
2. yarn dev_down
3. database/src/seeds/decay-start-block.seed.ts und das datum ändern .. (einen monat höher)
4. yarn seed 
5. mit bibi@blocksberg.de einloggen und eine transaktion an peter@lustig.de machen. 
die erste transaktion sollte dann den Startblock Vergänglichkeit enthalten.

Bildschirmfoto von 2021-12-16 07-27-18

Bildschirmfoto von 2021-12-16 07-28-31

### DIFF
--- a/database/src/index.ts
+++ b/database/src/index.ts
@@ -60,11 +60,11 @@ const run = async (command: string) => {
         root: process.cwd(),
         configName: 'ormconfig.js',
       })
-      await runSeeder(DecayStartBlockSeed)
       await runSeeder(CreatePeterLustigSeed)
       await runSeeder(CreateBibiBloxbergSeed)
       await runSeeder(CreateRaeuberHotzenplotzSeed)
       await runSeeder(CreateBobBaumeisterSeed)
+      await runSeeder(DecayStartBlockSeed)
       break
     default:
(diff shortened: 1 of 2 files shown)


ulfgebhardt:
@Mogge What you say? We need transaction data with the decay block in between - as I see this, that is not really possible at the moment, since the transaction or the decay blocks are seeded first?!

ulfgebhardt:
@Mogge We should have test data with transactions before the decay block. Even tho its only important for the gradido academy community - its the only one we currently have.

Issue #1196, opened 2021-12-16
🚀 [Feature] Admin interface Usermanagement - Register Mail send


ogerly:
## 🚀 Feature
it should be visible for the user search / user individual view in the admin area:

### Admin
- [ ] whether a user has completed the registration process: yes (green), no (red)
- [ ] if no: send an additional activation mail from the admin area to the user by a moderator / administrator
- [ ] filter search all inactive accounts 

### Frontend
- [ ] In frontend on 'thx/login/' page add a send activation mail button

### backend
- [ ] Send activation link mail to user in backend

PR #1199, opened 2021-12-18
1165 admin issues regarding code quality syle

### LINKED ISSUE #1165: 🔧 [Refactor] Admin issues regarding code quality & style

ogerly:
## 🍰 Pullrequest

unused elements removed. click function for show details for a user in the user table 

### Issues
- fixes #1165

### DIFF
--- a/admin/src/components/CreationFormular.vue
+++ b/admin/src/components/CreationFormular.vue
@@ -204,8 +204,6 @@ export default {
         // Die anzahl der Mitglieder aus der Mehrfachschöpfung
         const i = Object.keys(this.itemsMassCreation).length
         // hinweis das eine Mehrfachschöpfung ausgeführt wird an (Anzahl der MItgleider an die geschöpft wird)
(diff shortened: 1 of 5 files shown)


ogerly:
inhalt gemergt in #1196

PR #1212, opened 2021-12-21
handling email sending connection error 

einhornimmond:
- not sending a email if an error by sending email occured
- by loosing connection to email server, wait 5 minutes and then try to reconnect and resend email
- try resending email only 3 times


login_server/src/cpp/SingletonManager/EmailManager.cpp
```diff
@@ -141,6 +141,33 @@ int EmailManager::ThreadFunction()
...
+					// better wait instead and try again, it seems that the strato mail server sometimes discard or connection
+					// MAGIC NUMBER: sleep time if email sending failed (ssl connection was discarded from mailserver)
+					// wait 5 minute for the next try
+					Poco::Thread::sleep(5 * 60 * 1000);
+					// reconnect to mailserver
+					mailClientSession.login();
+					try {
+						mailClientSession.startTLS(ServerConfig::g_SSL_CLient_Context);
+						mailClientSession.login(Poco::Net::SMTPClientSession::AUTH_LOGIN, mEmailAccount.username, mEmailAccount.password);
+					}
```
ulfgebhardt, line 156: You create duplicate code here - this code part is already present in the top of the function. In general you should avoid that to make it easy to maintain. Now you have to maintain both if something changes in this block, which leads to potential errors in the future.

Furthermore I find it quite weird - or not self explanatory that you call
`mailClientSession.login();` and then `mailClientSession.login(Poco::Net::SMTPClientSession::AUTH_LOGIN, mEmailAccount.username, mEmailAccount.password);`. I did not check this, but I assume that calling `login()` without parameters opens the connections?! It is definitely not self-explanatory which all code should be at all time 🙄

login_server/src/cpp/SingletonManager/EmailManager.cpp
```diff
@@ -141,6 +141,33 @@ int EmailManager::ThreadFunction()
 				mailClientSession.sendMessage(mailMessage);
 					email_sended = true;
 				}
+				catch (Poco::Net::SSLConnectionUnexpectedlyClosedException& ex) {
+					// it is a mad idea to send an email if the email sending failed
+					// errors.sendErrorsAsEmail();
+					// better wait instead and try again, it seems that the strato mail server sometimes discard or connection
+					// MAGIC NUMBER: sleep time if email sending failed (ssl connection was discarded from mailserver)
+					// wait 5 minute for the next try
+					Poco::Thread::sleep(5 * 60 * 1000);
```
ulfgebhardt, line 150: I would define a constant here `EMAIL_SEND_RETRY_TIME` or alike. I think 5 minutes with 4 retries is quite alot for a connection issue.

login_server/src/cpp/SingletonManager/EmailManager.cpp
```diff
@@ -141,6 +141,33 @@ int EmailManager::ThreadFunction()
...
+						mailClientSession.login(Poco::Net::SMTPClientSession::AUTH_LOGIN, mEmailAccount.username, mEmailAccount.password);
+					}
+					catch (Poco::Net::SSLException& ex) {
+						errors.addError(new ParamError(function_name, "ssl certificate error", ex.displayText()));
+						printf("[PrepareEmailTask] ssl certificate error: %s\nPlease make sure you have cacert.pem (CA/root certificates) next to binary from https://curl.haxx.se/docs/caextract.html\n", ex.displayText().data());
+						return -1;
+					}
+					// retry only if not to many retries are already tried
+					// MAGIC NUMBER: Retry count if email sending failed
+					if (email->checkResendCounter() < 4) {
```
ulfgebhardt, line 164: I would define a constant for this `MAX_EMAIL_RESEND_COUNT`. We use the convention to CAPSLOCK CONSTANTS in order to easily identify them.

login_server/src/cpp/SingletonManager/EmailManager.cpp
```diff
@@ -141,6 +141,33 @@ int EmailManager::ThreadFunction()
 				mailClientSession.sendMessage(mailMessage);
 					email_sended = true;
 				}
+				catch (Poco::Net::SSLConnectionUnexpectedlyClosedException& ex) {
+					// it is a mad idea to send an email if the email sending failed
+					// errors.sendErrorsAsEmail();
+					// better wait instead and try again, it seems that the strato mail server sometimes discard or connection
```
ulfgebhardt, line 147: I would assume that this connection trouble is a problem on our side - strato is a professional hoster and certainly will be able to handle many connections without running into bootlnecks. It could be a rate-limit thing that blocks our service, since it sends too many EMail? This would be an interesting topic to research.

login_server/src/cpp/model/email/Email.cpp
```diff
@@ -156,6 +156,7 @@ Gradido Login-Server\n\
 		auto em = EmailManager::getInstance();
 		auto adminRecipient = Net::MailRecipient(Net::MailRecipient::PRIMARY_RECIPIENT, em->getAdminReceiver());
 		Poco::AutoPtr<model::table::User> userTableModel;
+		mResendCounter++;
```
ulfgebhardt, line 159: This is probably the most problematic code section - you count up the resend count at a place where neither the sending nor resending of EMails takes place - its a different file tho.
I understand that this makes it quite easy regarding lines of codes to be changed, but it also can have sie effects. Are you sure there will be 5 tries to send the EMail or will some/all emails call the draft method when creating the email and therefore start with `mResendCounter == 1`.
I would expect the counter to go up where the resend actually is qued.

ulfgebhardt (approved):
I made some remarks and gave you a thoroughly review as requested. The only possible mistake I saw (besides all the many comments) is the question if you are sure that the email is attempted to be sent 5 times or only 4 times?!

ulfgebhardt:
This PR is based on the wrong base. The correct base branch would be https://github.com/gradido/gradido/commits/login_hotfixes_1.5.1_reopen

login_server/src/cpp/SingletonManager/EmailManager.cpp
einhornimmond: "You create duplicate code here - this code part is already present in the top of the function. In general you should avoid that to make it easy to maintain. Now you have to maintain both if something changes in this block, which leads to potential errors in the future." - weiß ich, ich hatte schon überlegt ob ich es in eine separate Funktion auslagere und damit alles noch ein bisschen komplizierter machen soll. Aber ich habe mich dagegen entschieden. Im Regelfall bin ich mir bewusst wenn ich doppelten Code habe und refactore das spätestens wenn ich eine Änderung in diesem Code machen muss.

einhornimmond: "Furthermore I find it quite weird - or not self explanatory that you call
mailClientSession.login(); and then mailClientSession.login(Poco::Net::SMTPClientSession::AUTH_LOGIN, mEmailAccount.username, mEmailAccount.password);. I did not check this, but I assume that calling login() without parameters opens the connections?! It is definitely not self-explanatory which all code should be at all time 🙄"

Das habe ich einfach aus dem Beispiel übernommen ohne das weiter zu hinterfragen:
https://github.com/pocoproject/poco/blob/master/NetSSL_OpenSSL/samples/Mail/src/Mail.cpp

einhornimmond: From poco docu for Poco::Net::SecureSMTPClientSession:
```c++
void login(); 
                /// Calls login(hostname) with the current host name.
bool startTLS(Context::Ptr pContext);
		/// Sends a STARTTLS command and, if successful,
		/// creates a secure SSL/TLS connection over the
		/// existing socket connection.
		///
		/// Uses the given Context object for creating
		/// the SSL/TLS connection.
		///
		/// Must be called after login() or login(hostname).
		/// If successful, login() can be called again
		/// to authenticate the user.
		///
		/// Returns true if the STARTTLS command was successful,
		/// false otherwise.
void login(LoginMethod loginMethod, const std::string& username, const std::string& password);
		/// Logs in to the SMTP server using the given authentication method and the given
		/// credentials.
```

login_server/src/cpp/SingletonManager/EmailManager.cpp
einhornimmond: Yes and than it has maybe a cooldown and because of that I have choosen 5 minutes a relativ long time

login_server/src/cpp/model/email/Email.cpp
einhornimmond: I found it always hard to understand code changes only from the diff, I need the whole picture.
When you look at the whole picture your understand. 
You know I work object oriented with c++.
Mails are handled as own objects, if some part of the programm likes to send a email it creates an email object, add every needed data to it which is needed to create the actual email and push it to the email manager queue. 

The Email Manager get the first email from the queue, call draft and try to send the email. 
If the connection failed, now the email will be put back into the queue.

And yes I have checked that draft is only called from EmailManager.

If I would call mResendCounter++ after or before push I must put it into two places (double code), or I must call inside the EmailManager::ThreadFunction() the addEmail() function which will call condSignal() {condition.signal() }. 
I don't know the side effects from calling condition signal inside the thread which will be waiting on the condition.

ulfgebhardt (changes requested):
Removing my approve

PR #1215, opened 2021-12-27
Remove confirmation of own pending creation

Elweyn:
## 🍰 Pullrequest
A Moderator can prepare some pending creation for his own, but he is not allowed to confirm them!

### Issues
- None

### Todo
- [X] None


ogerly (approved):
i confirm the behaviour. 
i have made a creation as admin and cannot confirm it. i get an error message from the backend.

backend/src/graphql/resolver/AdminResolver.ts
```diff
@@ -66,7 +66,7 @@ export class AdminResolver {
     return await getUserCreations(user.id)
   }
 
-  // @Authorized([RIGHTS.SEARCH_USERS])
+  @Authorized([RIGHTS.SEARCH_USERS])
```
ulfgebhardt, line 69: This seems to be the wrong right - every function gets its own right.

backend/src/graphql/resolver/AdminResolver.ts
```diff
@@ -94,24 +94,9 @@ export class AdminResolver {
...
-    //   loginPendingTaskAdmin.date = creationDateObj
-    //   loginPendingTaskAdmin.memo = memo
-    //   loginPendingTaskAdmin.moderator = moderator
-    //
-    //   pendingCreationRepository.save(loginPendingTaskAdmin)
-    // }
-    // return await getUserCreations(user.id)
   }
 
+  @Authorized([RIGHTS.SEARCH_USERS])
```
ulfgebhardt, line 99: This seems to be the wrong right - every function gets its own right.

backend/src/graphql/resolver/AdminResolver.ts
```diff
@@ -139,6 +124,7 @@ export class AdminResolver {
     return pendingCreationsPromise.reverse()
   }
 
+  @Authorized([RIGHTS.SEARCH_USERS])
```
ulfgebhardt, line 127: This seems to be the wrong right - every function gets its own right.

backend/src/graphql/resolver/AdminResolver.ts
```diff
@@ -147,11 +133,17 @@ export class AdminResolver {
     return !!res
   }
 
+  @Authorized([RIGHTS.SEARCH_USERS])
```
ulfgebhardt, line 136: This seems to be the wrong right - every function gets its own right.

ulfgebhardt (commented):
Looks good besides the rights thingy

backend/src/graphql/resolver/AdminResolver.ts
Elweyn: OK, I was not sure while implementing it, I understood that SEARCH_USERS is the actual Admin role / right

backend/src/graphql/resolver/AdminResolver.ts
Elweyn: OK, I was not sure while implementing it, I understood that SEARCH_USERS is the actual Admin role / right

backend/src/graphql/resolver/AdminResolver.ts
Elweyn: OK, I was not sure while implementing it, I understood that SEARCH_USERS is the actual Admin role / right

backend/src/graphql/resolver/AdminResolver.ts
Elweyn: OK, I was not sure while implementing it, I understood that SEARCH_USERS is the actual Admin role / right

PR #1271, opened 2022-01-10
Admin Logout redirectes to the wallet login page.

### LINKED ISSUE #1241: 🐛 [Bug] Admin Interface Logout sends to 404, change link to Wallet login
## 🐛 Bugreport

When an user logout in the Admin Interface he is directed to a page that he can't see.
Better would be to send the user to the Wallet login page.

## TODOs
- [ ] Logout button change link to $(server_url)/login

Elweyn:
## 🍰 Pullrequest
When admin user logout they are send to a 404 not found, now redirects to the wallet-login page

### Issues

- fixes #1241

### Todo

- [X] Admin Logout redirects to Wallet Login

### DIFF
--- a/admin/src/components/NavBar.vue
+++ b/admin/src/components/NavBar.vue
@@ -33,8 +33,8 @@ export default {
   name: 'navbar',
   methods: {
     logout() {
+      window.location = CONFIG.WALLET_URL
       this.$store.dispatch('logout')
-      this.$router.push('/logout')
     },
     wallet() {
       window.location = CONFIG.WALLET_AUTH_URL.replace('$1', this.$store.state.token)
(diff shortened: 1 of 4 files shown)


admin/src/components/NavBar.vue
```diff
@@ -33,8 +33,8 @@ export default {
   name: 'navbar',
   methods: {
     logout() {
+      window.location = CONFIG.WALLET_URL
```
Mogge, line 36: I am not sure that the dispatch is executed after changing the window location. Furthermore we use `window.location.assign` so we are able to test it. See in frontend `Layout/DashboardLayout_gdd.vue`

Mogge (changes requested):
Please test the redirect and proof that logout is dispatched.

Mogge (approved):

PR #1285, opened 2022-01-12
Fix redirect to /logout, now redirects to the wallet login.

Elweyn:
## 🍰 Pullrequest

When the token expires the user is always send to a /logout page that has a not found page behind it.
Sends the user to the login page.

### Todo

- [X] Token expiration redirect to Wallet login


admin/src/plugins/apolloProvider.test.js
```diff
@@ -130,18 +135,13 @@ describe('apolloProvider', () => {
 
       describe('current route is not logout', () => {
         it('redirects to logout', () => {
-          expect(routerPushMock).toBeCalledWith('/logout')
+          expect(windowLocationMock).toBeCalledWith('http://localhost/vue/login')
         })
       })
 
       describe('current route is logout', () => {
```
Mogge, line 142: Please remove. We do not have any route anymore in the code.

Mogge: Also remove the mocks of the route (lines 62 to 67)

Mogge (dismissed):
Please keep the test clean.

Mogge (approved):

PR #1294, opened 2022-01-16
gradido proxy

einhornimmond:
## 🍰 Pullrequest
A nginx config template for a gradido domain proxy. 

Example: we have the domain gdd1.gradido.com as or main domain but our server will be shutdown sometime and we like easly switch over to or backup server without waiting that all dns server has get the update for the new domain. 
The solution: We use nginx as proxy and route the gdd1.gradido.com from dns to our final production server.
The final production server then proxy all calls to our current production server which get a second domain: gdd1-1fire.gradido.com.
If we swap to the backup server we need only update the nginx config on the final production server and don't need to wait on dns server updates.

If we copy over the ssl keys from production to the final production server, we can proxy from final production to current production and keep the config on current production. So all user can reach the production server nevertheless if there request go to current production ip or final production server ip.


ulfgebhardt:
closing as no longer needed

PR #1324, opened 2022-01-20
import_production_data

### LINKED ISSUE #909: 💥  [Devops] Migrate Accounts for those with only EloPage & Data Migration
## :boom: [Devops]

Migrate Accounts for those with only EloPage & Data Migration

We have aprox. 2k Users who are only present on EloPage, but not on our Systems. With the change of the Registration Process, we should create an Account for those with only an EloPage Account.

ulfgebhardt:
## 🍰 Pullrequest

This PR transform the production data and adjusts code parts to handle the production data without errors.

### Issues

- fixes #909

### Todo

- [X] None


backend/src/graphql/resolver/UserResolver.ts
```diff
@@ -585,15 +588,23 @@ export class UserResolver {
...
+
+    // Generate Passphrase if needed
+    if (!loginUserBackup) {
+      const passphrase = PassphraseGenerate()
+      loginUserBackup = new LoginUserBackup()
+      loginUserBackup.userId = loginUser.id
+      loginUserBackup.passphrase = passphrase.join(' ') // login server saves trailing space
+      loginUserBackup.mnemonicType = 2 // ServerConfig::MNEMONIC_BIP0039_SORTED_ORDER;
+      loginUserBackupRepository.save(loginUserBackup)
+    }
```
Elweyn, line 601: I think would be prettier like this:
```suggestion
    const loginUserBackup = await loginUserBackupRepository
      .findOneOrFail({ userId: loginUser.id })
      .catch(() => {
        const passphrase = PassphraseGenerate()
        loginUserBackup = new LoginUserBackup()
        loginUserBackup.userId = loginUser.id
        loginUserBackup.passphrase = passphrase.join(' ') // login server saves trailing space
        loginUserBackup.mnemonicType = 2 // ServerConfig::MNEMONIC_BIP0039_SORTED_ORDER;
        loginUserBackupRepository.save(loginUserBackup)
      })
```

database/migrations/0008-state_users_plug_holes.ts
```diff
@@ -0,0 +1,23 @@
...
+ * to mitigate 38 out of 50 emails could be restored from
+ * login_users.
+ */
+
+export async function upgrade(queryFn: (query: string, values?: any[]) => Promise<Array<any>>) {
+  // Fill in missing emails from login_users
+  await queryFn(
+    `UPDATE state_users
+     INNER JOIN login_users ON state_users.public_key = login_users.pubkey
+     SET state_users.email = IF(state_users.email = '', login_users.email, state_users.email)
```
Elweyn, line 14: state_users.email should always be empty since the where clause only take state_users.email = ''.
Would the check not be better like this:
```suggestion
     SET state_users.email = IF(login_users.email != '', login_users.email, state_users.email)
```

Elweyn (changes requested):

backend/src/graphql/resolver/UserResolver.ts
ulfgebhardt: I like to keep it like I have it, since its easier to read and understand. Hope thats ok.

database/migrations/0008-state_users_plug_holes.ts
ulfgebhardt: Good catch!

Mogge (approved):

Elweyn (approved):

Issue #1311, opened 2022-01-18
🐛 [Bug] Month not Localized on Creation


Mogge:
## 🐛 Bugreport
![image](https://user-images.githubusercontent.com/15882241/149957645-bd9404a7-6e79-4fc3-b50b-51c3ce67dcb5.png)

The month is displayed in English though everything else is in German.

Issue #1326, opened 2022-01-20
🐛 [Bug] Wrong Date in Admin Interface for Email Last Send


Mogge:
## 🐛 Bugreport
We do not display the date stored in the server when the last email was sent.

User table:
`:dateLastSend="$moment().subtract(1, 'month').format('dddd, DD.MMMM.YYYY HH:mm'),"`

PR #1327, opened 2022-01-20
fix: Localize Datetime in Admin Interface

Mogge:
## 🍰 Pullrequest
Removes moment and vue-moment. Uses vue-i18n to localize datetime.

fixes: #1311 

### Todo

- [ ] Test carefully


admin/src/components/UserTable.vue
```diff
@@ -110,7 +110,7 @@
             <confirm-register-mail-formular
               :checked="row.item.emailChecked"
               :email="row.item.email"
-              :dateLastSend="$moment().subtract(1, 'month').format('dddd, DD.MMMM.YYYY HH:mm'),"
+              :dateLastSend="$d(new Date(), 'long')"
```
Mogge, line 113: see #1326

ulfgebhardt (commented):
I just did a peliminary review of your code. I did not test it yet - tho I have some remarks.

I suggest - this will be part of the feature set after the release. This way we do not increase the package on the wagon and have a nice refactor right afterwards. Would you agree on that or do we **need** this in the current release?

admin/src/components/CreationFormular.vue
```diff
@@ -298,6 +283,36 @@ export default {
...
+      return {
+        short: this.$d(this.now, 'month'),
+        long: this.$d(this.now, 'short'),
+        year: this.$d(this.now, 'year'),
+        date: this.$d(this.now, 'short', 'en'),
+      }
+    },
+    lastMonth() {
+      const now = new Date(this.now)
+      const lastMonth = new Date(now.getFullYear(), now.getMonth() - 1, 1)
```
Elweyn, line 297: What happens if now is 01.01.2022 does last Month change to 01.12.2021?

admin/src/components/CreationFormular.vue
```diff
@@ -298,6 +283,36 @@ export default {
...
+      return {
+        short: this.$d(lastMonth, 'month'),
+        long: this.$d(lastMonth, 'short'),
+        year: this.$d(lastMonth, 'year'),
+        date: this.$d(lastMonth, 'short', 'en'),
+      }
+    },
+    beforeLastMonth() {
+      const now = new Date(this.now)
+      const beforeLastMonth = new Date(now.getFullYear(), now.getMonth() - 2, 1)
```
Elweyn, line 307: What happens if now is 01.01.2022 does last Month change to 01.11.2021?

admin/src/components/CreationFormular.vue
Elweyn: The test written below seems to say yes

admin/src/components/CreationFormular.vue
Elweyn: The test written below seems to say yes

admin/src/components/EditCreationFormular.vue
```diff
@@ -227,9 +216,36 @@ export default {
...
+    beforeLastMonth() {
+      const now = new Date(this.now)
+      const beforeLastMonth = new Date(now.getFullYear(), now.getMonth() - 2, 1, 0)
+      return {
+        short: this.$d(beforeLastMonth, 'month'),
+        long: this.$d(beforeLastMonth, 'short'),
+        date: this.$d(beforeLastMonth, 'short', 'en'),
+      }
+    },
+  },
```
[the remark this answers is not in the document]
Elweyn: Was thinking the same, but it is not exactly the same code

Elweyn (approved):

admin/package.json
```diff
@@ -12,7 +12,7 @@
     "dev": "yarn run serve",
     "build": "vue-cli-service build",
     "lint": "eslint --ext .js,.vue .",
-    "test": "jest --coverage",
+    "test": "TZ=UTC jest --coverage",
```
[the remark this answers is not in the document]
Mogge: I like this one more. I did not get the jest config working the way I did in the backend.

admin/src/components/CreationFormular.vue
```diff
@@ -298,6 +283,36 @@ export default {
         })
     },
   },
+  computed: {
+    currentMonth() {
+      return {
+        short: this.$d(this.now, 'month'),
+        long: this.$d(this.now, 'short'),
+        year: this.$d(this.now, 'year'),
+        date: this.$d(this.now, 'short', 'en'),
```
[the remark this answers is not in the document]
Mogge: `lastMonth` and `beforeLastMonth` have the first day of the month at 0:00h locale time. So when passing the date to the server with UTC it becomes the last day of the month before at 23:00h when local time is CET. (`2021-11-01T00:00:00 +01:00` becomes `2021-10-31T23:00:00`). So we have to strip the time when passing the date to the server. German date: `01-11-2021` as string becomes `11-01-2021` on server. Therefore I pass always the English date to the server as string.

admin/src/components/EditCreationFormular.vue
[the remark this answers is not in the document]
Mogge: I agree. I think of a mixin

admin/src/components/CreationFormular.vue
Mogge: Yes. You can try that out on the console.

Mogge:
```
const date = new Date('2022-01-01')
Date Sat Jan 01 2022 01:00:00 GMT+0100 (Central European Standard Time)
const newDate = new Date(date.getFullYear(), date.getMonth() -1, 1)
Date Wed Dec 01 2021 00:00:00 GMT+0100 (Central European Standard Time)
```

Mogge:
> I just did a peliminary review of your code. I did not test it yet - tho I have some remarks.
> 
> I suggest - this will be part of the feature set after the release. This way we do not increase the package on the wagon and have a nice refactor right afterwards. Would you agree on that or do we **need** this in the current release?

This means we accept a total mess of dates displayed in the admin interface. The user does not know in which locale the date is displayed.

Elweyn:
Would also prefer to see this before the deployment but if we can't do it before then introduce it as first bug fix!

PR #1328, opened 2022-01-20
fix: Set Locale after Login

Mogge:
## 🍰 Pullrequest
dito


frontend/src/views/Pages/Login.vue
```diff
@@ -101,6 +102,8 @@ export default {
             data: { login },
           } = result
           this.$store.dispatch('login', login)
+          this.$i18n.locale = login.language
+          localeChanged(login.language)
```
ulfgebhardt, line 106: should that be part of `store.dispatch.login` ? Or do we violate some rules there? Because if i am not mistaken we need to set the language everytime we login.

Mogge: Where else do we login than here?

Elweyn: I would also prefer to see this code in the dispatch methode...
@Mogge for now it would be OK but what if we add new entry points? Login with cookies instead of the form!?
Then it would be a mess since we would need to set this code in the second method also

Elweyn (approved):

frontend/src/views/Pages/Login.vue
Mogge: I do not agree. When the locale is changed, we do not dispatch login but commit language. The dispatch of login commits the language

PR #1329, opened 2022-01-20
fix_rejected_login

### LINKED ISSUE #1305: 🐛 [Bug] Error Message on first login
## 🐛 Bugreport

I often get an error message when trying to login with correct credentials - on the second try it works. Could this be redirect related @Mogge ? Can you reproduce the problem? It seems to happen when building a new version of the frontend - probl. with exist local storage

Beim 1. Versuch als peter@lustig.de einzuloggen kam Fehlermeldung.
Das 2. Mal ging es dann.

image

ulfgebhardt:
## 🍰 Pullrequest

fix broken login with existing token in local store

### Issues

- fixes #1305

### Todo

- [X] None


Mogge (approved):
Good catch. Great solution. I tested it locally and it works.

PR #1362, opened 2022-01-28
clean_database

ulfgebhardt:
## 🍰 Pullrequest

Clean the Database and remove unused tables and refactor thing on database level.

Tables before: 40
Tables after: 16

1. drop empty tables without an entity model (12)
2. drop unused tables with static data and without entity model (12)

### Issues

- None

### Todo

- [X] None


Mogge (changes requested):
Error: Wrong database version - the backend requires '0012-login_user_backups_unify_wordlist' but found '0013-drop_unused_tables.ts'

Mogge (approved):
I really do like that we save the data junk in our codebase :)

ulfgebhardt:
I agree, but otherwise its not 100% correct. We might consider to clean this up at some point - this way all the data stays available till we have a clean database

Issue #1373, opened 2022-01-29
🔧 [Refactor] ADMINAREA - check email confirmed before creation and multi creation saved


ogerly:
## 🔧 Refactor ticket
check on the backend page if the email is confirmed before a creation and multi creation , if not email confirmed, no creation may be performed and an error with info will be returned.

ogerly:
when confirming the transaction in the multiple creation, it is checked whether email is activated. in the current master, it is no longer possible to select a user for multiple creation who does not have a confirmed email.

PR #1374, opened 2022-01-29
if an email is not confirmed, a user cannot be added to any multiple …

### LINKED ISSUE #1372: 🔧 [Refactor] ADMINAREA - not activated email not be selectable for multiple creation
## 🔧 Refactor ticket
If a user has not activated their email they should not be selectable for multiple creation. 

Wenn ein Nutzer seine E-Mail nicht aktiviert hat sollte er nicht für die Mehrfachschöpfung auswählbar sein.

ogerly:
## 🍰 Pullrequest
if an email is not confirmed, a user cannot be added to any multiple creation. 
If there is no confirmed email for a user ... then **Email!** is written in the bookmar field.

### Issues
- fixes #1372 

FireShot Capture 956 - Gradido Admin Interface - localhost

TODO: https://github.com/gradido/gradido/issues/1373 check on the backend page if the email is confirmed before a creation, if not, no creation may be performed and an error with info will be returned.


admin/src/components/UserTable.vue
```diff
@@ -125,15 +125,18 @@
...
+          <b-button
+            v-if="row.item.emailChecked"
+            variant="warning"
+            size="md"
+            @click="bookmarkPush(row.item)"
+            class="mr-2"
+          >
+            <b-icon icon="plus" variant="success"></b-icon>
+          </b-button>
+          <div v-else>E-Mail!</div>
```
Mogge, line 138: Please localize.

Mogge (changes requested):

admin/src/components/UserTable.vue
ogerly: ok

Mogge (approved):

PR #1404, opened 2022-02-01
cleanups_refactors

ulfgebhardt:
## 🍰 Pullrequest

Remove Typeorm from backend in order to have only a single connection manager handling our connection. We can now use activeRecord and dataMapper pattern at the same time.

Other cleanups (Missing right check, code simplification, separate db version error and version info, define db version in config, type definitions and lint exception removal)

image

### Issues

- None

### Todo

- [X] None


backend/src/server/createServer.ts
```diff
@@ -58,15 +53,6 @@ const createServer = async (context: any = serverContext): Promise<any> => {
   // bodyparser urlencoded for elopage
   app.use(express.urlencoded({ extended: true }))
```
Elweyn, line 54: express same as Express? Why do we want to hold express?

backend/src/server/createServer.ts
```diff
@@ -1,15 +1,12 @@
-/* eslint-disable @typescript-eslint/no-explicit-any */
-/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
-
 import 'reflect-metadata'
 import 'module-alias/register'
 
 import { ApolloServer } from 'apollo-server-express'
-import express from 'express'
+import express, { Express } from 'express'
```
Elweyn, line 5: express same as Express? Why do we want to hold express if Express is used?

Elweyn: OK I answered myself the question: Express is the object and is needed for ServerDef! And express is also needed!

Elweyn (approved):

Elweyn:
Code runs on my local docker instance after a new build! 
Creation and transaction were successful! 

In my eyes it's a clean PR.

PR #1430, opened 2022-02-07
maria_db_docker_image

ulfgebhardt:
## 🍰 Pullrequest

since we do not do anything with the mariadb docker image, we just use the latest vendor one

### Issues

- None

### Todo

- [X] None


ulfgebhardt:
this seems too complicated - we keep it as it is

PR #1443, opened 2022-02-09
refactor: Split User Table Component in Admin Interface

### LINKED ISSUE #1417: 🐛 [Bug] Double Overlay in User Table
## 🐛 Bugreport
The User Table includes an overlay to confirm or cancel an action. This overlay has an id `overlay`. When using mass creation, we have to user tables and therefore two overlays with the same id. This makes it hard to test. Furthermore an id should always be unique.

### LINKED ISSUE #1434: 🔧 [Refactor] Split User Table Component in Components
## 🔧 Refactor ticket
We have three pages that user the User Table component:
 1. Creation Confirm
 2. Creation (2x)
 3. User Search

We should have four different components for the table:
 1. SearchUserTable
 2. SelectUserTable (left side of Creation)
 3. SelectedUsersTable (right side of Creation)
 4. ConfirmCreationTable

These components should only render the incoming data and emit the events to the parent component.

This way the components are easier to test, to improve and to add new features.

Mogge:
## 🍰 Pullrequest
Splits the user table component in four table components.

### Issues
- fixes #1434 
- fixes #1417 

### Todo

- [X] None


ulfgebhardt (approved):
Tested locally with production data - looks good, code seems very clean, good work ❤️

PR #1444, opened 2022-02-09
enforce_no_element_style

ulfgebhardt:
## 🍰 Pullrequest

Enforce style not to be defined on the element but external at least in a `<style>` tag

### Issues

- None

### Todo

- [X] None

### DIFF
--- a/admin/src/pages/UserSearch.vue
+++ b/admin/src/pages/UserSearch.vue
@@ -1,6 +1,6 @@
 <template>
   <div class="user-search">
-    <div style="text-align: right">
+    <div id="user-search-style-1">
       <b-button block variant="danger" @click="unconfirmedRegisterMails">
         <b-icon icon="envelope" variant="light"></b-icon>
         {{ filterCheckedEmails ? $t('all_emails') : $t('unregistered_emails') }}
@@ -116,3 +116,8 @@ export default {
   },
 }
 </script>
+<style>
+#user-search-style-1 {
+  text-align: right;
+}
+</style>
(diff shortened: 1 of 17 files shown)


admin/src/pages/UserSearch.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <div class="user-search">
-    <div style="text-align: right">
+    <div id="user-search-style-1">
```
Mogge, line 3: serious?

frontend/src/views/Pages/AccountOverview/GddTransactionList.vue
```diff
@@ -212,4 +208,18 @@ export default {
 .gdd-transaction-list-item {
   outline: none !important;
 }
+
+.red-icon {
+  color: red;
+}
+
+#collaps-button-style {
```
Mogge, line 216:
```suggestion
.collaps-button-style {
```

frontend/src/views/Pages/AccountOverview/GddTransactionList.vue
```diff
@@ -212,4 +208,18 @@ export default {
...
+  width: 95%;
+  position: absolute;
+}
+
+#collapse-div-style-1 {
```
Mogge, line 221:
```suggestion
.collapse-div-style-1 {
```

Mogge (changes requested):
Very quick and dirty :(

ulfgebhardt:
> Very quick and dirty :(

I think its adequate - it was you who said we do not invest into style, since we get a new design in the near future anyways. It is expected that all of those style- and the html tags will change completely. I believe its important to have this check in place before we do the design implementation.

PR #1448, opened 2022-02-10
fix: Admin Email Confirmation Date and Time

### LINKED ISSUE #1274: 🐛 [Bug] Wrong Date Format for English in Admin Interface
## 🐛 Bugreport
The dates are not formatted correctly when in English locale in the admin interface. Example:
```
The email was confirmed on Friday, 10.December.2021 16:05 clock.
```
should be
```
The email was confirmed on Friday, December 10th, 2021 4:05 pm
```

### LINKED ISSUE #1326: 🐛 [Bug] Wrong Date in Admin Interface for Email Last Send

Mogge:
## 🍰 Pullrequest
Does not display the information when the email was confirmed anymore as we do not have this data.
Displays the correct date and time in the case of unconfirmed emails when the last confirmation mail was sent.

### Issue
- fixes #1274
- fixes #1326

### Todo

- [X] None


ulfgebhardt (changes requested):
With production Data i get locally:

`The last email was sent to the member (some@email.tld) on .`

Using this query:
`SELECT email, users.id,user_id,verification_code,email_opt_in_type_id,login_email_opt_in.created,resend_count,updated FROM login_email_opt_in LEFT JOIN users on users.id = login_email_opt_in.user_id WHERE email = 'some@email.tld';`
those are the entries for `some@email.tld`:

image

The updatedAt field is not a very good source of imformation, since its written on import. Check if we can use created.

Mogge:
Good catch @ulfgebhardt 
Thanks for testing with live data. I take `createdAt` if `updatedAt` is not present

ulfgebhardt (approved):

PR #1474, opened 2022-02-13
Fix: Elopage Hook Crash

### LINKED ISSUE #1473: 🐛 [Bug] Crash bug on production due to problematic elopage call
## 🐛 Bugreport

Crash bug on production due to problematic elopage call.

The live server crashes on some elopage hooks due to `parseInt` not being an int, but `NaN`.
- [ ] Fix the problem
- [ ] Make issue to catch those errors

ulfgebhardt:
## 🍰 Pullrequest

Fixes a crash which occurs on production due to a field not reliable being an integer delivered by elopage. In order to fix this we check if the parsed fields are actually parsable as integer and make the fields in the database nullable. Exception is the price, here we write 0 if we cannot parse it.

### Issues

- fixes #1473

### Todo

- [ ] Make ticket for webhook failures not to crash server

### DIFF
--- a/backend/src/config/index.ts
+++ b/backend/src/config/index.ts
@@ -4,7 +4,7 @@ import dotenv from 'dotenv'
 dotenv.config()
 
 const constants = {
-  DB_VERSION: '0020-rename_and_clean_state_users',
+  DB_VERSION: '0021-elopagebuys_fields_nullable',
 }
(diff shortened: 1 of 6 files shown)


Mogge (approved):
I hope this fixes the problem

PR #1477, opened 2022-02-14
Fix: Do not log password or token to the console

### LINKED ISSUE #1475: 🐛 [Bug] User-Password are logged due to apollo logging
## 🐛 Bugreport

Currently we log all user password since we log the apollo queries. We need to mask the queries in oder to prevent password exposure to the log files

ulfgebhardt:
## 🍰 Pullrequest

Fix: Do not log password or token to the console

### Issues

- fixes #1475 
- relates https://github.com/shellscape/apollo-log/issues/17

### Todo

- [X] None

### DIFF
--- a/backend/src/server/plugins.ts
+++ b/backend/src/server/plugins.ts
@@ -1,7 +1,15 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 /* eslint-disable @typescript-eslint/explicit-module-boundary-types */
 
-import { ApolloLogPlugin } from 'apollo-log'
+import { ApolloLogPlugin, LogMutateData } from 'apollo-log'
+
+const copyInstance = (instance: any) => {
+  const data = Object.assign(
+    Object.create(Object.getPrototypeOf(instance)),
+    JSON.parse(JSON.stringify(instance)),
+  )
+  return data
+}
 
 const plugins = [
   {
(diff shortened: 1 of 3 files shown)


backend/src/server/plugins.ts
```diff
@@ -1,7 +1,15 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 /* eslint-disable @typescript-eslint/explicit-module-boundary-types */
 
-import { ApolloLogPlugin } from 'apollo-log'
+import { ApolloLogPlugin, LogMutateData } from 'apollo-log'
+
+const copyInstance = (instance: any) => {
+  const data = Object.assign(
```
Mogge:
```suggestion
  return Object.assign(
```

backend/src/server/plugins.ts
```diff
@@ -1,7 +1,15 @@
...
+    Object.create(Object.getPrototypeOf(instance)),
+    JSON.parse(JSON.stringify(instance)),
+  )
+  return data
```
Mogge:
```suggestion
```

Mogge (approved):
I prefer to use lodash.deepclone instead of writing code

ulfgebhardt:
> I prefer to use lodash.deepclone instead of writing code

I agree - but we don't have that package, doe we? Do we want to include it?

PR #1479, opened 2022-02-14
fix: Catch GDT Server Errors

### LINKED ISSUE #1469: 🔧 [Refactor] Decouple GDT and GDD
## 🔧 Refactor ticket

Currently we are relying on GDT to answer our requests - which will result in the failure of our service if the decay server doesnt work for whatever reason. This has been a problem in the past and with communities will be a problem in the future.

We have to find a way to decouple the GDT functionality (balance and transactions) from the rest of the software, if its via a seperate call or a timeout on server side are some of the possible choices we have here.

Mogge:
## 🍰 Pullrequest
Catch errors when GDT server is not available to keep the wallet working without GDT server

### Issues
- fixes #1469


backend/src/graphql/resolver/TransactionResolver.ts
```diff
@@ -331,11 +331,13 @@ export class TransactionResolver {
...
-    })
-    if (!resultGDTSum.success) throw new Error(resultGDTSum.data)
-    transactions.gdtSum = Number(resultGDTSum.data.sum) || 0
+    transactions.gdtSum = 0
+    try {
+      const resultGDTSum = await apiPost(`${CONFIG.GDT_API_URL}/GdtEntries/sumPerEmailApi`, {
+        email: userEntity.email,
+      })
+      if (resultGDTSum.success) transactions.gdtSum = Number(resultGDTSum.data.sum) || 0
+    } catch (err: any) {}
```
ulfgebhardt, line 340: since you do not throw here, we get a result from that call?

Can we also have a frontend implementation showing `--` or similar for GDT instead of 0?

backend/src/graphql/resolver/GdtResolver.ts
```diff
@@ -25,13 +25,17 @@ export class GdtResolver {
...
+    try {
+      const resultGDT = await apiGet(
+        `${CONFIG.GDT_API_URL}/GdtEntries/listPerEmailApi/${userEntity.email}/${currentPage}/${pageSize}/${order}`,
+      )
+      if (!resultGDT.success) {
+        throw new Error(resultGDT.data)
+      }
+      return new GdtEntryList(resultGDT.data)
+    } catch (err: any) {
+      throw new Error(err)
```
ulfgebhardt, line 37: So this will keep throwing?! 

I really would like a frontend implementation here telling the user something like `Due to an error we currently cannot show you your GDT` or alike in the GDT list to inform the users about whats happening.

ulfgebhardt (changes requested):

ulfgebhardt (approved):

PR #1530, opened 2022-02-23
fix: Possible SQL Exception in User Search

Mogge:
## 🍰 Pullrequest
If the filter of the user search returns no results, we get a SQL exception.

### Issues

- None

### Todo

- [X] None


ulfgebhardt (approved):
Same as the one with pending creations i guess ;)

PR #1534, opened 2022-02-24
fix: No Creations for Deleted Users

### LINKED ISSUE #1509: 🚀 [Feature] Disallow creation for deleted Accounts
## 🚀 Feature

While this might be already implemented in #1504 we should warn the admin when trying to create GDD for an disabled account.

Creating GDD for disabled accounts must be impossible

Scenario:
- Admin starts creation
- Other Admin disables account
- Admin confirms creation

Mogge:
## 🍰 Pullrequest
Checks that the user is not deleted when confirming a creation.

### Issues
- fixes #1509 

### Todo

- [X] None

### DIFF
--- a/backend/src/graphql/resolver/AdminResolver.ts
+++ b/backend/src/graphql/resolver/AdminResolver.ts
@@ -263,7 +263,7 @@ export class AdminResolver {
 
     const userIds = pendingCreations.map((p) => p.userId)
     const userCreations = await getUserCreations(userIds)
-    const users = await User.find({ id: In(userIds) })
+    const users = await User.find({ where: { id: In(userIds) }, withDeleted: true })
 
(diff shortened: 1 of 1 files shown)


backend/src/graphql/resolver/AdminResolver.ts
```diff
@@ -263,7 +263,7 @@ export class AdminResolver {
 
     const userIds = pendingCreations.map((p) => p.userId)
     const userCreations = await getUserCreations(userIds)
-    const users = await User.find({ id: In(userIds) })
+    const users = await User.find({ where: { id: In(userIds) }, withDeleted: true })
```
Mogge, line 266: When we not collect the deleted users as well, the UI shows empty name, email, etc.

ulfgebhardt (approved):

PR #1542, opened 2022-02-28
Fix: Upper case email on register breaks account

ulfgebhardt:
## 🍰 Pullrequest

This fixes a bug where users who registered with uppercase letters in their email not being able to login even after password reset.

This affected 20 users on production, of which 7 were test accounts. Of the 13 users one was from early 2020, while the remaining 12 were the earliest from 16.10.2021. This means this bug was introduced with the switch to Apollo.

- Corrects existing users in the database
- Fixes the register call to properly sanitize user input

Sideeffect:
- Since the email is part of our password generation mechanism this PR will invalidate the passwords of affected users (20). Those have to reset their passwords to regain access to their account.

### Repro
1. Check out `master`
2. Register with `sPoNgEbOb@wHaTeVeR.oCeAn`
3. Try login  -> should fail, even when typing the email in the same upper cases as in the register form
4. Try reset password -> no effect on login success
5. Checkout this branch
6. Try login -> should not work 
7. Try reset password -> things should get rolling now
8. Register another account with UpperCase letters in the email
9. login should work out of the box now

### Todo

- [] Please test and verify.

### DIFF
--- a/backend/src/config/index.ts
+++ b/backend/src/config/index.ts
@@ -4,7 +4,7 @@ import dotenv from 'dotenv'
 dotenv.config()
 
 const constants = {
-  DB_VERSION: '0024-combine_transaction_tables',
+  DB_VERSION: '0025-emails_to_lower',
   DECAY_START_TIME: new Date('2021-05-13 17:46:31'), // GMT+0
 }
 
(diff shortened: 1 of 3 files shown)


Elweyn (approved):
Test was succesful (With @ulfgebhardt, @ogerly and @Elweyn)

PR #1576, opened 2022-03-08
feat: Use Vue Filter to Display Gradido Amounts

### LINKED ISSUE #1575: 🚀 [Feature] Proper Minus Sign for Negative Amounts
## 🚀 Feature
¡Render a proper minus sign for negative amounts, avoiding multiplication by -1 and adding the minus sign by if conditions.

Mogge:
## 🍰 Pullrequest
Makes globally two filter available:
* `GDD`: renders the localized amount with leading plus or minus sign, adding GDD at end
* `amount`: renders the localized amount, when negative, with a leading minus sign

### Issues
- fixes #1575

### Todo

- [X] None


frontend/src/locales/de.json
```diff
@@ -37,9 +37,9 @@
     "months": "Monate",
     "noDecay": "Keine Vergänglichkeit",
     "past_time": "Vergangene Zeit",
-    "received": "Empfangen",
+    "receive": "Empfangen",
```
ulfgebhardt, line 40: ??? why not past tense?

frontend/src/locales/de.json
```diff
@@ -37,9 +37,9 @@
...
     "seconds": "Sekunden",
-    "sent": "Gesendet",
+    "send": "Gesendet",
```
ulfgebhardt, line 42: ??? why not past tense?

frontend/src/filters/amount.js
```diff
@@ -0,0 +1,18 @@
...
+  if (!value && value !== 0) return ''
+  return i18n.n(value.toString(), 'decimal').replace('-', '− ')
+}
+
+const GDD = (value) => {
+  value = amount(value)
+  if (value === '') return ''
+  if (!value.match(/^− /)) value = '+ ' + value
+  return value + ' GDD'
+}
```
ulfgebhardt, line 18: why two functions. You seem to never call amount outside of the GDD function. Combining both could also save you the regex match.

frontend/src/locales/de.json
Mogge: `<b-col cols="6" class="text-right">{{ $t(`decay.${typeId.toLowerCase()}`) }}</b-col>`

frontend/src/locales/de.json
Mogge: see above

frontend/src/filters/amount.js
Mogge: `amount` is used. I think of implement a `GDT` filter as well, so it is good to have a helper function.

frontend/src/main.js
```diff
@@ -20,6 +21,9 @@ Vue.use(DashboardPlugin)
 Vue.config.productionTip = false
 
 Vue.mixin(toasters)
+const filters = loadFilters(i18n)
+Vue.filter('amount', filters.amount)
```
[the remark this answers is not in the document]
Mogge: ?

ogerly (approved):

PR #1580, opened 2022-03-08
[WIP] Blockchain connector

einhornimmond:
## 🍰 Pullrequest
Use Blockchain Connector to transmit transaction via Iota to Gradido Nodes with updated protocol
- Register Group/Community on Blockchain if not already exist


backend/src/config/index.ts
```diff
@@ -54,6 +56,15 @@ const loginServer = {
   LOGIN_SERVER_KEY: process.env.LOGIN_SERVER_KEY || 'a51ef8ac7ef1abf162fb7a65261acd7a',
 }
 
+const blockchainConnector = {
+  BLOCKCHAIN_CONNECTOR_API_URL:
+    process.env.BLOCKCHAIN_CONNECTOR_API_URL || 'http://blockchain-connector:1271',
```
Elweyn, line 61: Not sure what this means?
Do we have on local a blockchain-connector:1271 or is this a URL that was registered and port 1271 is open for the tests?

backend/src/blockchain/Crypto.ts
```diff
@@ -0,0 +1,72 @@
...
+  const privKey = Buffer.alloc(sodium.crypto_sign_SECRETKEYBYTES)
+
+  sodium.crypto_sign_seed_keypair(
+    pubKey,
+    privKey,
+    outputHashBuffer.slice(0, sodium.crypto_sign_SEEDBYTES),
+  )
+
+  return [pubKey, privKey]
+}
```
Elweyn, line 40: Think we already have this function in UserResolver.ts in the backend, could you please unify them!
This location is good imo but please remove it from the UserResolver so it is lighter

backend/src/blockchain/GradidoNodeRequests.ts
```diff
@@ -0,0 +1,43 @@
+import CONFIG from '../config'
+import { RpcClient } from 'jsonrpc-ts'
+import { randomInt } from 'crypto'
+
+interface GroupDetails {
+  GroupName: string
+  GroupAlias: string
+  CoinColor: number | string
```
Elweyn, line 8: Hu? What is this for?

backend/src/blockchain/TransactionToIota.ts
```diff
@@ -0,0 +1,161 @@
...
+    apolloTransactionId: apolloTransactionId,
+    memo: memo,
+    recipientPubkey: recipientUser.pubKey.toString('hex'),
+    amount: amount,
+    targetDate: targetDate.toISOString(),
+  }
+  return signAndSendTransaction(signingUser, privateKey, packTransactionRequest, encryptedMemo)
+}
+
+export { sendCoins, registerNewGroup, creation }
```
Elweyn, line 161: Do I understand it right:
This are the connector methods that need to be added to the TransactionResolver for sendCoins and AdminResolver for creation?
A bit like registerNewGroup thas has been added to ctrateServer!?

backend/src/blockchain/TransactionToIota.ts
```diff
@@ -0,0 +1,161 @@
...
+  created: Date,
+  signingUser: User,
+  recipientUser: User,
+  apolloTransactionId: number,
+  memo: string,
+  amount: string,
+  targetDate: Date,
+): Promise<Buffer> {
+  const privateKey = await recoverPrivateKey(signingUser)
+  const encryptedMemo = encryptMemo(memo, privateKey, recipientUser.pubKey)
```
Elweyn, line 148: This could be done in signAndSendTransaction so you would only prepare the changing datas in creation and do the logic in signAndSendTransaction so you don't need to jump so much between code to have an overview.
```suggestion
```

backend/src/blockchain/TransactionToIota.ts
```diff
@@ -0,0 +1,161 @@
...
+  const packTransactionRequest = {
+    transactionType: 'creation',
+    created: created.toISOString(),
+    apolloTransactionId: apolloTransactionId,
+    memo: memo,
+    recipientPubkey: recipientUser.pubKey.toString('hex'),
+    amount: amount,
+    targetDate: targetDate.toISOString(),
+  }
+  return signAndSendTransaction(signingUser, privateKey, packTransactionRequest, encryptedMemo)
```
Elweyn, line 158:
```suggestion
  return signAndSendTransaction(signingUser, packTransactionRequest, memo)
```

backend/src/blockchain/TransactionToIota.ts
```diff
@@ -0,0 +1,161 @@
...
+): Promise<Buffer> {
+  const privateKey = await recoverPrivateKey(user)
+  const packTransactionRequest = {
+    transactionType: 'groupAdd',
+    created: created.toISOString(),
+    groupName: communityName,
+    groupAlias: communityAlias,
+    coinColor: communityCoinColor,
+  }
+  return signAndSendTransaction(user, privateKey, packTransactionRequest, '')
```
Elweyn, line 135:
```suggestion
  return signAndSendTransaction(user, packTransactionRequest, '')
```

backend/src/blockchain/TransactionToIota.ts
```diff
@@ -0,0 +1,161 @@
...
+    apolloTransactionId: apolloTransactionId,
+    senderPubkey: user.pubKey.toString('hex'),
+    recipientPubkey: recipientPublicHex,
+    amount: amount,
+    memo: encryptedMemo,
+    senderGroupAlias: '',
+    recipientGroupAlias: '',
+  }
+
+  return signAndSendTransaction(user, privateKey, packTransactionRequest, recipientGroupAlias)
```
Elweyn, line 117:
```suggestion
  return signAndSendTransaction(user, packTransactionRequest, recipientGroupAlias)
```

backend/src/blockchain/TransactionToIota.ts
```diff
@@ -0,0 +1,161 @@
...
+  created: Date,
+  user: User,
+  apolloTransactionId: number,
+  recipientPublicHex: string,
+  amount: string,
+  memo: string,
+  recipientGroupAlias: string | '',
+): Promise<Buffer> {
+  const privateKey = await recoverPrivateKey(user)
+  const encryptedMemo = encryptMemo(memo, privateKey, Buffer.from(recipientPublicHex, 'hex'))
```
Elweyn, line 104: This is not used if I see it right
```suggestion
```

backend/src/blockchain/TransactionToIota.ts
```diff
@@ -0,0 +1,161 @@
...
+  user: User,
+  privateKey: Buffer,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  packTransaction: any,
+  recipientGroupAlias: string,
+): Promise<Buffer> {
+  if (recipientGroupAlias !== '') {
+    packTransaction.senderGroupAlias = CONFIG.COMMUNITY_ALIAS
+    packTransaction.recipientGroupAlias = recipientGroupAlias
+  }
```
Elweyn, line 32:
```suggestion
  memo: string,
): Promise<Buffer> {
  const privateKey = await recoverPrivateKey(user)
  if (memo !== '') {
    packTransaction.senderGroupAlias = CONFIG.COMMUNITY_ALIAS
    const recipientGroupAlias = encryptMemo(memo, privateKey, Buffer.from(recipientPublicHex, 'hex'))
    packTransaction.recipientGroupAlias = recipientGroupAlias
  }
```

backend/src/blockchain/TransactionToIota.ts
```diff
@@ -0,0 +1,161 @@
...
+}
+
+async function registerNewGroup(
+  created: Date,
+  user: User,
+  communityName: string,
+  communityAlias: string,
+  communityCoinColor: number | string,
+): Promise<Buffer> {
+  const privateKey = await recoverPrivateKey(user)
```
Elweyn, line 127:
```suggestion
```

backend/src/server/createServer.ts
```diff
@@ -41,7 +47,37 @@ const createServer = async (context: any = serverContext): Promise<ServerDef> =>
...
+    // load first admin user for signing transaction
+    const created = new Date()
+    const adminUserRepository = getRepository(UserAdmin)
+    const adminUserResult = await adminUserRepository.find({
+      select: ['userId'],
+      order: {
+        userId: 'ASC',
+      },
+      take: 1,
+    })
```
Elweyn, line 63: You need an admin user to registrate a new group? But is it really ok to take a random / first found one?

Elweyn (changes requested):

backend/src/config/index.ts
einhornimmond: It will be local, but not yet. I haven't written the Docker and Docker-Compose things yet for that.

backend/src/blockchain/Crypto.ts
einhornimmond: ok

backend/src/blockchain/GradidoNodeRequests.ts
einhornimmond: On Blockchain-Side I have already implemented the coinColor concept Bernd talked about. So every transaction has also a coin color.
So user can decide by themself which coins they would like to accept. 
For that to work I need on the blockchain the info which group has which native coin color.
The Gradido-Node has a json-rpc function to get the GroupDetails for a group alias. For jsonrpc-ts I must define the returned data type. 
In my test Setup the Apollo Server uses this to check if his group was already registered and if not create a transaction for that. It isn't ideal, but enough for tests. Later it can be implemented in admin so that the Server-Admin can manually send this register group transaction as part of the server setup.

backend/src/blockchain/TransactionToIota.ts
einhornimmond: Yes

backend/src/server/createServer.ts
einhornimmond: No, but it will much more work for me to implement a proper process in admin for that. 
Also I don't know if you have already planned a kind of setup process on first start in admin. 
Currently creating an admin user isn't even possible from the UI. It is only possible with direct db access.
So my focus is to test the blockchain connector and gradido node together with apollo. We can later integrate it in an server-setup process when it exists.

backend/src/blockchain/TransactionToIota.ts
einhornimmond: Ich glaube du hast da etwas verwechselt. Du hast recipientGroupAlias und memo zusammengewürfelt. 
Die gehören nicht zusammen.

PR #1585, opened 2022-03-09
feat: Create Transaction Link Mutation

Mogge:
## 🍰 Pullrequest
Mutation to create transaction links

### Issues
- relates #1551

### Todo

- [X] Handle decay
- [X] Test available balance for existing transaction links

PR #1586, opened 2022-03-09
Query-transaction-link

### LINKED ISSUE #1557: 🚀 [Feature] Backend - Check redeem link
## 🚀 Feature

Implement a call in the backend which checks a redeem link and produces the information behind it.
This call needs to be publicly available

- requires #1551 
- requires #1552

Elweyn:
## 🍰 Pullrequest
The frontend needs a query where it can get the TransactionLink information for a given code.

### Issues
- fixes #1557 
- require #1585 

### Todo
- [X] Query transactionLink for a given code 
- [X] Get User information from the transactionLink.userId
- [X] Return TransactionLink with every information


ogerly (approved):

backend/src/auth/RIGHTS.ts
```diff
@@ -18,6 +18,8 @@ export enum RIGHTS {
   SET_PASSWORD = 'SET_PASSWORD',
   UPDATE_USER_INFOS = 'UPDATE_USER_INFOS',
   HAS_ELOPAGE = 'HAS_ELOPAGE',
+  CREATE_TRANSACTION_LINK = 'CREATE_TRANSACTION_LINK',
+  QUERY_TRANSACTION_LINK = 'QUERY_TRANSACTION_LINK',
```
ulfgebhardt, line 22: Do you assign this properly in inalienable rights?

Mogge: This must be set as an inalienable right

Mogge (changes requested):

backend/src/graphql/resolver/TransactionLinkResolver.ts
```diff
@@ -0,0 +1,74 @@
...
+    transactionLink.memo = memo
+    transactionLink.code = transactionLinkCode(createdDate)
+    transactionLink.createdAt = createdDate
+    transactionLink.validUntil = transactionLinkExpireDate(createdDate)
+    transactionLink.showEmail = showEmail
+    await dbTransactionLink.save(transactionLink).catch((error) => {
+      throw error
+    })
+
+    return new TransactionLink(transactionLink, new User(user))
```
ulfgebhardt, line 63: redeemed by user

backend/src/graphql/resolver/TransactionLinkResolver.ts
```diff
@@ -12,7 +12,7 @@ import { RIGHTS } from '@/auth/RIGHTS'
 import { randomBytes } from 'crypto'
 import { User } from '@model/User'
 import { calculateDecay } from '@/util/decay'
-
+^
```
Mogge, line 15: ??

backend/src/graphql/resolver/TransactionLinkResolver.ts
```diff
@@ -67,4 +67,13 @@ export class TransactionLinkResolver {
 
     return new TransactionLink(transactionLink, new User(user))
   }
+
+  @Authorized([RIGHTS.QUERY_TRANSACTION_LINK])
+  @Query(() => TransactionLink)
+  async queryTransactionLink(@Arg('code') code: string): Promise<TransactionLink> {
+    const transactionLink = await dbTransactionLink.findOneOrFail({ code })
+    const userRepository = getCustomRepository(UserRepository)
+    const user = await userRepository.findOneOrFail({ id: transactionLink.userId })
```
Mogge, line 76: This works when `transactionLink.userId === null`?

backend/src/graphql/resolver/TransactionLinkResolver.ts
```diff
@@ -67,4 +67,13 @@ export class TransactionLinkResolver {
...
+    return new TransactionLink(transactionLink, new User(user))
```
Mogge, line 77: What happens when `user = null`?

Mogge (approved):

backend/src/graphql/resolver/TransactionLinkResolver.ts
Elweyn: Can the userId of the transactionLink be null? And even when it will send an exception cause it does not find any user with id: null

backend/src/graphql/resolver/TransactionLinkResolver.ts
Elweyn: Not here! There can not be any redeemed user yet since it is the mutation that creates the TransactionLink

backend/src/graphql/resolver/TransactionLinkResolver.ts
Elweyn: Fail so it will throw an exception on line 76

PR #1616, opened 2022-03-14
refactor: No Reset DB in Backend Unit Tests

Mogge:
## 🍰 Pullrequest
Avoid using reset database from the database folder. Use a helper instead, that removes all data from the tables.

### Issues

- None

### Todo

- [X] None

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -489,7 +489,7 @@ jobs:
   unit_test_backend:
     name: Unit tests - Backend
     runs-on: ubuntu-latest
-    needs: [build_test_mariadb]
+    needs: [build_test_mariadb, build_test_database_up]
     steps:
       ##########################################################################
       # CHECKOUT CODE ##########################################################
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -510,15 +510,7 @@ jobs:
...
-        run: docker-compose -f docker-compose.yml -f docker-compose.test.yml up --detach --no-deps mariadb
-      - name: Sleep for 30 seconds
-        run: sleep 30s
-        shell: bash
-      - name: backend | docker-compose database
-        run: docker-compose -f docker-compose.yml -f docker-compose.test.yml up --detach --no-deps database
-      - name: Sleep for 30 seconds
-        run: sleep 30s
-        shell: bash
+        run: docker-compose -f docker-compose.yml -f docker-compose.test.yml up --detach --no-deps mariadb database
```
ulfgebhardt, line 513: we start mariadb appart from database, because the first startup of mariadb is very slow. Does this work reliable?

.github/workflows/test.yml
```diff
@@ -489,7 +489,7 @@ jobs:
   unit_test_backend:
     name: Unit tests - Backend
     runs-on: ubuntu-latest
-    needs: [build_test_mariadb]
+    needs: [build_test_mariadb, database_migration_test, build_test_database_up]
```
ulfgebhardt, line 492: Why?
```suggestion
    needs: [build_test_mariadb]
```

ulfgebhardt (approved):

PR #1623, opened 2022-03-17
fix: Query for Only Creations Transaction List

Mogge:
## 🍰 Pullrequest
As we query the DB when we are authenticated for the user, the user is added to the context. In resolvers that require the user it is taken from the context and not queried again anymore.

The transaction list query does not take a user id and a only creations argument anymore.

New query with admin rights to get the list of creation transactions to a given user id.

### Issues

I found a to do in the code but no issue.

- None

### Todo

- [X] None


backend/src/graphql/resolver/AdminResolver.ts
```diff
@@ -339,6 +342,27 @@ export class AdminResolver {
 
     return true
   }
+
+  @Authorized([RIGHTS.CREATION_TRANSACTION_LIST])
+  @Query(() => [Transaction])
+  async creationTransactionList(
+    @Args()
+    { currentPage = 1, pageSize = 25, order = Order.DESC }: Paginated,
+    @Arg('userId', () => Int) userId: number,
```
Mogge, line 351: This is how we get rid of the float IDs

backend/src/graphql/directive/isAuthorized.ts
```diff
@@ -35,6 +35,7 @@ const isAuthorized: AuthChecker<any> = async ({ context }, rights) => {
   const userRepository = await getCustomRepository(UserRepository)
   try {
     const user = await userRepository.findByPubkeyHex(context.pubKey)
+    context.user = user
```
ulfgebhardt, line 38: I like to discuss if we want `context.user` to be private and access it with `context.getUser()`.

The Idea behind is that we cannot be sure the user object is set, since it depends on what right is queried in the resolver function header.

Consider the following example:

```
  @Authorized([RIGHTS.SOME_PUBLIC_RIGHT])
  @Mutation(() => Boolean)
  async doSomething(@Arg('peter') peter: number, @Ctx() context: any): Promise<boolean> {
    [...]
    const moderatorUser = context.user
    [...]
```

We would not get an error in this case, since `moderatorUser` will just be `undefined` but will not necessarily break the program.

If we keep it like you implemented it its simple and elegant, but programmers must make sure themselves, that the right queried for also writes an user object to the context.

Alternatively we could define a function in the context(or outside for that matter):

```
context.getUser = () => {
   if(!this.user){
      throw new Error('No user found. Not authorized')
   }
  return this.user
}
```

ulfgebhardt (approved):

PR #1624, opened 2022-03-17
refactor: No Float Ids

Mogge:
## 🍰 Pullrequest
Proper type definitions for id parameters in the graphql schema (Int instead of Float).
Also fixes a wrong toast message when user is recovered. 

### Issues

- None

### Todo

- [X] None


ulfgebhardt (approved):

Elweyn (approved):
UUID would really great

Issue #1653, opened 2022-03-21
🚀 [Feature] Rounding errors in the frontend when calculating the decay


ulfgebhardt:
## 🚀 Feature

There are rounding errors in the frontend when calculating the decay resulting in wrong calculations to be displayed.

Issue #1608, opened 2022-03-11
🐛 [Bug] Decay in Frontend is Error Prone


Mogge:
## 🐛 Bugreport

![image](https://user-images.githubusercontent.com/15882241/157879872-c17824df-b8a4-4b69-9645-d0dafec8adf1.png)

relates #1596

PR #1665, opened 2022-03-23
refactor: Balance Resolver

### LINKED ISSUE #1609: 🔧 [Refactor] Investigate Balance Model
## 🔧 Refactor ticket

Investigate Balance Model - reuse it or throw it?

Mogge:
## 🍰 Pullrequest
Refactor balance resolver and model to make use of it.

### Issues
- fixes #1609

### Todo

- [X] None


backend/src/graphql/resolver/BalanceResolver.ts
```diff
@@ -5,18 +5,24 @@ import { Resolver, Query, Ctx, Authorized } from 'type-graphql'
...
 export class BalanceResolver {
   @Authorized([RIGHTS.BALANCE])
   @Query(() => Balance)
   async balance(@Ctx() context: any): Promise<Balance> {
-    // load user and balance
     const { user } = context
     const now = new Date()
 
+    const gdtResolver = new GdtResolver()
+    const balanceGDT = await gdtResolver.gdtSum(context)
```
ulfgebhardt, line 24: this implies to have a successful call on `balance` you need the right `BALANCE` and `GDT_SUM` ?!

backend/src/graphql/resolver/GdtResolver.ts
```diff
@@ -35,6 +33,25 @@ export class GdtResolver {
     }
   }
 
+  @Authorized([RIGHTS.GDT_SUM])
+  @Query(() => Number)
+  async gdtSum(@Ctx() context: any): Promise<number | null> {
```
ulfgebhardt, line 38: I vote for gdtBalance in this case - better name, more descriptive and aligned with `balance`

ulfgebhardt: Also: Why expose this to the GraphQL interface? Do we need that? Can't it just be part of `balance` without external exposure?

backend/src/graphql/resolver/TransactionResolver.ts
```diff
@@ -143,19 +143,8 @@ export class TransactionResolver {
...
-      })
-      if (!resultGDTSum.success) {
-        throw new Error('Call not successful')
-      }
-      balanceGDT = Number(resultGDTSum.data.sum) || 0
-    } catch (err: any) {
-      // eslint-disable-next-line no-console
-      console.log('Could not query GDT Server', err)
-    }
+    const gdtResolver = new GdtResolver()
```
ulfgebhardt, line 146: include a full balance object in this call and not only the gdt balance

backend/src/graphql/resolver/BalanceResolver.ts
Mogge: No. gdtBlanace returns null when API call is not seccessful

backend/src/graphql/resolver/GdtResolver.ts
Mogge: We expose the gst transaction list but not the gdt balance? Does not make sense for me. May be we have an use case in the future where we want only query the gdt balance. Here it is

ulfgebhardt:
What about #1653 , #1608 ? Does this fix this aswell?

backend/src/graphql/resolver/BalanceResolver.ts
```diff
@@ -5,36 +5,74 @@ import { Resolver, Query, Ctx, Authorized } from 'type-graphql'
...
-        decay_date: now.toString(),
+        lastBookedBalance: new Decimal(0),
+        balanceGDT,
+        count: 0,
+        linkCount: 0,
       })
     }
 
+    const count =
+      context.count || context.count === 0
```
ulfgebhardt, line 43: what is `context.count`? If we save stuff in the context at least have the courtesy to give it a name able to identify what it is (e.g. `transactionCount`)

backend/src/graphql/resolver/BalanceResolver.ts
```diff
@@ -5,36 +5,74 @@ import { Resolver, Query, Ctx, Authorized } from 'type-graphql'
...
+              userId: user.id,
+              redeemedAt: null,
+              validUntil: MoreThan(new Date()),
+            },
+          })
+
+    const transactionLinkRepository = getCustomRepository(TransactionLinkRepository)
+    const { sumHoldAvailableAmount } = context.sumHoldAvailableAmount
+      ? { sumHoldAvailableAmount: context.sumHoldAvailableAmount }
+      : await transactionLinkRepository.summary(user.id, now)
```
ulfgebhardt, line 60: if you want to optimize queries, this query could return the `linkCount` aswell and the caching mechanism becomes obsolete.

backend/src/graphql/resolver/BalanceResolver.ts
```diff
@@ -5,36 +5,74 @@ import { Resolver, Query, Ctx, Authorized } from 'type-graphql'
...
+      lastTransaction.balanceDate,
+      now,
+    )
+
     return new Balance({
-      balance: lastTransaction.balance,
-      decay: calculateDecay(lastTransaction.balance, lastTransaction.balanceDate, now).balance,
-      decay_date: now.toString(),
+      balance: calculatedDecay.balance.toDecimalPlaces(2, Decimal.ROUND_DOWN), // round towards zero
+      decay: calculatedDecay.decay.toDecimalPlaces(2, Decimal.ROUND_FLOOR), // round towards - infinity
```
ulfgebhardt, line 70: Why round towards -infinity and not +infinity? Do we not to round up the decay value?
image

ulfgebhardt (commented):
I am against this "we cache everything in the context and blow stuff in with weird names and hope we find it later again."

If we want to optimize queries we should use our ORM which can do this by default: https://github.com/typeorm/typeorm/blob/master/docs/caching.md

backend/src/graphql/resolver/BalanceResolver.ts
Mogge: OK

backend/src/graphql/resolver/BalanceResolver.ts
Mogge: I use the context to pass the information that is already queried by the transaction resolver to avoid further queries here. This is not optimization as I already know the values.

backend/src/graphql/resolver/BalanceResolver.ts
Mogge: Decay is always negative. E.g. `- 1.2398`. Round towards `+ infinity` gives `-1.23`, round towards `- infinity` becomes `-1.24`, this is what I want.

ulfgebhardt (approved):

PR #1692, opened 2022-03-30
fix: No Decay Calculation in Frontend

Mogge:
## 🍰 Pullrequest
A method provides the balance of the previous transaction to all transaction types. This way the decay transaction does not have to calculate the previous balance. This way we can transfer all the rounding problems on decay calculations to the backend.
This allows for the other components to show the whole calculation of the balance change, not only the decay.

### Issues
- relates #1653 

### Todo

- [X] None


frontend/src/components/GddTransactionList.vue
```diff
@@ -110,6 +117,10 @@ export default {
       })
       window.scrollTo(0, 0)
     },
+    previousBookedBalance(idx) {
```
ulfgebhardt, line 120: i consider this a bit hacky - tho its fine for now. In my opinion we should have a reference to the previous transaction instead of only the balance value if we do stuff like this.
Furthermore this brings more logic into the frontend - but I guess thats inevitable for this purpose if we do not want to transfer data twice

ulfgebhardt (approved):

PR #1700, opened 2022-03-31
Fix: Allow sending of more then half of my wealth via link

ulfgebhardt:
## 🍰 Pullrequest

This fix makes sure we can send all our GDD via link.

### Issues

- None

### Todo

- [X] None

### DIFF
--- a/backend/src/util/validate.ts
+++ b/backend/src/util/validate.ts
@@ -36,7 +36,7 @@ async function calculateBalance(
 
   // If we want to redeem a link we need to make sure that the link amount is not considered as blocked
   // else we cannot redeem links which are more or equal to half of what an account actually owns
-  const releasedLinkAmount = transactionLink ? transactionLink.amount : new Decimal(0)
+  const releasedLinkAmount = transactionLink ? transactionLink.holdAvailableAmount : new Decimal(0)
 
   if (
     balance.minus(sumHoldAvailableAmount.toString()).plus(releasedLinkAmount.toString()).lessThan(0)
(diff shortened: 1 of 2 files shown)


backend/src/util/validate.ts
```diff
@@ -32,7 +34,13 @@ async function calculateBalance(
...
   const { sumHoldAvailableAmount } = await transactionLinkRepository.summary(userId, time)
 
-  if (balance.minus(sumHoldAvailableAmount.toString()).lessThan(0)) {
+  // If we want to redeem a link we need to make sure that the link amount is not considered as blocked
+  // else we cannot redeem links which are more or equal to half of what an account actually owns
+  const sumHoldAvailableAmountMinusTransactionLink = transactionLink
+    ? sumHoldAvailableAmount.minus(transactionLink.amount.toString())
+    : sumHoldAvailableAmount
+
+  if (balance.minus(sumHoldAvailableAmountMinusTransactionLink.toString()).lessThan(0)) {
```
Mogge, line 43:
```suggestion
  const releasedLinkAmount = transactionLink ? transactionLink.amount : new Decimal(0)

  if (balance.minus(sumHoldAvailableAmount.toString()).plus(releasedLinkAmount.toString()).lessThan(0)) {
```

Mogge: Is it better to release `holdAvailableAmount` or even to calculate the real decay to now?

ulfgebhardt: i'd say its better to release the full amount for the link

Mogge (approved):

PR #1705, opened 2022-03-31
1703 submit button disabled when total amount to submit is minus

### LINKED ISSUE #1703: 🐛 [Bug] submit button disabled when total amount to submit is minus
## 🐛 Bugreport
submit button disabled when total amount to submit is minus

If I send the entire sum of my GDD then I must 
1. the resulting decay must be taken into account
2. a transaction should not be sendable if the send button on the transaction confirm is disabled when checking or offsetting and I have a minus amount in the total.

ogerly:
## 🍰 Pullrequest
submit button disabled when total amount to submit is minus

### Issues
- fixes #1703


Mogge (commented):
I do not understand why the mechanism using property `loading` here does prevent a double click, whereas it does not in the confirmation send component

ulfgebhardt (approved):

Issue #1485, opened 2022-02-15
🔧 [Refactor] Bake Community name and Description into frontend


ulfgebhardt:
## 🔧 Refactor ticket

We currently query the backend for community name and description and will never update those once we queried them and the store is available. This is error prone and requires initial network traffic.

We now have a centralized config for our deployment and therefor can bake in the community name and description in the build process of the frontend.

PR #1749, opened 2022-04-06
[WIP] Refactor: Frontend decay start block as static config value

ulfgebhardt:
### Issues

- fixes partially #1485 

### Todo

- [X] None

### DIFF
--- a/backend/src/graphql/model/Balance.ts
+++ b/backend/src/graphql/model/Balance.ts
@@ -1,6 +1,5 @@
 import { ObjectType, Field } from 'type-graphql'
 import Decimal from 'decimal.js-light'
-import CONFIG from '@/config'
 
 @ObjectType()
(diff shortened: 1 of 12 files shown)


Elweyn (commented):
I see the changes and they withdraw the decayStartBlock injection hell!
Now we need to configure it in frontend as well as in backend! Cause of the decay calculation!
I have some questions: 
In transaction we have a field decay_start, it has nothing to do with the decayStartBlock or is it linked?
And last question in backend/src/seeds/graphql/queries.ts transactionsQuery their is still a decayStartBlock occurence what does it do? Should it not also be withdrawen?

Mogge (dismissed):

ulfgebhardt:
Good catch @Elweyn

Mogge (approved):

PR #1762, opened 2022-04-11
refactor: Define Context Interface

Mogge:
## 🍰 Pullrequest
Define the context interface used in the resolvers.

### Issues

- None

### Todo

- [X] None

### DIFF
--- a/backend/src/graphql/resolver/AdminResolver.ts
+++ b/backend/src/graphql/resolver/AdminResolver.ts
@@ -1,6 +1,4 @@
-/* eslint-disable @typescript-eslint/no-explicit-any */
-/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
-
+import { Context } from '@/server/context'
 import { Resolver, Query, Arg, Args, Authorized, Mutation, Ctx, Int } from 'type-graphql'
 import {
   getCustomRepository,
@@ -137,7 +135,7 @@ export class AdminResolver {
   @Mutation(() => Date, { nullable: true })
   async deleteUser(
     @Arg('userId', () => Int) userId: number,
-    @Ctx() context: any,
+    @Ctx() context: Context,
   ): Promise<Date | null> {
     const user = await dbUser.findOne({ id: userId })
     // user exists ?
@@ -146,7 +144,7 @@ export class AdminResolver {
     }
     // moderator user disabled own account?
     const moderatorUser = context.user
-    if (moderatorUser.id === userId) {
+    if (moderatorUser && moderatorUser.id === userId) {
       throw new Error('Moderator can not delete his own account!')
     }
     // soft-delete user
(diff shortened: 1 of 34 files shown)


backend/src/server/context.ts
```diff
@@ -14,4 +29,9 @@ const context = (args: any) => {
   return context
 }
 
+export const getUser = (context: Context): dbUser => {
+  if (context.user) return context.user
+  throw new Error('No user given in context!')
+}
```
ulfgebhardt, line 35: cant we make this function a part of context?

`const user = context.getUser()`

This will automatically throw if there is no user and make those checks unnecessary `if (moderatorUser && moderatorUser.id....`

Furthermore  this function could be extended that it can actually query the user from the database when needed and serve as cache:

```
getUser(){
  if user return user
  if jwt query and return user
  throw no user
}
```

backend/src/graphql/resolver/AdminResolver.ts
```diff
@@ -146,7 +144,7 @@ export class AdminResolver {
     }
     // moderator user disabled own account?
     const moderatorUser = context.user
```
ulfgebhardt, line 146: cant we use `getUser` here aswell? We should not be able to not have a user on this call anyways - so the `getUser()` function should never throw.

backend/src/server/context.ts
ulfgebhardt: An advance validation mechanic would be:

```
@MiddelWare(requiresuserMiddelware)
myQuery(args, UserProvidedByMidelWare: dbUser){
use user....
}
```
But I guess this would be an bigger exploration to find out if this works and is useful.
The advantage of this mechanic is, that a call would need to state if it needs the user and if so it would fail when its not presented. Furthermore the code inside the call would not to handle error cases etc since thats handled in the middleware already. The compiler would be tasked with throwing errors when the user is required in the call's code, but not acquired via middleware.

As I said - a bit far fetched - I like your solution - if you don't mind check my remarks and we are good to go on this.

backend/yarn.lock
```diff
@@ -1704,9 +1704,9 @@ camelcase@^6.2.0:
   integrity sha512-c7wVvbw3f37nuobQNtgsgG9POC9qMbNuMQmTCqZv23b6MIz0fcYpBiOlv9gEN/hdLdnZTDQhg6e9Dq5M1vKvfg==
 
 caniuse-lite@^1.0.30001264:
-  version "1.0.30001265"
-  resolved "https://registry.yarnpkg.com/caniuse-lite/-/caniuse-lite-1.0.30001265.tgz#0613c9e6c922e422792e6fcefdf9a3afeee4f8c3"
-  integrity sha512-YzBnspggWV5hep1m9Z6sZVLOt7vrju8xWooFAgN6BA5qvy98qPAPb7vNUzypFaoh2pb3vlfzbDO8tB57UPGbtw==
+  version "1.0.30001325"
+  resolved "https://registry.npmjs.org/caniuse-lite/-/caniuse-lite-1.0.30001325.tgz"
+  integrity sha512-sB1bZHjseSjDtijV1Hb7PB2Zd58Kyx+n/9EotvZ4Qcz2K3d0lWB8dB4nb8wN/TsOGFq3UuAm0zQZNQ4SoR7TrQ==
```
ulfgebhardt, line 1709: what happend here?

ulfgebhardt (approved):

PR #1797, opened 2022-04-19
Concept to Introduce Gradido ID

clauspeterhuebner:
Please revie the concept UC_Introduction_of_Gradido-ID.md


docu/Concepts/TechnicalRequirements/UC_Introduction_of_Gradido-ID.md
```diff
@@ -0,0 +1,141 @@
...
+### Database-Schema
+
+#### Users-Table
+
+The entity users has to be changed by adding the following columns.
+
+| Column                   | Type   | Description                                                                            |
+| ------------------------ | ------ | -------------------------------------------------------------------------------------- |
+| gradidoID                | String | technical unique key of the user as UUID (version 4)                                   |
+| alias                    | String | a business unique key of the user                                                      |
```
ulfgebhardt, line 28: this is not applicable, because the user will have multiple aliases:
user <---1:n--> alias

a user might be reachable as `1234-5678-90` (UUID), `peter.lustig#1234` (his actual username containing some random value), `peter.lutig#1234` (his username with typo) and `slayer69#1234` (his former username).

It is to define how the alias is build and what are valid values, how long usernames are blocked once used by a user. How often he can change it. Furthermore it is to clarify if the random part for the username is needed (yes?!) and will the user carry that random part throughout username changes? (no!)

docu/Concepts/TechnicalRequirements/UC_Introduction_of_Gradido-ID.md
```diff
@@ -0,0 +1,141 @@
...
+| passphraseEncryptionType | int    | defines the type of encrypting the passphrase: 1 = email (default), 2 = gradidoID, ... |
+| emailID                  | int    | technical foreign key to the new entity Contact                                        |
+
+##### Email vs emailID
+
+The existing column `email`, will now be changed to the primary email contact, which will be stored as a contact entry in the new `UserContacts` table. It is necessary to decide if the content of the `email `will be changed to the foreign key `emailID `to the contact entry with the email address or if the email itself will be kept as a denormalized and duplicate value in the `users `table.
```
ulfgebhardt, line 34: this is a completely different feature.... multi-email has nothing todo with GradidoID. Even the user-alias is a far fetch from the original statement "GradidoID"

docu/Concepts/TechnicalRequirements/UC_Introduction_of_Gradido-ID.md
```diff
@@ -0,0 +1,141 @@
...
+
+### Database-Schema
+
+#### Users-Table
+
+The entity users has to be changed by adding the following columns.
+
+| Column                   | Type   | Description                                                                            |
+| ------------------------ | ------ | -------------------------------------------------------------------------------------- |
+| gradidoID                | String | technical unique key of the user as UUID (version 4)                                   |
```
ulfgebhardt, line 27: It might be wise to market the alias as `GradidoID` and not the technical key - it has no inherited value and is not human readable

docu/Concepts/TechnicalRequirements/UC_Introduction_of_Gradido-ID.md
```diff
@@ -0,0 +1,141 @@
...
+
+#### new UserContacts-Table
+
+A new entity `UserContacts `is introduced to store several contacts of different types like email, telephone or other kinds of contact addresses.
+
+| Column          | Type   | Description                                                                                                                                                            |
+| --------------- | ------ | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| id              | int    | the technical key of a contact entity                                                                                                                                  |
+| type            | int    | Defines the type of contact entry as enum: Email, Phone, etc                                                                                                           |
+| userID          | int    | Defines the foreign key to the `Users` table                                                                                                                         |
```
ulfgebhardt, line 46: this field was called `GradidoID` before

docu/Concepts/TechnicalRequirements/UC_Introduction_of_Gradido-ID.md
```diff
@@ -0,0 +1,141 @@
...
+
+#### Primary Email Contact
+
+In a one-time migration read for each entry of the `Users `table the `Users.id` and `Users.email` and create for it a new entry in the `UsersContact `table, by initializing the contact-values with:
+
+* id = new technical key
+* type = Enum-Email
+* userID = `Users.id`
+* email = `Users.email`
+* phone = null
```
ulfgebhardt, line 67: do not introduce incomplete features please

docu/Concepts/TechnicalRequirements/UC_Introduction_of_Gradido-ID.md
```diff
@@ -0,0 +1,141 @@
...
+
+After this one-time migration the column `Users.email` can be deleted.
+
+### Adaption of BusinessLogic
+
+The following logic or business processes has to be adapted for introducing the Gradido-ID
+
+#### Read-Write Access of Users-Table especially Email
+
+The ORM mapping has to be adapted to the changed and new database schema.
```
ulfgebhardt, line 80: This is actually part of the migration implementation. New entity models are defined alongside the migration

docu/Concepts/TechnicalRequirements/UC_Introduction_of_Gradido-ID.md
```diff
@@ -0,0 +1,141 @@
...
+
+#### Registration Process
+
+The logic of the registration process has to be adapted by
+
+* initializing the `Users.userID` with a unique UUID
+* creating a new `UsersContact `entry with the given email address and *maincontact* as `usedChannel `
+* set `emailID `in the `Users `table as foreign key to the new `UsersContact `entry
+* set `Users.passphraseEncrpytionType = 2` and encrypt the passphrase with the `Users.userID` instead of the `UsersContact.email`
```
ulfgebhardt, line 89: I think its worth to think about removing this salt altogether and replace it with a server-wide salt. This is a bit less security, but might be worthwhile to not be depending on a per-user-key for encryption. It might even be that our encryption of passwords have a salt already baked in by the sodium library and the whole salt thing is not a security feature at all. This has to be researched tho. I believe that if we change the password encryption algorithm, we should think it from scratch again and not just adjust some parameters. It might be that our current method is completely insecure - if we have a look, just dive a bit deeper and consider current web standards and best practises.

docu/Concepts/TechnicalRequirements/UC_Introduction_of_Gradido-ID.md
```diff
@@ -0,0 +1,141 @@
...
+
+#### Login Process
+
+The logic of the login process has to be adapted by
+
+* search the users data by reading the `Users `and the `UsersContact` table with the email (or alias as soon as the user can maintain his profil with an alias)   as input
```
ulfgebhardt, line 95: do we login with a user alias or still with the email?
Can a user login all his email contacts or just one?
Does an user use multiple emails(usernames) and the same password for all of them?....

docu/Concepts/TechnicalRequirements/UC_Introduction_of_Gradido-ID.md
```diff
@@ -0,0 +1,141 @@
...
+* depending on the `Users.passphraseEncryptionType` decrypt the stored password
```
ulfgebhardt, line 96: This process should also update the password, since the password reaches us in plaintext (super insecure in my opinion, but apparently web-standard).

For security we could use something like SRP6. See [Wikipedia](https://en.wikipedia.org/wiki/Secure_Remote_Password_protocol) [Stanford](http://srp.stanford.edu/design.html)
This would be quite secure and quite applicable to a banking software, but holds quite some complexity.
We then would no longer transmit the password in plain text over the network, but an proof, that we own that password.

docu/Concepts/TechnicalRequirements/UC_Introduction_of_Gradido-ID.md
clauspeterhuebner: As i understand Bernd's definition of an alias, that a user can set a free string as an alias. This must be unique in the community the user is member of.
- So a user can set his name like Ulf or Bernd or Moritz1 or Dario05 or something else..., if it is not used yet in the community
- there is no request for more than one alias per user
- a user can be identified by:
  * UUID (one time set and fix for ever)
  * alias (changeable)
  * email (changeable)
- there is no limit how often the user can change his alias
- if a blocked period after usage or before change will be necessary has to be researched, but currently i find no reason why

docu/Concepts/TechnicalRequirements/UC_Introduction_of_Gradido-ID.md
clauspeterhuebner: That's correct, introducing the GradidoID could be a simply new added attribute.
But...
Introducing a new primary identifier for an user, has effect on the existing primary identifier 'email'.
And what will happen with the existing identifier, if it will be not the primary, but the second or third ranged identifier?
On my opinion the concept for the new GradidoID have to describe the possible changes for other identifiers like email and alias too.

Let us devide this concept perhaps in separated tickets, which can be implemented sequencially for different releases. 
But for the migration efforts it will be more efficient to do all migration at once. For sure the risk of errors will rise up the more changes are done during migration.

docu/Concepts/TechnicalRequirements/UC_Introduction_of_Gradido-ID.md
clauspeterhuebner: I would prefer to use the wording:
GradidoID: UUID = never changeable Identifier
UserID:  alias = unique String = changeable Identifier

docu/Concepts/TechnicalRequirements/UC_Introduction_of_Gradido-ID.md
clauspeterhuebner: not correct.
the userID is the foreign key from UsersContact tabel to the Users table.
The used pattern for foreignkey naming is
<referred table name>ID

-> so this attribut will be renamed from userID to usersID

docu/Concepts/TechnicalRequirements/UC_Introduction_of_Gradido-ID.md
clauspeterhuebner: i do not understand. What is incomplete on your opinion?

these lines describe how are the attributes for a UsersContact entry have to be initialized during the migration in a loop reading all Users entries and create for each Users entry a new UsersContact entry by initializing as follow:
...
UsersContact.type = Enum-Email
UsersContact.usersID = Users.id
UsersContact.email = Users.email
UsersContact.phone = null
...

What is incomplete?

docu/Concepts/TechnicalRequirements/UC_Introduction_of_Gradido-ID.md
clauspeterhuebner: Ok if the migration will use the ORM logic than it must be adapted before.
If the migration will be done directly on database per sql scripts, it will be indepent from ORM.

Feel free how to implement the migration.

docu/Concepts/TechnicalRequirements/UC_Introduction_of_Gradido-ID.md
clauspeterhuebner: i think there is a missunderstanding about the attribute passphraseEncryptionType.

This attribut will only distigue between the 
- encryption of the password is done by the email or 
- will be done by the Users.gradidoID, which will be the UUID.

In future there will be perhaps an other passphraseEncryptionType, which use the UUID but with a new encryption algorithm. 
Which input attributes and values for a password encryption are used can be detected by the value and the definition behind the value of a passphraseEncryptionType and is not an input value for encryption itself

docu/Concepts/TechnicalRequirements/UC_Introduction_of_Gradido-ID.md
clauspeterhuebner: as i understand the user will be able to login with
- gradidoID
- email and only the one declared as main-contact
- alias
The password will be the one for his account, independent from the used login email/alias/gradidoID

docu/Concepts/TechnicalRequirements/UC_Introduction_of_Gradido-ID.md
clauspeterhuebner: that's right and should be done in the future - see #1811

docu/Concepts/TechnicalRequirements/UC_Introduction_of_Gradido-ID.md
ulfgebhardt: A users alias must be stored "forever" tho(old transactions) and a user might be reachable under his old alias for a while - a more general approach with the alias might be useful here.

docu/Concepts/TechnicalRequirements/UC_Introduction_of_Gradido-ID.md
ulfgebhardt: I am a fan of splitting up things ;) KISS

docu/Concepts/TechnicalRequirements/UC_Introduction_of_Gradido-ID.md
ulfgebhardt: the phone number, we have no input nor usage for that yet, we might wanna define what we do with that and how to motivate the user to input it before we introduce it as database field

docu/Concepts/TechnicalRequirements/UC_Introduction_of_Gradido-ID.md
ulfgebhardt: This sounds scarily insecure ;)

docu/Concepts/TechnicalRequirements/UC_Introduction_of_Gradido-ID.md
ulfgebhardt: I fully understand the concept of `passphraseEncryptionType`. What I suggest is to really think about the new version of password encryption which you describe with `will be done by the Users.gradidoID, which will be the UUID.`. We should not jump to conclusions how to do this in the future. We should evaluate current web-standards and compare it to our current process and then make a decision how to change the password algo.

Mogge (approved):

PR #1808, opened 2022-04-20
refactor: Drop Server User Table

Mogge:
## 🍰 Pullrequest
drop server users table. Add isAdmin to users table.

### Issues

- None

### Todo

- [X] None


Mogge:
Has the deployment in backend branch merged

backend/src/graphql/model/User.ts
```diff
@@ -58,11 +58,11 @@ export class User {
 
   // `passphrase` text COLLATE utf8mb4_unicode_ci DEFAULT NULL,
 
+  @Field(() => Boolean)
+  isAdmin: boolean
```
ulfgebhardt, line 62: use date instead of boolean? We would have the info when the user was promoted to the admin

database/entity/0034-drop_server_user_table/User.ts
```diff
@@ -0,0 +1,81 @@
...
+  createdAt: Date
+
+  @Column({ name: 'email_checked', type: 'bool', nullable: false, default: false })
+  emailChecked: boolean
+
+  @Column({ length: 4, default: 'de', collation: 'utf8mb4_unicode_ci', nullable: false })
+  language: string
+
+  @Column({ name: 'is_admin', type: 'bool', nullable: false, default: false })
+  isAdmin: boolean
```
ulfgebhardt, line 62: date?

database/migrations/0034-drop_server_user_table.ts
```diff
@@ -0,0 +1,37 @@
...
+add isAdmin COLUMN to users TABLE */
+
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+export async function upgrade(queryFn: (query: string, values?: any[]) => Promise<Array<any>>) {
+  await queryFn('ALTER TABLE `users` ADD COLUMN `is_admin` datetime DEFAULT NULL AFTER `language`;')
+
+  await queryFn(
+    'UPDATE `users` AS `users`, (SELECT  * FROM `server_users`) AS `server_users` SET users.`is_admin` = server_users.`modified` WHERE users.`email` IN (SELECT email from `server_users`);',
```
ulfgebhardt, line 11: Does this work? The `(Select * ...)` in the select statement seems dagerously unspecific. I would use a join here - tho this might be a valid solution.

I would do something like this:
` UPDATE users INNER JOIN server_users ON email = email SET is_admin = modified`

ulfgebhardt (approved):

PR #1810, opened 2022-04-20
[WIP] 🍰 Refactor THX Page

### LINKED ISSUE #1697: 🔧 [Refactor] Thx Page
## 🔧 Refactor ticket
The thx page is a complete mess. I suggest to get rid of this page and create a Response component, which is handled by the page that now redirects to thx. So we have one layout for the response page, but the setup (like button text, where to link to, text and header) is done via props on the page which needs to render a response.

Tirokk:
## 🍰 Pullrequest

Refactor THX page.

### Issues

- fixes #1697

### Todo

- [ ] search for `forgot-password`, `checkEmail`, `reset-password`
  - [ ] and think if changes are neccessary in links
- [ ] change position of languages snippets in language files?
  - [ ] especially for `site.thx.XXX`
- [ ] clean up: search for `Wolle`

### DIFF
--- a/frontend/src/components/Message/Message.vue
+++ b/frontend/src/components/Message/Message.vue
@@ -1,19 +1,27 @@
 <template>
   <div>
     <!-- Header -->
-    <div class="header py-7 py-lg-8 pt-lg-9">
+    <div class="header py-lg-6">
       <b-container>
         <div class="header-body text-center mb-7">
           <!-- eslint-disable-next-line @intlify/vue-i18n/no-dynamic-keys-->
-          <p class="h1">{{ $t(displaySetup.headline) }}</p>
+          <p class="h1 test-message-headline">{{ $t(displaySetup.headline) }}</p>
           <!-- eslint-disable-next-line @intlify/vue-i18n/no-dynamic-keys-->
-          <p class="h4">{{ $t(displaySetup.subtitle) }}</p>
+          <p class="h4 test-message-subtitle">{{ $t(displaySetup.subtitle) }}</p>
           <hr />
-          <b-button v-if="$route.params.code" :to="`/login/${$route.params.code}`">
+          <b-button
+            v-if="$route.params.code"
+            class="test-message-button"
+            :to="`/login/${$route.params.code}`"
+          >
             <!-- eslint-disable-next-line @intlify/vue-i18n/no-dynamic-keys-->
             {{ $t(displaySetup.button) }}
           </b-button>
-          <b-button v-else-if="displaySetup.linkTo" :to="displaySetup.linkTo">
+          <b-button
+            v-else-if="displaySetup.linkTo"
+            class="test-message-button"
+            :to="displaySetup.linkTo"
+          >
             <!-- eslint-disable-next-line @intlify/vue-i18n/no-dynamic-keys-->
             {{ $t(displaySetup.button) }}
           </b-button>
@@ -26,13 +34,19 @@
 
 <script>
 const textFields = {
-  forgotPassword: {
+  'forgot-password-success': {
     headline: 'site.thx.title',
     subtitle: 'site.thx.email',
     button: 'login',
     linkTo: '/login',
   },
-  resetPassword: {
+  'forgot-password-error': {
+    headline: 'site.thx.errorTitle',
+    subtitle: 'error.email-already-sent',
+    button: 'login',
+    linkTo: '/login',
+  },
+  'reset-password': {
     headline: 'site.thx.title',
     subtitle: 'site.thx.reset',
     button: 'login',
@@ -44,7 +58,7 @@ const textFields = {
     button: 'site.login.signin',
     // linkTo: '/login',
   },
-  checkEmail: {
+  'check-email': {
     headline: 'site.thx.title',
     subtitle: 'site.thx.checkEmail',
     button: 'login',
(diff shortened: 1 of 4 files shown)


frontend/src/components/Message/Message.vue
```diff
@@ -0,0 +1,38 @@
+<template>
+  <div>
+    <!-- Header -->
+    <div class="header py-lg-6">
+      <b-container>
+        <div class="header-body text-center mb-7">
+          <!-- eslint-disable-next-line @intlify/vue-i18n/no-dynamic-keys-->
+          <p class="h1 test-message-headline">{{ $t(headline) }}</p>
```
ulfgebhardt, line 8: as we discussed before, this method will have the problem that the compiler will not be able to determine the used text by itself. Consider to have the property be the actual text already, resolved by `$t` so that the localization will not be task of this component.

frontend/src/components/Message/Message.vue
```diff
@@ -0,0 +1,38 @@
...
+
+<script>
+export default {
+  name: 'Message',
+  props: {
+    headline: { type: String, required: true },
+    subtitle: { type: String, required: true },
+    buttonText: { type: String, required: false, default: null },
+    linkTo: { type: String, required: false, default: null },
+    code: { type: String, required: false, default: null },
```
ulfgebhardt, line 35: Also consider to not include the code part in this component but have the component calling this one to provide a correct `linkTo` including the code already.

**I am unsure which way is better for this question**

frontend/src/pages/ForgotPassword.vue
```diff
@@ -36,24 +36,46 @@
         <router-link to="/login" class="mt-3">{{ $t('back') }}</router-link>
       </div>
     </b-container>
+    <b-container v-else class="mt--8 p-1">
+      <!-- eslint-disable @intlify/vue-i18n/no-dynamic-keys-->
```
ulfgebhardt, line 40: is this still needed?

frontend/src/pages/ForgotPassword.vue
```diff
@@ -66,15 +88,17 @@ export default {
...
         .catch(() => {
-          this.toastError(this.$t('error.email-already-sent'))
-          this.$router.push('/thx/forgotPassword')
+          this.showPageMessage = true
+          this.success = false
         })
     },
   },
   created() {
+    // Wolle: what shall happen here? change needed?
```
ulfgebhardt, line 101: We reuse the form for Forget Password for initially setting the password as well. This is done via a different URL provided after registering, but is essentially the same process.
I would suggest to keep this as it is for now. Later on this might be worth a refactor aswell: Have two pages and reuse the same password set-component on both.

frontend/src/components/Message/Message.vue
Tirokk: Thanks for the hint.
I left the `$t` here unintentional and faulty.

frontend/src/pages/ForgotPassword.vue
Tirokk: Yes, unfortunatelly!

frontend/src/components/Message/Message.vue
Tirokk: Now since I give the `linkTo` as general prop your suggestion is the best way.

I put it here before just to don't forget about it and care for a solution later if more gets clear.

frontend/src/pages/ForgotPassword.vue
Tirokk: Supi 🤗

Tirokk:
I get this error on `yarn lint --fix` even though `error` is excluded in locales the eslint.

Bildschirmfoto 2022-04-28 um 10 11 37

How can this error be ignored? @ogerly @mogge @ulfgebhardt @Elweyn

frontend/src/pages/Register.vue
```diff
@@ -105,6 +105,7 @@
                       </b-form-checkbox>
                     </b-col>
                   </b-row>
+                  <!-- Wolle: remove this? or shall the alert or a toaster be shown? -->
                   <b-alert
                     v-if="showError"
                     show
```
Tirokk, line 111: I guess, the alert shall not be user anymore? @Mogge @ogerly @ulfgebhardt

frontend/src/pages/Register.vue
```diff
@@ -240,20 +264,43 @@ export default {
...
+              break
+            default:
+              this.messageError = this.translateErrorMessage(error.message)
+              break
+          }
+          // Wolle: this.toastError(this.$t('error.email-already-sent'))
+          // Wolle: shall the alert be replaced by a toaster or shall only the page message be shown?
         })
     },
+    // Wolle: remove this?
```
Tirokk, line 287: For me it would be nice if we have an additional error toast message even when we show the error directly on the page.

This gets more attention of the user in my eyes.
@ogerly @Mogge

frontend/src/pages/ResetPassword.vue
```diff
@@ -1,56 +1,68 @@
...
-              <h1>{{ $t(displaySetup.title) }}</h1>
-              <div class="pb-4">
-                <span>
-                  <!-- eslint-disable-next-line @intlify/vue-i18n/no-dynamic-keys-->
-                  {{ $t(displaySetup.text) }}
-                </span>
-              </div>
-            </b-col>
-          </b-row>
+    <div v-if="!showPageMessage">
```
Tirokk, line 3: Is this file be used at all?
@ogerly @Mogge @ulfgebhardt

PR #1822, opened 2022-04-22
Fix: cross-env for windows

clauspeterhuebner:
## 🍰 Pullrequest

fixes yarn for windows in backend, admin, frontend, database packages.json by adding cross-env


ogerly (approved):
I wanted to do the seeding process with Claus Peter and with Windows machines I have the problem that they do not understand TZ=UTC. cross-env solves this problem.

Issue #955, opened 2021-10-08
💥 [DevOps] Enforce 60% coverage (backend)


ulfgebhardt:
## 💥 DevOps ticket

Enforce 60% coverage (backend)

ulfgebhardt:
This is not done yet or is it? @Mogge

PR #1827, opened 2022-04-25
feat: More User Resolver Tests

Mogge:
## 🍰 Pullrequest
Adds more tests to the user resolver.

### Issues
- relates #955

### Todo

- [X] None


backend/src/graphql/resolver/UserResolver.ts
```diff
@@ -529,15 +529,7 @@ export class UserResolver {
   @Mutation(() => Boolean)
```
ulfgebhardt, line 529: Remove line 254 & 255

```
// TODO: Check if we can use updateUserInfos
// await this.updateUserInfos({ publisherId }, { pubKey: loginUser.pubKey })
```

PR #1846, opened 2022-04-27
🍰 Add NVM Config Files To '.gitignore'

### LINKED ISSUE #1845: 💥 [DevOps] Set Node Version By Config File In Folders
## 💥 DevOps ticket

In the folders `frontend`, `admin`, `database`, and `backend` I have to change the node version with the Node Version Manager NVM always to `nvm use 17.9.0`, after every restart of my computer.

For the [ocelot.social](ocelot.social) I need an other one.

I would like to add a NVM config file `.nvmrc` to these folders.
See:

[Set node version differently for specific project(folder) via NVM](https://stackoverflow.com/questions/60051990/set-node-version-differently-for-specific-projectfolder-via-nvm)

Tirokk:
## 🍰 Pullrequest

Add NVM config files.

### Issues

- fixes #1845

### Todo

- [x] with this solution you still have to run `nvm use` after comupter restart.
  - can we organize that `nvm use` will run always when we run `yarn install`? @Mogge 
  - no
- [x] just ignore the files I have decided with @Mogge and @ulfgebhardt


.nvmrc
```diff
@@ -0,0 +1 @@
+v17.9.0
```
Mogge, line 1: `yarn` will not run with this version

Mogge (commented):
This way you force other developers to have a specific node installation on their machine. I personally do not like this. If this solution works for you, I suggest to add the `.nvmrc` to `.gitignore`

.nvmrc
Tirokk: My yarn runs with this version.
But like we discussed: I put it in ignore. @Mogge

PR #1847, opened 2022-04-27
1843 integrate and test vue clipboards

### LINKED ISSUE #1843: 🔧 [Refactor] integrate and test the behaviour of vue-clipboards
integrate and test the behaviour of vue-clipboards
https://www.npmjs.com/package/vue-clipboards

ogerly:
## 🍰 Pullrequest
I have integrated the package vue-clipboards. 
let this copy the text of the link first and if this throws an error then the current navigator.clipboard api is started. 

i guess this covers the case that older browsers can copy and if the newer browsers have the old function disabled then automatically by throwing an error of the old browser function from the vue-clipboard the navigator.clipboard api is thrown. 
so now two cases should be covered. 

vue-clipboard2 had a behaviour which was buggy. i tested it in our wallet and removed it again. 

note: i tested this in my current browsers. i can't promise that the problems will work in other older browser versions.  

### Issues
- fixes #1843

### DIFF
--- a/frontend/src/components/TransactionLinks/TransactionLink.vue
+++ b/frontend/src/components/TransactionLinks/TransactionLink.vue
@@ -23,7 +23,7 @@
                 class="test-copy-link"
                 v-clipboard="link"
                 @success="toastSuccess($t('gdd_per_link.link-copied'))"
-                @error="CopyLink"
+                @error="copyLink"
               >
                 <b-icon icon="clipboard"></b-icon>
                 {{ $t('gdd_per_link.copy') }}
@@ -94,7 +94,7 @@ export default {
     validUntil: { type: String, required: true },
   },
   methods: {
-    CopyLink() {
+    copyLink() {
       navigator.clipboard
         .writeText(this.link)
         .then(() => {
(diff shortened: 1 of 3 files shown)


frontend/src/components/TransactionLinks/TransactionLink.vue
```diff
@@ -18,7 +18,13 @@
                 <b-icon icon="three-dots-vertical"></b-icon>
               </template>
 
-              <b-dropdown-item v-if="validLink" class="test-copy-link" @click="copy">
+              <b-dropdown-item
+                v-if="validLink"
+                class="test-copy-link"
+                v-clipboard="link"
+                @success="toastSuccess($t('gdd_per_link.link-copied'))"
+                @error="copyLink"
```
ulfgebhardt, line 26: are you sure you fix the problem this way?

ogerly: no

Mogge: LOL

Issue #1784, opened 2022-04-14
🚀 [Feature] Federation 1.0: Implement DHT-Node as part of Apollo


ulfgebhardt:
## 🚀 Feature

Implement DHT-Node as part of Apollo like this:

![image.png](https://images.zenhubusercontent.com/5ca1e741e2f21b570ab87e6e/97470c34-86b5-4ef1-911f-de5a7da9257a)

We use the existing typeorm infrastructure of Apollo to connect to the database, but the node must be a separate component, since we want to externalize it at some point.

For details please take a look in the technical concept of federation:
https://github.com/gradido/gradido/blob/1676-feature-federation-technical-concept/docu/Concepts/TechnicalRequirements/Federation.md

Preparation of Apollo and typeORM will be implemented by #1887 

![image](https://user-images.githubusercontent.com/1238238/163461879-477b99d8-af0d-42c2-9da5-f440cb3a3afe.png)

We need the following functionality for the node:
- announce https://github.com/hyperswarm/dht#const-stream--nodelookuptopic-options
- lookup https://github.com/hyperswarm/dht#const-stream--nodelookuptopic-options
- unannounce https://github.com/hyperswarm/dht#await-nodeunannouncetopic-keypair-options

for announce&lookup the technical concept contains details here
https://github.com/gradido/gradido/blob/1676-feature-federation-technical-concept/docu/Concepts/TechnicalRequirements/Federation.md#sequence-joinconnect-joinconnect

Repeat announceing and lookup in an configurable interval (like 10min as default)

PR #1874, opened 2022-05-04
feat(backend): setup hyperswarm

Mogge:
## 🍰 Pullrequest

Use DHT hyperswarm when DHT_TOPIC is set in .env

### Issues
- relates #1784 

### Todo

- [x] Get it running


Mogge:
For some reason the topic generated by the prove of concept does not match the topic generated by this PR :(

backend/src/federation/index.ts
```diff
@@ -0,0 +1,122 @@
...
+          successfulRequests.push(remotePubKey)
+        })
+        // pipe it somewhere like any duplex stream
+        // process.stdin.pipe(noiseSocket).pipe(process.stdout)
+      })
+    }, POLLTIME)
+  } catch (err) {
+    console.log(err)
+  }
+}
```
clauspeterhuebner, line 122: it seems to be clear logically, but its hard to give a statement about correctness or functionality with a unknown framework. So we will see and research by using it.

backend/src/server/createServer.ts
```diff
@@ -52,6 +55,11 @@ const createServer = async (
     throw new Error('Fatal: Database Version incorrect')
   }
 
+  // start DHT hyperswarm when DHT_TOPIC is set in .env
+  if (CONFIG.DHT_TOPIC) {
+    await startDHT(con, CONFIG.DHT_TOPIC)
+  }
+
```
clauspeterhuebner, line 62: As i understand this here, the hyperswarm will be in the first steps part of our apollo-server.
In the future we should extract it as a separated modul, which will communicate with the existing moduls, right?

deployment/bare_metal/.env.dist
```diff
@@ -26,7 +26,7 @@ COMMUNITY_REDEEM_CONTRIBUTION_URL=https://stage1.gradido.net/redeem/CL-{code}
 COMMUNITY_DESCRIPTION="Gradido Development Stage1 Test Community"
 
 # backend
-BACKEND_CONFIG_VERSION=v10.2022-09-20
+CONFIG_VERSION=v11.2022-10-27
```
clauspeterhuebner, line 29: was the replaced BACKEND_CONFIG_VERSION-property never used in the baremetal-scripts?
If it is in use, where are the changes about this exchange to CONFIG_VERSION?

clauspeterhuebner (approved):
Some open questions, but i approve it to go further.

deployment/bare_metal/.env.dist
Mogge: Good catch. Must be `BACKEND_CONFIG_VERSION` here. As in the `backend/.env.template` it states:
`CONFIG_VERSION=$BACKEND_CONFIG_VERSION`

ulfgebhardt (approved):

PR #1912, opened 2022-05-19
refactor: Use Logger Categories

Mogge:
## 🍰 Pullrequest
Create different log files for backend and apollo via categories.
Log apollo on info.

### Issues

- None

### Todo

- [X] None


backend/src/server/plugins.ts
```diff
@@ -32,16 +32,16 @@ const logPlugin = {
...
       willSendResponse(requestContext: any) {
-        if (requestContext.context.user) logger.trace(`User ID: ${requestContext.context.user.id}`)
+        if (requestContext.context.user) logger.info(`User ID: ${requestContext.context.user.id}`)
         if (requestContext.response.data)
-          logger.trace(`Response-Data:
+          logger.info(`Response-Data:
 ${JSON.stringify(requestContext.response.data, null, 2)}`)
         if (requestContext.response.errors)
-          logger.trace(`Response-Errors:
+          logger.info(`Response-Errors:
```
clauspeterhuebner, line 44: For errors isn't it better to use logger.error() ?

Mogge: Right! Good catch

Tirokk (approved):
Great! 🚀💫

Works fine for me! 🙏🏼

PR #1960, opened 2022-06-01
refactor: 🍰 Delete `user_setting` Table From DB

### LINKED ISSUE #1948: 🔧 [Refactor] delete settings animationcoin in database
## 🔧 Refactor ticket
delete settings animationcoin in database

via migration

Leftover of #1946

Tirokk:
## 🍰 Pullrequest

Delete `user_setting` table from db.

### Issues

- fixes #1948

### Question

Is this really clever to do?
Because we will need user settings in the future.

If I would do it consequently I would have to remove the class `UserSetting` and all related stuff as well.
We would have to implement it new in the future.

### Todo

- [x] fixed linting of database `README.md`
- [x] remove `UserSetting` and all related stuff?
  - search for:
    - [x] `UserSetting`
    - [x] `UserSettingRepository`
    - [x] `user_setting`
- [x] questions:
  - [x] should file `database/entity/UserSetting.ts` be deleted or shall it stay for ever?
    - [x] i have renamed all the imports from `../UserSetting` to `../0002-add_settings/UserSetting`. is that correct in case?
  - [x] is the reference in `database/entity/0002-add_settings/UserSetting.ts` for the user import lasst fitting one? or was it if unchanged with the actual `../User` correct?
- [x] cleanup: search for `Wolle`


Tirokk:
This is how the db looks if a user setting is set:

Bildschirmfoto 2022-06-01 um 12 23 07

database/entity/0002-add_settings/User.ts
```diff
@@ -1,5 +1,5 @@
 import { BaseEntity, Entity, PrimaryGeneratedColumn, Column, OneToMany } from 'typeorm'
-import { UserSetting } from '../UserSetting'
+import { UserSetting } from '../0002-add_settings/UserSetting'
```
Tirokk, line 2: Because I deleted the `UserSetting.ts` file in the entity folder I changed the reference here and in all the other similar cases.

Is that correct @Mogge ? Or shall it stay undeleted for ever?
To leave it undeleted would be confusing for our future development I think …

database/entity/0002-add_settings/UserSetting.ts
```diff
@@ -1,5 +1,5 @@
 import { BaseEntity, Entity, PrimaryGeneratedColumn, Column, ManyToOne } from 'typeorm'
-import { User } from '../User'
+import { User } from '../0034-drop_server_user_table/User' // Wolle: last fitting one or old was correct?
```
Tirokk, line 2: Because the new `database/entity/0037-drop_server_user_table/User.ts` file does not fit to this reference I changed it to the last fitting one.

Is that correct @Mogge ? Or shall it reference to the actual for ever?
To leave the reference to a not fitting dataset would be confusing for our future development I think …

Mogge (approved):

ogerly (approved):

PR #2001, opened 2022-06-22
automatic session logout with info modal

### LINKED ISSUE #1736: 🐛 [Bug] When session is expired redeem link is not transported
## 🐛 Bugreport
When session is expired redeem link is not transported properly.

- Use Vue Times <https://www.npmjs.com/package/vue-timers>
- Check every 10 seconds if the auth token is still valid.
- 1 minute before the session expires please open a modal, with time expiration and the possibility to extend the session

### LINKED ISSUE #1781: 🚀 [Feature] On Auto Logout Show Automatically The Login Page
## 🚀 Feature

```gherkin
Feature: On auto logout go to login page
As a user I want to recognice if I am already logged out or not.
```

At the moment we have this szenario:

After a pause I went back to the Gradido page and do a transaction.
When I do a transaction and click send I am redirected to the login page.
If I login again I eventually can see that the transaction wasn't done.
I can guess that I was automatically logged out already,

ogerly:
## 🍰 Pullrequest
  es kann wie folgt getestet werden.
1. setze die zeit auf 2m (zwei Minuten) in **backend/src/config/index.ts**  Zeile: 28   

### Issues
 - fixes #1736 
 - fixes #1781 
 
FireShot Capture 1225 -  - localhost

 

https://user-images.githubusercontent.com/1324583/175221192-0de2e4f6-d733-48df-a474-6260caac058b.mp4


ogerly:
I just noticed the following error: 
I have to stop the first interval in the time where the 60 seconds are counted down ... otherwise every 15 seconds a -1 is added. 
if you click on extend, the first interval must be restarted.

ogerly:
there is now also a hint sound every 13 seconds when the session expires ... even if the tab should run in the background. 
if we leave this in then should be given sometime in the settings for the user ... a way to turn this off. 
I chose it but so that it is not annoying. I find it discreet. that can of course be a matter of taste.
https://user-images.githubusercontent.com/1324583/175220439-4ab93b11-dc48-4721-a776-43feb937938c.mp4

ogerly:
FireShot Capture 1226 -  - localhost

ogerly:
Bildschirmfoto vom 2022-06-27 09-05-03

ogerly:
image

frontend/src/store/store.test.js
```diff
@@ -59,6 +65,23 @@ describe('Vuex store', () => {
...
       })
+      describe('token has a value', () => {
+        it('sets the state of tokenTime', () => {
+          const state = { token: null, tokenTime: null }
+          token(state, 'token')
+          expect(jwtDecode).toBeCalledWith('token')
+          expect(state.tokenTime).toEqual('1234')
+        })
+      })
+      describe('token has null value', () => {
```
Tirokk, line 76:
```suggestion
      })
      
      describe('token has null value', () => {
```

frontend/src/store/store.test.js
```diff
@@ -59,6 +65,23 @@ describe('Vuex store', () => {
         token(state, '1234')
         expect(state.token).toEqual('1234')
       })
+      describe('token has a value', () => {
```
Tirokk, line 68:
```suggestion
      })
      
      describe('token has a value', () => {
```

Tirokk (changes requested):
Everything looks good in general.
But I have some suggestions …

Mogge (dismissed):

PR #2016, opened 2022-06-29
change slidertime from 13 to 10 seconds

ogerly:
## 🍰 Pullrequest
chnage slidertime from 13 to 10 seconds


ogerly:
After checking with bernd, let's stick with 13 seconds after all.

PR #2029, opened 2022-07-04
[POW] Migrate transaction to valid dataset for gradido node

einhornimmond:
## 🍰 Pullrequest
- split invalid creation transaction with 3000 GDD in three separate transactions
- update all creation transaction with creation_date == balance_date, set creation_date to date mentioned in memo or if no month was mentioned, simply move creation_date on month before
- use a separate table (`transactions_temp`) for transaction with updated dates as long as it isn't finished, so it is more easy to compare, debug and retry

### Issues
- relates #1971


database/migrations/0041-update_transactions_for_blockchain.ts
```diff
@@ -0,0 +1,114 @@
...
+  
+  await queryFn(
+    `INSERT INTO \`transactions\`(
+            \`user_id\`, \`previous\`, \`type_id\`, \`amount\`, \`balance\`,
+            \`balance_date\`, \`decay\`, \`memo\`, \`creation_date\`
+          ) VALUES(
+            275, 150, 1, 1000, 2000,
+            ?, 0, ?, ?
+          )`,
+    [creationDate, transactionMemos[1], creationDate],
```
Mogge, line 31:
```suggestion
    [creationDate, transactionMemos[1], new Date(creationDate.getFullYear(), creationDate.getMonth() - 1, 1)],
```

database/migrations/0041-update_transactions_for_blockchain.ts
```diff
@@ -0,0 +1,114 @@
...
+  )
+  await queryFn(
+    `INSERT INTO \`transactions\`(
+            \`user_id\`, \`previous\`, \`type_id\`, \`amount\`, \`balance\`,
+            \`balance_date\`, \`decay\`, \`memo\`, \`creation_date\`
+          ) VALUES(
+            275, LAST_INSERT_ID(), 1, 1000, 3000,
+            ?, 0, ?, ?
+          )`,
+    [creationDate, transactionMemos[2], creationDate],
```
Mogge, line 41:
```suggestion
    [creationDate, transactionMemos[2], new Date(creationDate.getFullYear(), creationDate.getMonth() - 2, 1)]],  
```

database/migrations/0041-update_transactions_for_blockchain.ts
```diff
@@ -0,0 +1,114 @@
...
+  interface ReplaceSet {
+    monthName: string
+    monthValue: number
+    yearValue: number
+  }
+  const replaceSets: ReplaceSet[] = [
+    { monthName: '%Dez%', monthValue: 12, yearValue: 2019 },
+    { monthName: '%Jan%', monthValue: 1, yearValue: 2020 },
+    { monthName: '%Feb%', monthValue: 2, yearValue: 2020 },
+    { monthName: '%M_rz%', monthValue: 3, yearValue: 2020 },
```
Mogge, line 64: Das funktioniert?

Mogge (changes requested):

Mogge:
I do not understand why the `creation_date` must be different than `balance_date`.
When I run 
```
SELECT user_id, creation_date, sum(amount) as sum from transactions where balance_date = creation_date group by user_id, creation_date;
```
then I only have one entry, that is not valid. But this can be fixed as indicated in my review.

database/migrations/0041-update_transactions_for_blockchain.ts
einhornimmond: Je nach dem was du darunter verstehst. 
Ich konnte leider nicht testen ob es auf jeglicher möglichen Konfiguration auf der Welt funktioniert.
Aber zumindest bei mir funktioniert es. 
- unter Windows 10 und unter Ubuntu, über den C-Mariadb-Connector also vom Blockchain-Connector aus aufgerufen
- unter Windows 10 über phpmyadmin das bei xampp dabei ist
- unter Linux über apollo database migration.. also genau das hier

Ich habe es nicht mit Docker gestestet, nur nativ.

database/migrations/0041-update_transactions_for_blockchain.ts
```diff
@@ -0,0 +1,1919 @@
...
+*/
+
+  // Erste Schöpfung, viel Spaß damit ;), original creation date: 2019-12-17 13:06:13
+  await queryFn(
+    `UPDATE \`transactions\` SET creation_date = '2019-11-17 13:06:13' WHERE \`id\` = 1`,
+  )
+
+  // Erste Schöpfung, viel Spaß damit ;), original creation date: 2019-12-17 13:06:16
+  await queryFn(
+    `UPDATE \`transactions\` SET creation_date = '2019-11-17 13:06:16' WHERE \`id\` = 2`,
```
ulfgebhardt, line 128: You habe to correct every following Transaktion for that particular user, since the Transaktion table is considered fixed and unchangeable we cache the data for the final balance in a transaction. If you change the date of an existing transaction you always change the final balance of all following transactions, since the decay will tick more or less depending if you move the date forward or backward. If its before the decay start you still have to adjust the balance of all 3 transactions.

The simplest solution to your problem was the initially proposed one when you raised your validation problem: make an exception:
"Every transaction older or equal to $date can have a maximum of $amount while when newer then $date the amount mussnt be higher then $amount_b"
We do the same with decay:
"If transaction date older then $date do not calculate decay, if newer or equal to $date calculate decay."

database/migrations/0041-update_transactions_for_blockchain.ts
```diff
@@ -0,0 +1,1919 @@
...
+
+  // rewind transaction to split
+  await queryFn(
+    `UPDATE \`transactions\` set \`amount\` = 3000, \`memo\` = 'Aktives Grundeinkommen für GL. Dez, Jan, Feb' 
+     WHERE \`id\` = 150`,
+  )
+
+  await queryFn(`UPDATE \`transactions\` set \`previous\` = 150 WHERE \`id\` = 278`)
+  // drop temp table
+  await queryFn(`DROP TABLE IF EXISTS \`transactions_temp\``)
```
ulfgebhardt, line 1918: Screenshot_20220710-125001_Firefox

database/migrations/0041-update_transactions_for_blockchain.ts
einhornimmond: Decay start time: 2021-05-13 17:46:31-0000
Creation Date of last transaction which will be changed by my migration: 2020-05-03 09:00:07
So none of this transactions are influenced from decay.
But you are right, maybe a follow up transaction cames after the decay, I will check this.

Mogge (changes requested):
run
```
yarn lint --fix
```
in database folder. @ulfgebhardt has a point. It is not only about the transaction where the date is manipulated. It is also about the date of the following transaction:
```sql
WHERE previous = id
```
This date must also be before the decay start block for all transactions. If not, we cannot do it like this.

Mogge (approved):
You are right. Of course the creation date does never effect the decay calculation

PR #2034, opened 2022-07-05
add code for moving user creation date if transaction before exist

einhornimmond:
### Issues

- None

### Todo
[x] Test with Gradido Node


Mogge (commented):
Looks good!

Mogge (changes requested):
I suggest to rename the migration file to `0041-...`
So we can merge this one. To get it pass the CI tests, you have to update the `DB_VERSION` in `backend/src/config/index.ts`

einhornimmond:
> I suggest to rename the migration file to `0041-...` So we can merge this one. To get it pass the CI tests, you have to update the `DB_VERSION` in `backend/src/config/index.ts`

This get's the 42 because the 41 is already used from #2029, so than we can merge this one after merging the #2029

Mogge (approved):
I really suggest to merge this first.

backend/src/config/index.ts
```diff
@@ -10,7 +10,7 @@ Decimal.set({
 })
 
 const constants = {
-  DB_VERSION: '0040-add_contribution_link_id_to_user',
+  DB_VERSION: '0042-move_users_creation_date',
```
Mogge, line 13:
```suggestion
  DB_VERSION: '0041-move_users_creation_date',
```

database/migrations/0042-move_users_creation_date.ts
```diff
@@ -0,0 +1,130 @@
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
```
Mogge, line 1: Change file name to `0041`

Mogge (changes requested):

database/migrations/0042-move_users_creation_date.ts
einhornimmond: > I really suggest to merge this first.

ok, should also work in this order.

Mogge (approved):

PR #2050, opened 2022-07-12
 chore: 🍰 Change `image` Entries In Docker Compose Files And Get Apple M1 Running

### LINKED ISSUE #2049: 💥 [DevOps] Change `image` Entries In `docker-compose*.yml` Files
## 💥 DevOps ticket

Change `image` entries in:

- `docker-compose.override.yml`
- `docker-compose.test.yml`?
- `docker-compose.yml`

Because:

- in case there would be a Docker image in the repository it will be downloaded instead of building it fresh
- in newer Docker or Docker Desktop versions an error is shown if the image is not available in the image repo, see below

Bildschirmfoto 2022-07-12 um 18 22 45

Tirokk:
## 🍰 Pullrequest

Remove `image` entries in:

- `docker-compose.override.yml`
- `docker-compose.test.yml`?
- `docker-compose.yml`

Get Apple M1 running.

Because:

- in case there would be a Docker image in the repository it will be downloaded instead of building it fresh
- in newer Docker or Docker Desktop versions an error is shown if the image is not available in the image repo, see below

Bildschirmfoto 2022-07-12 um 18 22 45

### Issues

- fixes #2049

### Questions

- [x] can the `image` entries really be removed from `docker-compose.yml`?
  - [x] or do we have to override them by emptying them somehow by `docker-compose.override.yml`?
  - [x] !!! we rename them
- [x] remove `image` entries from `docker-compose.test.yml` as well?
  - [x] !!! no
- [ ] may change command `docker-compose` to new `docker compose` notation?
  - see https://docs.docker.com/compose/cli-command-compatibility/
- [x] fix Apple M1 Docker Compose build problem, see https://stackoverflow.com/questions/65612411/forcing-docker-to-use-linux-amd64-platform-by-default-on-macos
  - [x] may add `docker-compose.apple-m1.override.yml` for overriding?
    - [x] refine docs with Apple M1 informations and the new command?
      - `$ docker-compose -f docker-compose.yml -f docker-compose.override.yml -f docker-compose.apple-m1.override.yml up`
  - [x] after that `export DOCKER_DEFAULT_PLATFORM=linux/amd64` is an other possibility that work for me
    - [ ] should be set then in a global and stable manner
- [ ] would it be possible and wishfull to change our developing `locale` port to `3000`?
  - then I wouldn't have to switch on and off my own locale Apache server all the time …


docker-compose.apple-m1.override.yml
```diff
@@ -0,0 +1,52 @@
...
+  nginx:
+    platform: linux/amd64
+  
+networks:
+  external-net:
+  internal-net:
+    internal: true
+
+volumes:
+  db_vol:
```
Tirokk, line 51: Ups …
This has to be removed!

docker-compose.override.yml
```diff
@@ -5,7 +5,7 @@ services:
   # FRONTEND #############################################
   ########################################################
   frontend:
-    image: gradido/frontend:development
+    # image: gradido/frontend:development
```
Tirokk, line 8: Comment out this and all the others and leave them in the code?
Or remove them?

Mogge (commented):
What is the motivation of this PR?

docker-compose.apple-m1.override.yml
Tirokk: Done

docker-compose.yml
```diff
@@ -10,7 +10,7 @@ services:
   # FRONTEND #############################################
   ########################################################
   frontend:
-    image: gradido/frontend:latest
+    # image: gradido/frontend:latest
```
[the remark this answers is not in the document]
Tirokk: Renamed

docker-compose.override.yml
Tirokk: Renamed

Tirokk:
> What is the motivation of this PR?

@Mogge 

- The main motivation of this PR is at first to avoid pulling of the Docker images from our repo instead of building them locally. Add the reason why as a comment.
- Second is to give developers with Apple M1 enviroments the possibility to develop.
- Third is to add a Docker documentation how to do the second and how to analyze Docker builds.

ogerly (approved):

DOCKER_MORE_CLOSELY.md
```diff
@@ -0,0 +1,44 @@
...
+
+### Docker Compose Override File For Apple M1 Platform
+
+For Docker compose `up` or `build` commands, you can use our Apple M1 override file that specifies the M1 platform:
+
+```bash
+# in main folder
+
+# for development
+$ docker compose -f docker-compose.yml -f docker-compose.override.yml -f docker-compose.apple-m1.override.yml up
```
Mogge, line 24:
```suggestion
$ docker-compose -f docker-compose.yml -f docker-compose.override.yml -f docker-compose.apple-m1.override.yml up
```

DOCKER_MORE_CLOSELY.md
```diff
@@ -0,0 +1,44 @@
...
+
+# for production
+$ docker compose -f docker-compose.yml -f docker-compose.apple-m1.override.yml up
```
Mogge, line 27:
```suggestion
$ docker-compose -f docker-compose.yml -f docker-compose.apple-m1.override.yml up
```

Tirokk: The new command is without the dash @Mogge , but both will work for a while …
See https://docs.docker.com/compose/cli-command-compatibility/

DOCKER_MORE_CLOSELY.md
Tirokk: same same 👇🏼

Mogge (approved):

PR #2100, opened 2022-07-28
fix: Use Inner Join for Contribution and User

Mogge:
## 🍰 Pullrequest
Fixes the `Cannot read properties of null (reading 'id')` error for the list all contributions query.
 
### Issues

- None

### Todo

- [X] None


Elweyn (approved):
Seems good!
Working on localhost & don't throw any errors anymore

PR #2102, opened 2022-07-28
Fix navbar community item

Elweyn:
## 🍰 Pullrequest

In the mobile devices we could not see the community item & Remove block that insert an exclamation point for members area.

### Issues

- None

### Todo

- [X] Add community item
- [X] Remove exclamation point for members area


mahula:
Looks good.

mahula (approved):
Looks good.

PR #2107, opened 2022-08-04
fix: 🍰 Add `emailChecked` Before Changing `optIn` State & Log Error On klicktipp Middleware

### LINKED ISSUE #2105: 🚀 [Feature] Add Klicktipp-Check and -Registration to Password-Forgotten-Process
## 🚀 Feature

Extent the Passwod-Forgotten-Process by
- evaluation of Email-Checked-Flag
- in case the Email-Check-Flag is false then 
  * add the Klicktipp-Registration in the Password-Forgotten-Process

=> This solves to avoid further new Accounts without a Klicktipp-Registration, because of not using the Registration-Confirmation-Email but the Passwort-Forgotten-Link

- create and add Log-Output for KlickTipp in both Processes
  * Registration
  * Password-Forgotten

Elweyn:
## 🍰 Pullrequest

We need to check the emailChecked of a User if forgot password is used
Log which user is not registered in Klicktipp at login

### Issues

- fixes #2105 

### Todo

- [X] Check user.emailChecked before changing emailOptInTypeId in checkOptInCode method
- [X] Log error in KlickTippMiddleware on each user login


backend/src/middleware/klicktippMiddleware.ts
```diff
@@ -29,7 +30,9 @@ export const klicktippNewsletterStateMiddleware: MiddlewareFn = async (
       if (klickTippUser) {
         klickTipp = new KlickTipp(klickTippUser)
       }
-    } catch (err) {}
+    } catch (err) {
+      logger.info(`Their is no user for (email='${result.email}') ${err}`)
```
Mogge, line 34:
```suggestion
      logger.info(`There is no user for (email='${result.email}') ${err}`)
```

Mogge: error?

Mogge (changes requested):
Some small typo

Mogge (approved):

Mogge (approved):

PR #2126, opened 2022-08-09
2125 feature gradido id: new column gradidoid in users table

clauspeterhuebner:
### Issues

- None

### Todo

- [X] new column gradidoID in users table
- [X] new column alias in users table
- [X] adapt entity DbUser to new columns
- [X] install package uuid in database and backend modul
- [X] new migration script to generate for each existing user-entry a valided unique UUID-V4
- [X] add in UserResolver the generation of a valided unique UUID-V4 during createUser invokation
- [X] add in UserResolverTest the additional columns and the UUID-Validation on correct version and validation
-


backend/src/graphql/resolver/UserResolver.ts
```diff
@@ -344,11 +358,13 @@ export class UserResolver {
...
 
     if (userFound) {
-      logger.info('User already exists with this email=' + email)
+      // ATTENTION: this logger-message will be exactly expected during tests
+      logger.info(`User already exists with this email=${email}`)
       // TODO: this is unsecure, but the current implementation of the login server. This way it can be queried if the user with given EMail is existent.
 
       const user = new User(communityDbUser)
       user.id = sodium.randombytes_random() % (2048 * 16) // TODO: for a better faking derive id from email so that it will be always the same id when the same email comes in?
+      user.gradidoID = '11111111-2222-4333-4444-55555555'
```
Elweyn, line 367: Not sure why we set a constant here, I thaught that we need something that can identify a user like on line 397

backend/src/util/communityUser.ts
```diff
@@ -6,6 +6,8 @@ import { User } from '@model/User'
 
 const communityDbUser: dbUser = {
   id: -1,
+  gradidoID: '11111111-2222-4333-4444-55555555',
```
Elweyn, line 9: Why this default value?

Elweyn (dismissed):
Seems good, only some questions for the default value of gradidoID!

backend/src/graphql/resolver/UserResolver.ts
clauspeterhuebner: this default constant value will only be set to satisfy the NOT NULL of the column gradidoID in the table users or in the typeorm class User. There is no identification of a user on the gradidoID at the moment. And if you take a look in the surrounding code of the UserResolver here, this code creates a faked user instance only to transfer the firstname, lastname and email for sending a email and to satisfy that a user instance has to be returned.

backend/src/util/communityUser.ts
clauspeterhuebner: there is no need here for an identical and unique gradidoID, it has only to satisfy that gradidoID is NOT NULL

backend/src/graphql/resolver/UserResolver.ts
Mogge: This code is to avoid email spying. When a user exists to a given email, we return a faked user object, simulating a successful user creation. Therefore `gradidoID = uuidv4()` should do it here. If we use the constant, you can identify the faked user and know, that the email exists

backend/src/util/communityUser.ts
Mogge: This one is fine

Mogge (changes requested):
Please ensure that create user returns a faked uuid

Tirokk (changes requested):
Cool next step 🍀 @clauspeterhuebner 

I tested it with

```bash
yarn dev_up
```

and

```bash
yarn dev_down
```

and all works fine for me.
I recommended a change with @Mogge .

Do we have the backend version to increment?

backend/src/graphql/resolver/UserResolver.ts
```diff
@@ -224,6 +225,19 @@ export const activationLink = (optInCode: LoginEmailOptIn): string => {
...
+  do {
+    gradidoId = uuidv4()
+    countIds = await DbUser.count({ where: { gradidoID: gradidoId } })
+    if (countIds > 0) {
+      logger.info('Gradido-ID creation conflict...')
+    }
+  } while (countIds > 0)
+  return gradidoId
+}
+
```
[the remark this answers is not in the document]
Mogge: I do not understand your point. How can `new User(dbUser)` create a new `gradidoID`? It is only the `dbUser` that may create a new `gradidoID`

Mogge:
> Do we have the backend version to increment?
@Tirokk We only have to increment the backend version when the `.env` changes. This is not the case here.

backend/src/graphql/resolver/UserResolver.ts
clauspeterhuebner: Yes that's right, to avoid email spying it is better to use uuidv4() instead of a constant. It is changed.

backend/src/graphql/resolver/UserResolver.ts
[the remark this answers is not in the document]
clauspeterhuebner: @Tirokk you are right, that between the initialization of a User with a newGradidoID and the database write access finished per commit, it will be possible that an other user could be created with the same gradidoID, because both users are still not committed till now.
But in the end the first user, which will be written and committed in database wins the game. The commit for the second user will run in an unique key constraint. It is undefined which of the two new users will be commited first.
So i would say, that we should start with the current implementation and collect experiences. There will follow further features with the gradidoID and then we could improve it, if necessary. Ok?

Mogge (approved):

PR #2171, opened 2022-09-01
fix: Remove Statistics from Wallet

Mogge:
## 🍰 Pullrequest
As it takes far to long to generate the statistics, for now it is no longer shown in the wallet.
Locales are only checked for English and German

### Issues

- None

### Todo

- [X] None


ogerly (approved):
i hope they are back in soon ... this is definitely an important part

PR #2180, opened 2022-09-06
Add is moderator to contribution message

Elweyn:
## 🍰 Pullrequest
To have a difference in the contribution messages between user message and admin message we add a is_moderator field to the contribution_messages table. 

### Issues

- None

### Todo

- [X] Create new migration that adds is_moderator to the database
- [X] Add new field isModerator to ContributionMessage entity
- [X] Add new isModerator Field to ContributionMessage graphql model
- [X] Set value on isModerator for the two calls adminCreateContributionMessage => true, createContributionMessage => false


Mogge (approved):
Tested and checked. Great work!

PR #2206, opened 2022-09-15
feat: Add Client Request Time to Context

### LINKED ISSUE #2199: 🚀 [Feature] Add Timestamp to Request Header in All Clients
## 🚀 Feature
Add a timestamp of now, including the timezone, in the request header of both clients (wallet and admin interface).
The purpose is to have this timestamp in the context object for the resolver.

Mogge:
## 🍰 Pullrequest
Client sends client request time in request header to server. Server reads client request time and adds it to resolver context as string

### Issues
- fixes #2199 

### Todo

- [X] None


backend/src/server/context.ts
```diff
@@ -18,13 +19,16 @@ export interface Context {
 
 const context = (args: ExpressContext): Context => {
   const authorization = args.req.headers.authorization
-  let token: string | null = null
+  const clientRequestTime = args.req.headers.clientrequesttime
```
clauspeterhuebner, line 22: clientrequesttime now all in lowercase is correct?

clauspeterhuebner (approved):

clauspeterhuebner (changes requested):
I merge this solution to #2225 and the context will only contain the clientRequestTime as soon as a token will be in the context too.
So it seems your solution will only add the clientRequestTime for token authenticated requests.

backend/src/server/context.ts
Mogge: Yes, it looks like the request headers are all lowercase. I tried it first in camel case

Mogge:
You are right, the time is only available if authenticated. I think this is enough for now, as you need authentication to do any time related operations.

clauspeterhuebner (approved):

PR #2210, opened 2022-09-15
change support mail

ogerly:
## 🍰 Pullrequest

the info page still contained an incorrect e-mail which was only a placeholder. 
the correct e-mail is now in it...


Mogge (approved):

Mogge (changes requested):
Please add a `SUPPORT_EMAIL` variable to the `.env.dist` and set a default value in `src/config/index.js` to the given Email. Update the `CONFIG_VERSION` in the frontend as well

Mogge:
Do not forget to set the `SUPPORT_EMAIL` in the `.env.template`!

Mogge (approved):

Issue #2225, opened 2022-09-20
🚀 [Feature] use FE-Time in BE to request user creations for the last 3 month


clauspeterhuebner:
## 🚀 Feature

- get frontend time from request context
- transfer it as parameter to the creations query method
- modify the query to use the give FE-time instead of the current BE time
- create some test-cases to ensure the timeshift between FE and BE

clauspeterhuebner:
corresponding PR is closed, so the ticket will be closed too

Issue #2202, opened 2022-09-15
🔧 [Refactor] Union Query in `util/creation.ts`


Mogge:
## 🔧 Refactor ticket
As we have a confirmed contribution for all creation transactions, we do not need a union any more. Write the query as typorm query.
Include pending means that we also read the contributions that are not confirmed yet. Without pending contributions confirmed at must be null.
Please write some extensive tests for this query. Create lots of contributions for the last three months, and see that errors are thrown if expected. Confirm some contributions and test, that everything behaves as expected.

PR #2229, opened 2022-09-21
2225 feature use client time in server to request user creations for the last 3 month

clauspeterhuebner:
### Issues

- None

### Todo

- [X] get clientRequestTime from context and if it not exists use current backend time
- [X] introduce the clientRequestTime as parameter to query the creations
- [X] adopt the query-changes from #2202
- [X] exchange the used date in the query with the clientRequestTime
- [X] several tests with clientRequestTime in the future against the backend time
- [X]  additional tests with clientRequestTime in the past against the backend time


backend/src/server/context.ts
```diff
@@ -38,4 +39,15 @@ export const getUser = (context: Context): dbUser => {
   throw new Error('No user given in context!')
 }
 
+export const getClientRequestTime = (context: Context): Date => {
+  // console.log('context:', context)
+  if (context.clientRequestTime) {
+    logger.info(`context with clientRequestTime=${context.clientRequestTime}`)
+    return new Date(context.clientRequestTime)
```
Mogge, line 46: This is not working. With the example of Tokyo.
```
conext.clientRequestTime = 2022-11-01T04:49:32.648+09:00
```
You get the server time as result:
```
2022-10-31T19:49:32.648Z
```

backend/src/seeds/users/bibi-bloxberg.ts
```diff
@@ -8,4 +8,5 @@ export const bibiBloxberg: UserInterface = {
   emailChecked: true,
   language: 'de',
   publisherId: 1234,
+  password: 'Aa12345_',
```
Mogge, line 11: Why? Is there a case we want another password in the seeded data?

backend/src/seeds/users/UserInterface.ts
```diff
@@ -9,4 +9,5 @@ export interface UserInterface {
   deletedAt?: Date
   publisherId?: number
   isAdmin?: boolean
+  password?: string
```
Mogge, line 12: Why? Do we want to have another password in seeded data?

backend/src/seeds/factory/contribution.ts
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  if (contribution.deleted) {
+    await mutate({
+      mutation: deleteContribution,
+      variables: {
+        id: pendingCreation.id,
+        amount: contribution.updatedAmount,
+        memo: contribution.updatedMemo,
+        creationDate: contribution.creationDate,
```
Mogge, line 72:
```suggestion
```

Mogge (changes requested):
I am sorry. This is not doing what it should. We have to simulate the date with input strings for `clientRequestTime` as sent by the client. See my comments to the code.

backend/src/server/context.ts
Mogge: We could just cut off the timezone offset and the test suite should pass again. But I fear we run in far more difficult problems, if we do so.

backend/test/helpers.ts
```diff
@@ -46,3 +47,15 @@ export const resetEntity = async (entity: any) => {
 export const resetToken = () => {
   context.token = ''
 }
+
+export const setClientRequestTime = (tm: Date) => {
```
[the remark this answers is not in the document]
clauspeterhuebner: in the context a string is written as it will be done by the client.
the signature with a date type for invokations from the test classes helps to use a flexible date without always converting it to a string before, because it will be done implicit

backend/src/seeds/contribution/updatedContribution101OneMonthBefore.ts
```diff
@@ -0,0 +1,18 @@
+import { ContributionInterface } from './ContributionInterface'
```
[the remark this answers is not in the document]
clauspeterhuebner: ou yes, that's right. i can remove it, but have to add it again for an update test case again...

backend/src/seeds/contribution/capturedContribution100OneMonthAgo.ts
```diff
@@ -0,0 +1,11 @@
+import { ContributionInterface } from './ContributionInterface'
+
+export const capturedContribution100OneMonthAgo: ContributionInterface = {
```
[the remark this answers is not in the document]
clauspeterhuebner: Äh...
amount = 100
OneMonthAgo = 1 month before current time

which name will be more expressive?

backend/src/server/context.ts
```diff
@@ -38,4 +39,15 @@ export const getUser = (context: Context): dbUser => {
...
+  }
+  // throw new Error('No clientRequestTime given in context!')
+  logger.info(`missing clientRequestTime, using BackendTime...`)
+  return new Date(Date.now())
```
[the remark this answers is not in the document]
clauspeterhuebner: that's right

backend/src/seeds/users/bibi-bloxberg.ts
clauspeterhuebner: the password is used in the contribution factory to login the user

backend/src/seeds/users/UserInterface.ts
clauspeterhuebner: the password is used in the contribution factory to login the user

backend/src/graphql/resolver/ContributionResolver.test.ts
```diff
@@ -160,11 +177,452 @@ describe('ContributionResolver', () => {
...
+      beforeAll(async () => {
+        await cleanDB()
+        resetToken()
+
+        await userFactory(testEnv, bibiBloxberg)
+        // create one contribution with 100GDD for bibi one month ago in the past
+        await contributionFactory(testEnv, bibiBloxberg, capturedContribution100OneMonthAgo)
+        // set clientRequestTime at the 1st day of the next month against the backend time
+        const clientRequestTime = new Date()
+        clientRequestTime.setDate(1)
```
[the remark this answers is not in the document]
clauspeterhuebner: solved
your example does not wrok because with setHours(4) your Date = 01.10.2022T04:00:00.000Z 
but with +09:00 you set the Timestring to 01.10.2022T04:00:00.000+09:00 which is in UTC 31.09.2022T19:00:00.000 and again the same month like the server

backend/src/server/context.ts
clauspeterhuebner: to process the testcases successfully the settings of the clientRequestTime have to fullfill some prerequisits:
- if the timezone offset > 0 the time have to be > the timezone offset to avoid switching the date back to the same server date in UTC
- if the timezone offset < 0 the time have to be < the timezone offset to avoid switching the date forward to the same server date in UTC

ulfgebhardt:
fixes #2225 ?

clauspeterhuebner:
The End-To-End Tests fail, because this PR will be blocked by the issue #2326 to adapt on the backend api changes.

clauspeterhuebner:
this PR will be closed, because of solution with #2393

clauspeterhuebner:
On my opinion we can close this PR, because Moritz created a better solution to solve the timezone topic.

This PR contains additional features like savings in database and graphQL API changes to deliver the structured object ContributionMonth with amount and month to the frontend.

The branch should NOT be deleted to keep the possiblility to have a look on code later...

PR #2232, opened 2022-09-22
refactor: 🍰 Login And Logout To Mutations

### LINKED ISSUE #2218: 🔧 [Refactor] Login and Logout are Mutations
## 🔧 Refactor ticket

So far we treated login and logout as queries. But now we write the login and logout events to the database. So login and logout must be handled as mutations from both sides, frontend and backend.

Elweyn:
## 🍰 Pullrequest

Since login and logout writes or will write events to the database, they are no queries anymore.
In this PR we changed them to mutations.

### Issues

- fixes #2218

### Todo

- [X] Backend change login to mutation
- [X] Backend change logout to mutation
- [X] Frontend change call to login 
- [X] Frontend change call to logout
- [X] Backend fix tests 
- [X] Frontend fix tests


Mogge (changes requested):
Looks very good. I looks like that the admin interface does not call the logout method. Please add the logout call if logged out from admin interface.

PR #2239, opened 2022-09-26
2238 feature configure git to keep linebreaks on lf even on windows

clauspeterhuebner:
### Issues

- None

### Todo

- [X] None

### DIFF
--- a/yarn_all.sh
+++ b/yarn_all.sh
@@ -1,0 +1,19 @@
+cd admin
+echo "admin yarn..."
+yarn
+cd ../backend
+echo "backen yarn..."
+yarn
+cd ../database
+echo "database yarn..."
+yarn
+cd ../frontend
+echo "frontend yarn..."
+yarn
+cd ../mariadb
+echo "mariadb yarn..."
+yarn
+cd ../nginx
+echo "nginx yarn..."
+yarn
+cd ..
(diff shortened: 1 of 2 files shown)


yarn_all.sh
```diff
@@ -0,0 +1,19 @@
...
+yarn
+cd ../database
+echo "database yarn..."
+yarn
+cd ../frontend
+echo "frontend yarn..."
+yarn
+cd ../mariadb
+echo "mariadb yarn..."
+yarn
```
Mogge, line 15: No package json in mariadb

yarn_all.sh
```diff
@@ -0,0 +1,19 @@
...
+cd ../nginx
+echo "nginx yarn..."
```
Mogge, line 17: No package json in nginx

Mogge (changes requested):
What do you need this script for?

yarn_all.sh
clauspeterhuebner: lines for mariadb deleted

yarn_all.sh
clauspeterhuebner: line for nginx deleted

clauspeterhuebner:
> What do you need this script for?

small helper script, because several times the application doesn't work based on missing or not updated yarn references. And after doing this several times manually i create this script.

PR #2255, opened 2022-10-07
fix(backend): devops backend test for adminresolvers confirm two creations one after is flaky

clauspeterhuebner:
### Issues

- None

### Todo

i had several ideas to solve this problem:
- [X] introduce milliseconds in datetime columns to ensure correct order even when writing in the same second
- [X] enable the database to lock single row instead of whole table during write access by defining indexes; 
       described here: https://stackoverflow.com/questions/2332768/how-to-avoid-mysql-deadlock-found-when-trying-to-get-lock-try-restarting-trans
- [X] surround the write access with a retry logic to solve deadlocks of concurrent invocations of the same method; 
       described here: https://blog.tekenlight.com/2019/02/21/database-deadlock-mysql.html

All three ideas are implemented, but not till the third one with the retry logic solves the problem


backend/src/graphql/resolver/AdminResolver.ts
```diff
@@ -478,99 +478,116 @@ export class AdminResolver {
...
-      let newBalance = new Decimal(0)
-      let decay: Decay | null = null
-      if (lastTransaction) {
-        decay = calculateDecay(
-          lastTransaction.balance,
-          lastTransaction.balanceDate,
-          receivedCallDate,
-        )
-        newBalance = decay.balance
+    let retryCount = 0
```
ulfgebhardt, line 481: I am against this implementation - its the task of the database to handle locks and retries - the database handles this when properly locked and resolves locks internally.

TypeORM offers the ability to lock a row or table when using the `QueryRunner`
See: https://stackoverflow.com/questions/69012855/typeorm-database-lock-please-explain-how-to-use-database-lock-in-typeorm-using

ulfgebhardt: `SQLSTATE[HY000]: General error: 1205 Lock wait timeout exceeded; try restarting transaction`

The above simply means the transaction has reached the innodb_lock_wait_timeout while waiting to obtain an exclusive lock which defaults to 50 seconds. 

Source: https://severalnines.com/blog/how-fix-lock-wait-timeout-exceeded-error-mysql/

This is what suppose to happen if we lock the tables right - and 50sec is plenty in my opinion. As said, the database handles this internally - to my knowledge - means it retries for 50sec to get a lock.

clauspeterhuebner: On my opinion your solution will only technically solve the deadlock by a simple retry of the blocked transaction on database level. 

But in our case the retry on database will write incorrect data, because the business logic of "confirmContribution" needs to know the last written and committed transaction including the current balance.

The test case starts two confirmContribution invocations, which both reads the same last gradidoTransaction with its balance. The first one wins and writes correct values. If the database will repeate internally the blocked transaction without rereading the previous commited data, the second transaction will write corrupt data. The second confirmContribution will create a gradidoTransaction with the same "last transaction id" than the previous commited gradidoTransaction.

If you want to avoid such situations proactive, you need other mechanism like optimistic lock or blocking monitors to synchronize concurrent processing of the same business logic.

clauspeterhuebner: In your link i found the following description:

The [InnoDB documentation](https://dev.mysql.com/doc/refman/8.0/en/innodb-parameters.html#sysvar_innodb_rollback_on_timeout) clearly says “InnoDB rolls back only the last statement on a transaction timeout by default”. In this case, we do not get the transaction atomicity offered by InnoDB. The atomicity in ACID compliant is either we get all or nothing of the transaction, which means partial transaction is merely unacceptable.

So if i understand this right, the innodb_timeout handling will corrupt the transaction atomicity!!!
This is FATAL and absolut not acceptable on my option.

clauspeterhuebner:
there is a better solution in #2458 
this PR will be closed

PR #2258, opened 2022-10-11
fix: 🍰 Disable Change Of Month For Update Contribution (wallet and admin)

Mogge:
## 🍰 Pullrequest
The tests of the admin resolver shows clearly, that changing the month of a existing contribution is not working as expected. Therefore it is disabled for users in the wallet as well as for admins

### Issues

- None

### Todo

- [X] None

### DIFF
--- a/backend/src/graphql/resolver/AdminResolver.ts
+++ b/backend/src/graphql/resolver/AdminResolver.ts
@@ -339,6 +339,9 @@ export class AdminResolver {
     let creations = await getUserCreation(user.id)
     if (contributionToUpdate.contributionDate.getMonth() === creationDateObj.getMonth()) {
       creations = updateCreations(creations, contributionToUpdate)
+    } else {
+      logger.error('Currently the month of the contribution cannot change.')
+      throw new Error('Currently the month of the contribution cannot change.')
     }
 
     // all possible cases not to be true are thrown in this function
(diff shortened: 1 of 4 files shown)


backend/src/graphql/resolver/AdminResolver.ts
```diff
@@ -339,6 +339,9 @@ export class AdminResolver {
     let creations = await getUserCreation(user.id)
     if (contributionToUpdate.contributionDate.getMonth() === creationDateObj.getMonth()) {
       creations = updateCreations(creations, contributionToUpdate)
+    } else {
+      logger.error('Currently the month of the contribution cannot change.')
+      throw new Error('Currently the month of the contribution cannot change.')
```
Tirokk, line 344: Currently this thrown error is not tested like in the `ContributionResolver.ts`.
If this should be then, please write a test.

If it is not neseccary in your eyes for the `AdminResolver.ts` then leave it like it is.

Tirokk (approved):
This works well @Mogge 👍🏼

Just one litte question. If you don't mind then skip it.

Therefor I approve 🚀💫

backend/src/graphql/resolver/AdminResolver.ts
Mogge: This is only a temporal solution. Therefore I do not want to test it

PR #2260, opened 2022-10-12
fix: 🐛 Prevent Loosing Redeem Code When Changing Between Register and Login in Auth Navbar

### LINKED ISSUE #2246: 🐛 [Bug] Redeem Code is Lost
## 🐛 Bugreport
 * have a valid transaction link
 * logout
 * use the transaction link
 * when on login page, click on register on the top
 * see that the redeem code is lost

The redeem code must not be lost when changing between register and login on the top buttons

mahula:
### Issues
- fixes #2246

### Todo
- [X] use computed links to cover links containing redeem codes
- [x] use the computed properties as mixins in components `AuthNavbar`, `AuthNavbarSmall`, and `Redeem LoggedOut`


frontend/src/components/LinkInformations/RedeemLoggedOut.vue
```diff
@@ -25,23 +25,17 @@
 </template>
 <script>
 import RedeemInformation from '@/components/LinkInformations/RedeemInformation.vue'
+import { authLinks } from '../../mixins/authLinks'
```
Mogge, line 28:
```suggestion
import { authLinks } from '@/mixins/authLinks'
```

Mogge (changes requested):
Looks very good. Great work with the mixin. Please avoid relative paths if possible

frontend/src/components/LinkInformations/RedeemLoggedOut.vue
mahula: Sure.
Fixed with https://github.com/gradido/gradido/commit/3439bdc67337f0041719a9a3636dec2a5a16d856.

Mogge (approved):

PR #2261, opened 2022-10-12
feat: 🍰 Global Jest Extension For Decimal Equal

Mogge:
## 🍰 Pullrequest
Add a global jest extension to the unit test setup for decimal equal

Usage:
`amount: expect.decimalEqual(200)`

### Issues

- None

### Todo

- [X] None


backend/test/extensions.ts
```diff
@@ -0,0 +1,33 @@
...
+        pass: true,
+      }
+    } else {
+      return {
+        message: () => `expected ${received} to equal ${value}`,
+        pass: false,
+      }
+    }
+  },
+})
```
Tirokk, line 20: Is this correct?

If it passes – because it equals – to say then `to not equal` and otherwise `to equal` … 🤔

backend/test/extensions.ts
```diff
@@ -0,0 +1,33 @@
...
+}
+
+declare global {
+  // eslint-disable-next-line @typescript-eslint/no-namespace
+  namespace jest {
+    interface Expect extends CustomMatchers {}
+    interface Matchers<R> extends CustomMatchers<R> {}
+    interface InverseAsymmetricMatchers extends CustomMatchers {}
+  }
+}
```
Tirokk, line 33: I didn't made an investigation yet and therefore I can not say if this is okay.
Shall I do or are you save?

Tirokk (changes requested):
Hey, very cool feature. 😍

Is this no standard in jest? 🤔

I have a question.

PS: I didn't merged the master here, so would be best do merge all the other PR's first to have my master fixes.

Tirokk (approved):
Very very cool 😍

I approve 🚀🚀💫💫

PR #2263, opened 2022-10-13
Feature: Test Transaction Resolver

### LINKED ISSUE #2250: 🚀 [Feature] Test Transaction Resolver
## 🚀 Feature
 * Setup up unit tests for the transaction resolver
 * Add the missing events
 * Log all errors
 * Test all possible errors
 * Test the events

jjimenezgarcia:
## 🍰 Pullrequest
Add Test Resolver tests

### Issues
- fixes #2250 

### Todo
- [X] Tests (without events)
- [X] Logger tests
- [X] Error tests
- [x] Events implementation
- [x] Events testing


backend/src/graphql/resolver/TransactionResolver.test.ts
```diff
@@ -0,0 +1,318 @@
...
+    }
+
+    user = await User.find({ relations: ['emailContact'] })
+  })
+
+  afterAll(async () => {
+    await cleanDB()
+  })
+
+  describe('wrong recipient', () => {
```
Mogge, line 65: I suggest `unknown recipient` here

backend/src/graphql/resolver/TransactionResolver.test.ts
```diff
@@ -0,0 +1,318 @@
...
+    it('unknown recipient', async () => {
```
Mogge, line 66: `throws an error`

backend/src/graphql/resolver/TransactionResolver.test.ts
```diff
@@ -0,0 +1,318 @@
...
+          },
+        }),
+      ).toEqual(
+        expect.objectContaining({
+          errors: [new GraphQLError('No user with this credentials')],
+        }),
+      )
+    })
+
+    it('deleted recipient', async () => {
```
Mogge, line 87: new describe block with this description. Then an it with `throws an error`

backend/src/graphql/resolver/TransactionResolver.test.ts
```diff
@@ -0,0 +1,318 @@
...
+      // delete bob
+      const bob = await User.findOneOrFail({ id: user[1].id })
```
Mogge, line 89: You could seed Stephen Hawking. This is a deleted user

backend/src/graphql/resolver/TransactionResolver.test.ts
```diff
@@ -0,0 +1,318 @@
...
+          errors: [new GraphQLError('The recipient account was deleted')],
+        }),
+      )
+
+      // make bob active again
+      bob.deletedAt = null
+      await bob.save()
+    })
+
+    it('recipient account not activated', async () => {
```
Mogge, line 118: New describe, then an it as above

backend/src/graphql/resolver/TransactionResolver.test.ts
```diff
@@ -0,0 +1,318 @@
...
+      resetToken()
+
+      await mutate({
```
Mogge, line 121: You can seed Garrick Ollivander. He has the email account not activated

backend/src/graphql/resolver/TransactionResolver.test.ts
```diff
@@ -0,0 +1,318 @@
...
+            mutation: sendCoins,
+            variables: {
+              email: 'peter@lustig.de',
+              amount: 100,
+              memo: 'testing',
+            },
+          }),
+        ).toEqual(
+          expect.objectContaining({
+            errors: [new GraphQLError(`user hasn't enough GDD or amount is < 0`)],
```
Mogge, line 244: What happens if we send an negative amount?. Can we have a test for this

backend/src/graphql/resolver/TransactionResolver.test.ts
```diff
@@ -0,0 +1,318 @@
...
+        expect(logger.error).toBeCalledWith(
+          `user hasn't enough GDD or amount is < 0 : balance=null`,
+        )
+      })
+    })
+  })
+
+  describe('transaction correct', () => {
+    it('sends the coins', async () => {
+      // make Peter Lustig Admin
```
Mogge, line 259: Peter Lustig is always admin

backend/src/graphql/resolver/TransactionResolver.test.ts
```diff
@@ -0,0 +1,318 @@
...
+
+      // login as bob again
+      await query({ query: login, variables: bobData })
+
+      expect(
+        await mutate({
+          mutation: sendCoins,
+          variables: {
+            email: 'peter@lustig.de',
+            amount: 100,
```
Mogge, line 287: This could be flaky. If the CI needs too much time, the decay applied could become relevant and the test fails because there are not enough GDD left. Take something like 50

backend/src/graphql/resolver/TransactionResolver.test.ts
```diff
@@ -0,0 +1,318 @@
...
+          },
+        }),
+      )
+    })
+
+    it('stores the send transaction event in the database', async () => {
+      expect(EventProtocol.find()).resolves.toContainEqual(
+        expect.objectContaining({
+          type: EventProtocolType.TRANSACTION_SEND,
+          userId: user[1].id,
```
Mogge, line 304: Test xUser, transactionID and amount as well

Mogge: Though I do not know how to get the transaction ID

backend/src/graphql/resolver/TransactionResolver.test.ts
```diff
@@ -0,0 +1,318 @@
...
+        }),
+      )
+    })
+
+    it('stores the receive event in the database', async () => {
+      expect(EventProtocol.find()).resolves.toContainEqual(
+        expect.objectContaining({
+          type: EventProtocolType.TRANSACTION_RECEIVE,
+          userId: user[0].id,
```
Mogge, line 313: xuser and amount as well

backend/src/graphql/resolver/TransactionResolver.ts
```diff
@@ -74,6 +76,7 @@ export const executeTransaction = async (
   )
   logger.debug(`calculated Balance=${sendBalance}`)
   if (!sendBalance) {
+    // josejgi: wrong messages from my point of view, at this point balance is always null, should log inside calculate balance or handle returns in a different way
```
Mogge, line 79: sendBalance is null if there is no transaction present (means that the balance of the user is 0), or the balance is less than the amount to send. I agree that the return value should be improved in the case of having not enough GDD

Mogge (changes requested):
Looks very good. Could be less code when seeding more users (see comments). Please add a test for sending a negative amount. Take a look at the line coverage of the backend unit tests in the CI. From 68%  it is up to 75%,

backend/src/graphql/resolver/TransactionResolver.test.ts
jjimenezgarcia: I agree with you, solved

backend/src/graphql/resolver/TransactionResolver.test.ts
jjimenezgarcia: hahahaha, nasty humor, didn't know that :p 
implemented like that now

backend/src/graphql/resolver/TransactionResolver.test.ts
jjimenezgarcia: you're right

backend/src/graphql/resolver/TransactionResolver.test.ts
jjimenezgarcia: love HP, 100% using that!

backend/src/graphql/resolver/TransactionResolver.test.ts
jjimenezgarcia: okay, thanks

backend/src/graphql/resolver/TransactionResolver.test.ts
jjimenezgarcia: validate.ts file is very messy for me, and clearly not working, cause a negative amount can be sent, you were right

jjimenezgarcia: solved now, still a mess that function... from my point of view it should be reviewed

backend/src/graphql/resolver/TransactionResolver.test.ts
jjimenezgarcia: okay, anyway the user had 1000GDD, but no problem on setting that to 50

backend/src/graphql/resolver/TransactionResolver.test.ts
jjimenezgarcia: solved the same way

backend/src/graphql/resolver/TransactionResolver.ts
jjimenezgarcia: and notice that using sendBalance inside the log message is useless, as it will be null,
but as mentioned many times, I think that checkBalance() is a mess

backend/src/graphql/resolver/TransactionResolver.test.ts
```diff
@@ -0,0 +1,348 @@
...
+          mutation: sendCoins,
+          variables: {
+            email: 'wrong@email.com',
+            amount: 100,
+            memo: 'test',
+          },
+        }),
+      ).toEqual(
+        expect.objectContaining({
+          errors: [new GraphQLError('No user with this credentials')],
```
Mogge, line 86: Test the logging

backend/src/graphql/resolver/TransactionResolver.test.ts
```diff
@@ -0,0 +1,348 @@
...
+            mutation: sendCoins,
+            variables: {
+              email: 'stephen@hawking.uk',
+              amount: 100,
+              memo: 'test',
+            },
+          }),
+        ).toEqual(
+          expect.objectContaining({
+            errors: [new GraphQLError('The recipient account was deleted')],
```
Mogge, line 109: Same

backend/src/graphql/resolver/TransactionResolver.test.ts
```diff
@@ -0,0 +1,348 @@
...
+            mutation: sendCoins,
+            variables: {
+              email: 'garrick@ollivander.com',
+              amount: 100,
+              memo: 'test',
+            },
+          }),
+        ).toEqual(
+          expect.objectContaining({
+            errors: [new GraphQLError('The recipient account is not activated')],
```
Mogge, line 133: Same

backend/src/graphql/resolver/TransactionResolver.test.ts
```diff
@@ -0,0 +1,348 @@
...
+        const transaction = await Transaction.find({
+          userId: user[0].id,
+          memo: 'unrepeateable memo',
+        })
+
+        expect(EventProtocol.find()).resolves.toContainEqual(
+          expect.objectContaining({
+            type: EventProtocolType.TRANSACTION_RECEIVE,
+            userId: user[0].id,
+            transactionId: transaction[0].id,
```
Mogge, line 341: How can this be the same id? There are two transactions in the DB. One with type receive and one with type send. This looks strange to me.

backend/src/util/validate.ts
```diff
@@ -26,6 +26,8 @@ async function calculateBalance(
 ): Promise<{ balance: Decimal; decay: Decay; lastTransactionId: number } | null> {
   const lastTransaction = await Transaction.findOne({ userId }, { order: { balanceDate: 'DESC' } })
   if (!lastTransaction) return null
+  // negative amount should not be allowed
+  if (amount.greaterThan(0)) return null
```
Mogge, line 30: Now I finally understand your point! We call `calculateBalance` twice. Once from the transaction resolver and once from the transaction link resolver. In both cases we multiply the amount by -1. We should do this operation in this function and the test here would be `amount <= 0`. Makes much more sense

backend/src/graphql/resolver/TransactionResolver.test.ts
```diff
@@ -0,0 +1,348 @@
...
+  beforeAll(async () => {
+    await userFactory(testEnv, peterLustig)
+    await userFactory(testEnv, bobBaumeister)
+    await userFactory(testEnv, stephenHawking)
+    await userFactory(testEnv, garrickOllivander)
+
+    bobData = {
+      email: 'bob@baumeister.de',
+      password: 'Aa12345_',
+      publisherId: 1234,
```
Mogge, line 53: Publisher Id is not needed. May be we should include the password in the user interface definition of the seed and add the password as well to the user objects. Then we do not need this helpers. We can pass the imported user objects directly to the login call. This makes this test scenarios much better to read. Would you mind doing this?

backend/src/util/validate.ts
Mogge: Go ahead. Get rid of the `mul(-1)` in the calls of `calculateBalance` and do it here instead.

Mogge (changes requested):
Very good code!

backend/src/graphql/resolver/TransactionResolver.test.ts
jjimenezgarcia: done

backend/src/graphql/resolver/TransactionResolver.test.ts
jjimenezgarcia: done

backend/src/graphql/resolver/TransactionResolver.test.ts
jjimenezgarcia: We have two users: `sender` and `receiver`
We create a transaction with that _unique memo_:

- **Sent transaction** - `user: sender` `xUser: receiver` `memo:  'unrepeateable memo'`
-  **Received transaction** - `user: receiver` `xUser: sender` `memo:  'unrepeateable memo'`

Now, if you search by:
`id = sender` && `memo = 'unrepeateable memo'` you will only get the sent transaction, as it is the one assigned to the sender

`id = receiver` && `memo = 'unrepeateable memo'` you will only get the received transaction, as it is the one assigned to the receiver

That is how in both cases `transaction[0]` is used, cause we are finding different transactions in each case

backend/src/graphql/resolver/TransactionResolver.test.ts
jjimenezgarcia: I agree that it is not needed.
However, are you sure User objects should store their password? I don't think that is the responsability of a User object. I mean, when we are aiming at an instantiated User, we are talking about a user interacting with the app, and so already logged. For me it is DBUser handling the password (which is already implemented).
I'm not sure of this, of course I can implement what you say, but let me know if I am wrong prior to that.
Thanks!!!

backend/src/util/validate.ts
jjimenezgarcia: I also added some utility to avoid all the mess regarding the toString() methods.
Take a quick look at it, you may like it or not hahaha

backend/src/graphql/resolver/TransactionResolver.ts
```diff
@@ -135,6 +146,20 @@ export const executeTransaction = async (
 
     await queryRunner.commitTransaction()
     logger.info(`commit Transaction successful...`)
+
+    const eventTransactionSend = new EventTransactionSend()
+    eventTransactionSend.userId = transactionSend.userId
+    eventTransactionSend.xUserId = transactionSend.linkedUserId
+    eventTransactionSend.transactionId = transactionSend.id
+    eventTransactionSend.amount = transactionSend.amount
```
Mogge, line 154: negative

Mogge (changes requested):
I am a bit scared of your last changes. We should first implement the unit tests, and test the behavior, including the transactions table itself. Then we can do the refactor, to ensure that the new code reproduces the old one 100%.

backend/src/graphql/resolver/TransactionResolver.ts
```diff
@@ -89,7 +83,7 @@ export const executeTransaction = async (
     transactionSend.memo = memo
     transactionSend.userId = sender.id
     transactionSend.linkedUserId = recipient.id
-    transactionSend.amount = amount.mul(-1)
+    transactionSend.amount = amount
```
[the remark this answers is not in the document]
jjimenezgarcia: You told me to get rid of the mul(-1) on the calls, and implement it as positive.

Anyways, calling it with mul(-1) is not very useful, as it cannot be called with a positive number, it is only used as a sender.

You tell me what to do next, but I changed it because you asked me to do so😅

backend/src/graphql/resolver/TransactionResolver.ts
jjimenezgarcia: same

jjimenezgarcia:
As you prefer, but notice that even there is "a lot" of modified code, I only modified de exception handling part, the rest is exactly the same, sharing de exact same functionality and using the exact same functions, didn't modify logic. Anyways,I have no problem on going back to an older commit.

Again, you tell me what you prefer.

backend/src/graphql/resolver/TransactionResolver.ts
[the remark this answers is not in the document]
Mogge: I thought, the we call `calculateBalance` with a positive amount. In our transaction table, the amount of the send transaction must remain negative. This ha nothing to do with the call of `calculateBalance`

backend/src/graphql/resolver/TransactionResolver.ts
Mogge: As the amount in the transaction table for the send event is negative I suggest to have a negative value here as well.

Mogge (changes requested):
The problem with the code is, that it changes the structure of the data stored in the database. So far, we have the send amount stored negative and the receive amount positive. We the change you made, you store the send amount as positive value. We have to keep this as ii is. It makes more sense for me as well

backend/src/graphql/resolver/TransactionResolver.ts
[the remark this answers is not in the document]
jjimenezgarcia: that is very true, sorry for the misstake
fixed now

backend/src/graphql/resolver/TransactionResolver.ts
jjimenezgarcia: as before, sorry for that, clearly my mistake, solved

jjimenezgarcia:
I fixed that, now I have only modified the calculateBalance() function, sorry again :S

Mogge (approved):

PR #2310, opened 2022-10-30
fix(frontend): submit contribution without gdd value

### LINKED ISSUE #2300: 🐛 [Bug] Submit Contribution - without GDD value
## 🐛 Bugreport

You can submit a GDD contribution even tho you did not enter a GDD amount.
- [ ] Make the GDD value a required field

image

ogerly:
## 🍰 Pullrequest
we have created our own unput components for email and password and have so far only used them on the login and register page. 

I have upgraded this with this PR. 
the email component is now also used for sending. 

there are also components for amount and memo textarea that can be used together in transactionForm and contributionForm. 

### Issues
- fixes #2300 

https://user-images.githubusercontent.com/1324583/201469985-0d6930c7-89ec-47e2-a42f-756502b7bb47.mp4


frontend/src/components/Contributions/ContributionForm.vue
```diff
@@ -50,15 +50,27 @@
...
-          id="contribution-amount"
-          v-model="form.amount"
-          type="text"
-          :formatter="numberFormat"
-        ></b-form-input>
-      </b-input-group>
+      <validation-provider
+        :name="$t('form.amount')"
+        :rules="{
+          required: true,
```
Mogge, line 56: Use the same validation mechanism as in `components/GddSend/TransactionForm.vue`
```
                :rules="{
                  required: true,
                  gddSendAmount: [0.01, balance],
                }"
```

Mogge (changes requested):
It would be nice to have the max available amount for the contribution to the selected month in the validation of the amount field.

frontend/src/components/Contributions/ContributionForm.vue
ogerly: added

frontend/src/components/Contributions/ContributionForm.vue
```diff
@@ -50,15 +50,28 @@
...
+          gddSendAmount: [0.01, balance],
```
Mogge, line 57: Balance is not defined. We need the max available amount for the selected month here. This should be available in the parent component and must be past as a property. Or you pass both max amounts for the two possible months to this component ...

Mogge (changes requested):

frontend/src/components/Contributions/ContributionForm.vue
ogerly: ok. done.
the maximum betag is calculated in the computed validMaxGDD. 
I would first leave the data in the ContributionForm and let it be calculated. 
since we are changing a lot graphically ... this will become a central method in the future.

frontend/src/components/Contributions/ContributionForm.vue
```diff
@@ -49,28 +58,40 @@
...
-        ></b-form-input>
-      </b-input-group>
-      <div
-        v-if="isThisMonth && parseInt(form.amount) > parseInt(maxGddThisMonth)"
-        class="text-danger text-right"
-      >
-        {{ $t('contribution.formText.maxGDDforMonth', { amount: maxGddThisMonth }) }}
+          :name="$t('form.amount')"
+          :label="$t('form.amount')"
+          :placeholder="'1'"
```
ulfgebhardt, line 66:
```suggestion
          :placeholder="'20'"
```

frontend/src/components/Contributions/ContributionForm.vue
```diff
@@ -49,28 +58,40 @@
...
+          :rules="{ required: true, gddSendAmount: [1, validMaxGDD] }"
+          typ="ContributionForm"
+        ></input-amount>
       </div>
-      <div
-        v-if="!isThisMonth && parseInt(form.amount) > parseInt(maxGddLastMonth)"
-        class="text-danger text-right"
+      <div v-else class="mb-5">{{ $t('contribution.exhausted') }}</div>
+      <!-- <validation-provider
```
ulfgebhardt, line 72: please remove comment

ulfgebhardt (changes requested):

frontend/src/components/Contributions/ContributionForm.vue
ulfgebhardt: Why are there two quotations?

Issue #1872, opened 2022-05-04
🔧 [Refactor] Properly Update Admin Interface


ulfgebhardt:
## 🔧 Refactor ticket

When confirming an creation in the admin interface or adding a new pending creation not all values are properly updated (e.g. the amount of available remaining creation GDD). This should be handled better in the vue application. The datamodel seems not be good enough at the given moment.

PR #2330, opened 2022-11-03
fix(admin): update contribution list after admin updates contribution

### LINKED ISSUE #2097: 🐛 [Bug] Update Admin Contribution Does not Update in List
## 🐛 Bugreport
 * Create an admin contribution
 * Edit the contribution memo text
 * Save
 
The Memo text does not update in the open contribution list.

Mogge:
## 🍰 Pullrequest

This not only fixes the bug, but also uses apollo to fetch data on creation as it should be done everywhere, including unit tests with mock apollo client

### Issues
- fixes #2097 
- relates #1872 

### Todo

- [X] None


admin/src/pages/CreationConfirm.vue
```diff
@@ -71,21 +72,6 @@ export default {
           this.toastError(error.message)
         })
     },
```
ulfgebhardt, line 74: whats with `confirmCreation`? Why can this mutation not live in the `apollo` field? Can we not annotate the corresponding button aswell?

ulfgebhardt (commented):
All good, but what about the mutation still in old style?

admin/src/pages/CreationConfirm.vue
Mogge: Docu says: Use `this.$apollo.mutate()` to send a GraphQL mutation.
See: https://apollo.vuejs.org/guide/apollo/mutations.html
I think we use it correctly

ulfgebhardt (approved):

PR #2363, opened 2022-11-10
feat(backend): 🍰 Mark creation via link

### LINKED ISSUE #2285: 🚀 [Feature] Mark Gradido Creation as `via link` when created via automatic link
## 🚀 Feature

Mark Gradido Creation as `via link` when created via automatic link.

The same way we mark transactions as `via link` we should discuss if we want such a marking for creations which were created via link.

Elweyn:
## 🍰 Pullrequest

If the creation is created via Link then shows a text to mark it

### Issues

- fixes #2285 

### Todo

- [X] None


backend/src/graphql/resolver/TransactionLinkResolver.ts
```diff
@@ -300,6 +300,7 @@ export class TransactionLinkResolver {
         transaction.balanceDate = now
         transaction.decay = decay ? decay.decay : new Decimal(0)
         transaction.decayStart = decay ? decay.start : null
+        transaction.transactionLinkId = contributionLink.id
```
ulfgebhardt, line 303:
```suggestion
        transaction.linkId = contributionLink.id
```

Please correct the field name to `linkId` - this is more clear then that the transaction type defines the context of the linkId (either transactionLinkId or contributionLinkid)

ulfgebhardt: @Mogge Please give us feedback on this!

ulfgebhardt (changes requested):

backend/src/graphql/resolver/TransactionLinkResolver.ts
ulfgebhardt: After talking to @Mogge the following:
You cannot write the `contributionLink.id` into the field `transaction.transactionLinkId`. You must join on the query  the transaction table with the contribution table. There is the field `contributionLinkId` available.

In the GraphQL Interface you can merge both fields to a `linkId` field and handle it like you did in the client.

Meaning: You can leave the client as it is except the naming of the field - you have to adjust the backend and join with contributions table.

Mogge (dismissed):
Looks very good! I have little changes

database/entity/0052-add_updated_at_to_contributions/Transaction.ts
```diff
@@ -0,0 +1,99 @@
...
+    name: 'transaction_link_id',
+    type: 'int',
+    unsigned: true,
+    nullable: true,
+    default: null,
+  })
+  transactionLinkId?: number | null
+
+  @OneToOne(() => Contribution, (contribution) => contribution.transaction)
+  @JoinColumn({ name: 'id', referencedColumnName: 'transactionId' })
```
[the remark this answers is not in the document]
Elweyn: No it does not work , problem is that the field transaction_id on the Contribution table.
With ``@JoinColumn({ name: 'id', referencedColumnName: 'transactionId' })`` we join with Transaction.id = Contribution.transactionId what we want.

backend/src/graphql/model/Transaction.ts
```diff
@@ -42,7 +42,9 @@ export class Transaction {
     this.creationDate = transaction.creationDate
     this.linkedUser = linkedUser
     this.linkedTransactionId = transaction.linkedTransactionId
-    this.transactionLinkId = transaction.transactionLinkId
+    this.transactionLinkId = transaction.contribution
```
ulfgebhardt, line 45: if you combine it please name it `linkId`. If you do not combine it have both fields

backend/src/graphql/resolver/TransactionLinkResolver.ts
```diff
@@ -265,6 +265,7 @@ export class TransactionLinkResolver {
         const creations = await getUserCreation(user.id, clientTimezoneOffset)
```
ulfgebhardt, line 265: revert changes on this file. You do not change things here.

database/entity/0052-add_updated_at_to_contributions/Contribution.ts
```diff
@@ -8,10 +8,12 @@ import {
   JoinColumn,
```
ulfgebhardt, line 8: change for `database/entity/contribution.ts` is missing

ulfgebhardt (approved):

PR #2372, opened 2022-11-10
[PoC] add graphql resolver and client for federation handshake

clauspeterhuebner:
### Issues

- None

### Todo

- [X] None

Issue #2368, opened 2022-11-10
🚀 [Feature] add graphql-resolver and -client for federation handshake


clauspeterhuebner:
## 🚀 Feature

- define federationresolver for handshake api
- add graphgl-client as new package for backend
- define federation-client requests for handshake api

PR #2339, opened 2022-11-04
feat(backend): add community federation tables

clauspeterhuebner:
### Issues

- None

### Todo

- [X] None

Issue #1887, opened 2022-05-10
🚀 [Feature] Federation 0.9: create database and typeORM model


clauspeterhuebner:
## 🚀 Feature

As preparation of Apollo and typeORM please take a look at
https://github.com/gradido/gradido/blob/1676-feature-federation-technical-concept/docu/Concepts/TechnicalRequirements/Federation.md#prerequisits-prerequisits

ulfgebhardt:
Can we close this @clauspeterhuebner ?

clauspeterhuebner:
As this ticket was a base part of the PoC and the federation/dht will be implemented on an other chain of tickets, this issue can be closed

PR #2374, opened 2022-11-10
feat(backend): add federation config properties

### LINKED ISSUE #2375: 🚀 [Feature] [Federation] Allow federation to persist its secret via Config
## 🚀 Feature

Allow federation to persist its secret via Config

- [x] Allow the federation (keypair) to be initialized with a seed
- [x] This seed must be configureable via config
- [x] If the value is not set, select a random seed (does the library do this if you provide `keyPair=null`?).

clauspeterhuebner:
### Issues

- fixes #2375 

### Todo

- [X] None


backend/.env.dist
```diff
@@ -63,5 +63,12 @@ EVENT_PROTOCOL_DISABLED=false
...
-# on an hash created from this tpoic
+# on an hash created from this topic
 # DHT_TOPIC=GRADIDO_HUB
+# Define the following federation properties to individualize this community instance
+# for federation per hyperswarm
+# FEDERATE_COMMUNITY_NAME=Gradido Community Development
+# FEDERATE_COMMUNITY_URL==http://localhost/
+# FEDERATE_COMMUNITY_PORT=4000
+# FEDERATE_COMMUNITY_UUID=14466b15-4631-4b1b-9110-dd6694356d9c
+# FEDERATE_KEY_SECRET=a51ef8ac7ef1abf162fb7a65261acd7a
```
ulfgebhardt, line 74: is this the secret key or the key seed?

while I agree on `FEDERATE` on the above, this one is part of the DHT hyperswarm right? So it should be `DHT_...`

backend/.env.template
```diff
@@ -56,5 +56,10 @@ WEBHOOK_ELOPAGE_SECRET=$WEBHOOK_ELOPAGE_SECRET
...
 EVENT_PROTOCOL_DISABLED=$EVENT_PROTOCOL_DISABLED
 
-# DHT
+# DHT/Federation
 DHT_TOPIC=$DHT_TOPIC
+FEDERATE_COMMUNITY_NAME=$FEDERATE_COMMUNITY_NAME
+FEDERATE_COMMUNITY_URL=$FEDERATE_COMMUNITY_URL
+FEDERATE_COMMUNITY_PORT=$FEDERATE_COMMUNITY_PORT
+FEDERATE_COMMUNITY_UUID=$FEDERATE_COMMUNITY_UUID
+FEDERATE_KEY_SECRET=$FEDERATE_KEY_SECRET
```
ulfgebhardt, line 65: you use none of this yet.
Consider to only introduce the secret/seed, nothing else to limit the scope of the PR. Changing the config again is cheap and a well defined process.

If you need the work you done in this PR for followup work, just merge this branch into your next to keep working on that state.

backend/.env.dist
clauspeterhuebner: it is the secret used as seed while creating a  key.

I would prever to rename the DHT_TOPIC to FEDERATE_DHT_TOPIC, because the hyperswarm is the first part of the whole federation process.

backend/.env.template
clauspeterhuebner: - Correct in this PR none of them is used,
- the next PR #2339 uses them and for working on, i merged the previous PR in branch of issue #1887
- the third PR #2372 uses then all changes from the previous two PRs and i merged both in the branch of issue #2368 for working on
- the problem will be during working on the 3rd PR to seperate changings belonging to the previous PRs. That means as soon as such uncritical changes can be merged in the master, as easier will be the separation efforts. And the amount of changes can be keept smaller as you all mentioned about my other big PRs ;-)

ulfgebhardt: We will not merge this PR, this is off scope and implements parts which are not used. Please stick to the negotiated scope: I even made a ticket to specify it correctly -> https://github.com/gradido/gradido/issues/2375

backend/.env.dist
ulfgebhardt: That would be ok with me.

But please name the key seed as such, its a better technical description

backend/.env.template
clauspeterhuebner: - I followed your requirements to create small PRs with less changed files.
- To go further the next steps will be done on PR #2339 and have to merge the changes from this PR
- To work on the the 3rd step #2372 a merge of all changes from the previous PRs is necessary
- Now my question:
  *  If the previous PRs will not be merged, how can then the amount of changes for the following PRs be kept low?
  * I do not understand what the ticket #2375 will correct against this PR?

clauspeterhuebner: Sorry i recognize, that i wrote this comment, but doesn't save it... ;-)

- Correct in this PR none of them is used,
- the next PR #2339 uses them and for working on, i merged the previous PR in branch of issue #1887
- the third PR #2372 uses then all changes from the previous two PRs and i merged both in the branch of issue #2368 for working on
- the problem will be during working on the 3rd PR to seperate changings belonging to the previous PRs. That means as soon as such uncritical changes can be merged in the master, as easier will be the separation efforts. And the amount of changes can be keept smaller as you all mentioned about my other big PRs ;-)

backend/.env.dist
clauspeterhuebner: Done

backend/.env.dist
```diff
@@ -61,7 +61,8 @@ EVENT_PROTOCOL_DISABLED=false
 # POSSIBLE VALUES: all | trace | debug | info | warn | error | fatal
 # LOG_LEVEL=info
 
-# DHT
-# if you set this value, the DHT hyperswarm will start to announce and listen
-# on an hash created from this tpoic
-# DHT_TOPIC=GRADIDO_HUB
+# Federation / DHT
```
ulfgebhardt, line 64:
```suggestion
# Federation
```

backend/.env.template
```diff
@@ -56,5 +56,6 @@ WEBHOOK_ELOPAGE_SECRET=$WEBHOOK_ELOPAGE_SECRET
 # EventProtocol
 EVENT_PROTOCOL_DISABLED=$EVENT_PROTOCOL_DISABLED
 
-# DHT
-DHT_TOPIC=$DHT_TOPIC
+# DHT/Federation
```
ulfgebhardt, line 59:
```suggestion
# Federation
```

backend/src/federation/index.ts
```diff
@@ -4,11 +4,24 @@
...
+const sodium = require('sodium-native')
 
 function between(min: number, max: number) {
   return Math.floor(Math.random() * (max - min + 1) + min)
 }
 
+const getSeed = (): Buffer | null => {
+  if (CONFIG.FEDERATE_DHT_SEED) {
+    const secret = CONFIG.FEDERATE_DHT_SEED
+    const seed = sodium.sodium_malloc(sodium.crypto_box_SEEDBYTES)
```
ulfgebhardt, line 18: what we need sodium for?
Just check the config for the required length, dont include sodium to have a definition of `32`.

You can just use `Buffer.alloc(CONFIG.FEDERATE_DHT_SEED, ' hex' )`

backend/src/federation/index.ts
```diff
@@ -28,7 +41,9 @@ export const startDHT = async (
   try {
     const TOPIC = DHT.hash(Buffer.from(topic))
 
-    const keyPair = DHT.keyPair()
+    const keyPair = DHT.keyPair(getSeed())
```
ulfgebhardt, line 44: I think you can omit the function and just write
```suggestion
    const keyPair = DHT.keyPair(Buffer.alloc(CONFIG.FEDERATE_DHT_SEED, ' hex' ))
```

You might require to check length of the secret tho - consider to use the config for this (?) or do the check before

```suggestion
    if(CONFIG.FEDERATE_DHT_SEED.length < DHT.minKeyLength){
      throw new Error()
    }
    const keyPair = DHT.keyPair(...)
```

backend/src/federation/index.ts
```diff
@@ -28,7 +41,9 @@ export const startDHT = async (
...
+    logger.info(`keyPairDHT: publicKey=${keyPair.publicKey.toString('hex')}`)
+    logger.info(`keyPairDHT: secretKey=${keyPair.secretKey.toString('hex')}`)
```
ulfgebhardt, line 46: do we want to log the secret key?

backend/src/index.ts
```diff
@@ -19,8 +19,10 @@ async function main() {
   })
 
   // start DHT hyperswarm when DHT_TOPIC is set in .env
-  if (CONFIG.DHT_TOPIC) {
-    await startDHT(CONFIG.DHT_TOPIC) // con,
+  if (CONFIG.FEDERATE_DHT_TOPIC) {
+    // eslint-disable-next-line no-console
+    console.log(`Federation/DHT active on ${CONFIG.FEDERATE_DHT_TOPIC}`)
```
ulfgebhardt, line 24:
```suggestion
    console.log(`Federation active on ${CONFIG.FEDERATE_DHT_TOPIC}`)
```

backend/src/federation/index.ts
ulfgebhardt: https://github.com/hyperswarm/dht/blob/4190b7505c365ef8a6ad607fc3862780c65eb482/test/connections.js#L554

This one allocs a 32 byte buffer and writes stuff into it - this might be the smartest way?

backend/src/federation/index.ts
clauspeterhuebner: Not necessary at all.
both keys are logged to see differences between with and without seed creation.
shift the secret-log to debug.level

backend/src/federation/index.ts
ulfgebhardt: https://github.com/hyperswarm/dht/blob/43b789fd4ed9f12f077350a411bf6cc5c79d6df7/lib/crypto.js#L10

The library does not expose any constant itself. It also just uses plaintext definition of 32 - even tho not ideal I would consider to also do this instead of including sodium

backend/.env.dist
```diff
@@ -61,7 +61,8 @@ EVENT_PROTOCOL_DISABLED=false
...
 
-# DHT
-# if you set this value, the DHT hyperswarm will start to announce and listen
-# on an hash created from this tpoic
-# DHT_TOPIC=GRADIDO_HUB
+# Federation
+# if you set the value of FEDERATE_DHT_TOPIC, the DHT hyperswarm will start to announce and listen
+# on an hash created from this topic
+# FEDERATE_DHT_TOPIC=GRADIDO_HUB
+# FEDERATE_DHT_SEED=64ebcb0e3ad547848fef4197c6e2332f
```
ulfgebhardt, line 68: I wondered in your PR already why you call those vars `FEDERATE_` instead of `FEDERATION_`. I think its valid to do so, but I expected the latter there - in the end chose as you please. I think its both valid.

backend/src/federation/index.ts
ulfgebhardt: if you use `hex` the key must be twice the length: https://www.geeksforgeeks.org/java-program-to-convert-hex-string-to-byte-array/

Two hex chars become 1 byte. 16^2  = 256
You can also choose not to use hex but just plain ascii

clauspeterhuebner: i will find a solution without sodium...

backend/src/federation/index.ts
clauspeterhuebner: found a solution to fix the buffer to the required size of 32 independent from input
i kept the getSeed(), because the Buffer.alloc doesn't handle a set or not set property to return null for the DHT.keyPair invokation.

backend/.env.dist
clauspeterhuebner: renamed to FEDERATION_xxx

backend/.env.dist
```diff
@@ -1,4 +1,4 @@
-CONFIG_VERSION=v11.2022-10-27
+CONFIG_VERSION=v12.2022-11-10
```
ulfgebhardt, line 1: do we miss an adjustment in `deployment/bare_metal/.env.dist` ?

backend/src/federation/index.ts
```diff
@@ -4,11 +4,22 @@
...
   return Math.floor(Math.random() * (max - min + 1) + min)
 }
 
+const getSeed = (): Buffer | null => {
+  if (CONFIG.FEDERATION_DHT_SEED) {
+    logger.debug(`with seed='${CONFIG.FEDERATION_DHT_SEED}'`)
+    return Buffer.alloc(KEY_SECRET_SEEDBYTES, CONFIG.FEDERATION_DHT_SEED)
+  }
+  return null
+}
```
ulfgebhardt, line 19: excluding the log this can be reduced to a one-liner. I agree to have this in a function with the null value as option.

```suggestion
const getSeed = (): Buffer | null => CONFIG.FEDERATION_DHT_SEED ? Buffer.alloc(KEY_SECRET_SEEDBYTES, CONFIG.FEDERATION_DHT_SEED) : null
```

backend/src/federation/index.ts
```diff
@@ -4,11 +4,22 @@
...
+
+const KEY_SECRET_SEEDBYTES = 32
```
ulfgebhardt, line 21: funny that this is working - but it must be defined before used - i think that would be more safe and appropriate. Move above the `getSeed()` function.

backend/src/index.ts
```diff
@@ -19,8 +19,10 @@ async function main() {
   })
 
   // start DHT hyperswarm when DHT_TOPIC is set in .env
-  if (CONFIG.DHT_TOPIC) {
-    await startDHT(CONFIG.DHT_TOPIC) // con,
+  if (CONFIG.FEDERATION_DHT_TOPIC) {
+    // eslint-disable-next-line no-console
+    console.log(`Federation active on ${CONFIG.FEDERATION_DHT_TOPIC}`)
```
ulfgebhardt, line 24: Here you can produce the seed if you think thats important - `starting Federation with seed ...`

backend/src/federation/index.ts
clauspeterhuebner: good solution, i've tried some expressions with such a one-line logic, but they doesn't work as expected...
now used your one!

backend/src/index.ts
clauspeterhuebner: This strange behaviour in the keypair with or without seed doesn't appear on using a normal Buffer instead of the sodium malloced one. Don't know why, but with the current solution both key looks quite good or as cryptical as expected.

backend/src/federation/index.ts
clauspeterhuebner: done

backend/src/index.ts
clauspeterhuebner: ...but i add the seed config-info in this log-output

backend/.env.dist
clauspeterhuebner: done

deployment/bare_metal/.env.dist
```diff
@@ -59,10 +59,11 @@ WEBHOOK_ELOPAGE_SECRET=secret
...
 EVENT_PROTOCOL_DISABLED=false
 
-## DHT
-## if you set this value, the DHT hyperswarm will start to announce and listen
-## on an hash created from this tpoic
-# DHT_TOPIC=GRADIDO_HUB
+# Federation
+# if you set the value of FEDERATION_DHT_TOPIC, the DHT hyperswarm will start to announce and listen
+# on an hash created from this topic
+FEDERATION_DHT_TOPIC=GRADIDO_HUB
```
ulfgebhardt, line 65: I think in the current state of the federation the default should be `disabled`
```suggestion
# FEDERATION_DHT_TOPIC=GRADIDO_HUB
```

ulfgebhardt (commented):
Think besides the `default = on` in the deployment config this is ready to merge.

ulfgebhardt (approved):

PR #2385, opened 2022-11-12
fix(backend): moved all jest & type-definition related packages into the `devDependencies` section

ulfgebhardt:
## 🍰 Pullrequest

moved all jest & type-definition related packages into the `devDependencies` section, since thats where they belong

### Issues

- None

### Todo

- [X] None


Mogge (commented):
Are you sure that we do not need type definitions for uuid and lodash clonedeep?

ulfgebhardt:
> Are you sure that we do not need type definitions for uuid and lodash clonedeep?

We need the types in dependencies only if we export such a type over an interface. I believe we do not do that. If you have reason to think differently please tell.

Ressource: https://stackoverflow.com/questions/45176661/how-do-i-decide-whether-types-goes-into-dependencies-or-devdependencies

Mogge (approved):
We will see if this causes problems ...

PR #2422, opened 2022-11-24
fix(backend): wrong balance after transaction receive

### LINKED ISSUE #2419: 🐛 [Bug] Account Balance incorrect after receiving links
## 🐛 Bugreport

Account Balance incorrect after receiving links

We have wrong balances for user accounts on production.

On first guess is that the redeemed link is subtracted instead of added to the balance. See the first screenshot. We miss aprox. 800 GDD on the balance and the user has received 400 GDD via link. Presumably the Amount was substracted not added resulting in a difference of 800.

image

image

- [ ] This might require the recalculation of all decay values if its not just a frontend / summation error. Make sure to check that out!

Mogge:
### Issues
- fixes #2419 

### Todo

- [ ] Fix production data (recalculation of decay and balance of transactions)


backend/src/graphql/resolver/TransactionResolver.test.ts
```diff
@@ -324,7 +326,7 @@ describe('send coins', () => {
             variables: {
               email: 'peter@lustig.de',
               amount: 50,
-              memo: 'unrepeatable memo',
+              memo: 'unrepeateable memo',
```
ulfgebhardt, line 329: spelling

PR #2424, opened 2022-11-25
fix(frontend): in contribution messages formular a message can be send twice, when clicking the submit button fast

### LINKED ISSUE #2342: 🐛 [Bug] In Contribution Messages Formular a message can be send twice, when clicking the submit button fast
## 🐛 Bugreport
When communication about a contribution, the users in the wallet frontend and in the admin (moderator) frontend can send messages twice by clicking fast:

Auswahl_018

- [x] Block the button after clicking till the request is completed

mahula:
## 🍰 Pullrequest
The submit button in the contribution messages formular has to be disabled while the submit request is happening in
- Wallet frontend
- Admin frontend

### Issues
- fixes #2342 2342


admin/src/components/ContributionMessages/ContributionMessagesFormular.vue
```diff
@@ -38,10 +38,12 @@ export default {
       form: {
         text: '',
       },
+      isSubmitting: false,
```
ulfgebhardt, line 41: do we not use `loading` normally? Please check with @Mogge if this is correct naming.

mahula: You are right, we normally use `loading`.
This is fixed by https://github.com/gradido/gradido/commit/403e68480c0cb7efa3f84544028398fb5a774d76.

ulfgebhardt (approved):

PR #2427, opened 2022-11-28
feat(backend): federation implement exchange of api versions persist in table

### LINKED ISSUE #2420: 🚀 [Feature] [Federation] Implement exchange of API Versions & persist in table
## 🚀 Feature

Implement exchange of API Versions & persist in table.

Use the DHT-Socket to transport API Versions & Endpoints to other clients in the network. Persist those information in a table with a last-seen timestamp.

- [ ] Send a list of supported API versions and there corresponding URL endpoints [over the socket](https://github.com/gradido/gradido/pull/2400/files#diff-371cd57fb51672f4f3de54a8f95b3e9717e68e2a32f66d4eef192097c33a28e1R153). The list should have the length of 1 and define the version `1.0` only for now, but it must be a list nevertheless.

Remarks:
- Define a new [`COMMUNITY_API_ENDPOINT_URL`](https://github.com/gradido/gradido/blob/master/backend/src/config/index.ts#L54) or `FEDERATION_API_BASE_URL` or something similar in the config.
- The url should be something like `http://localhost/api/1.0/` where the `http://localhost/api/` part is defined in the config, the version part is generated by the versions defined in the code.
- The version list definition can be an encapsulated thing like an enum you put into the federation index.ts - we need to reuse this later in other parts of the software.
- You can consider to write multiple messages to the socket. One for each api version. I consider [the way you currently implemented it](https://github.com/gradido/gradido/pull/2400/files#diff-371cd57fb51672f4f3de54a8f95b3e9717e68e2a32f66d4eef192097c33a28e1R153) as smart since you reduce complexity. Always send 1 version and 1 URL at a time, ut make sure we can send multiple of those messages.

---

- [ ] When receiving such a request from a client [over the socket](https://github.com/gradido/gradido/pull/2400/files#diff-371cd57fb51672f4f3de54a8f95b3e9717e68e2a32f66d4eef192097c33a28e1R58) persist the data into the database.

Remarks:
- You need to persist `pubKey`, `api_version`, `endpoint` and `last_seen`
- the table must have `pubKey` and `api_version` as primary
- use `REPLACE INTO`  or similar to make this happen in one DB statement (preferable with our ORM) [read here](https://thispointer.com/insert-into-a-mysql-table-or-update-if-exists/)
- consider to name the table `communities` (and the model `Community`)
- the standard fields `createdAt` (and `updatedAt` - not sure about this one - lets discuss) are considered worthwhile and can be included

clauspeterhuebner:
### Issues

- fixes #2420

### Todo

- [X] None


backend/src/federation/index.ts
```diff
@@ -36,18 +52,47 @@ export const startDHT = async (
...
+      socket.on('data', async (data: Buffer) => {
+        logger.info(`data: ${data.toString('ascii')}`)
+        const json = JSON.parse(data.toString('ascii'))
+
+        if (json.apiVersions && json.apiVersions.length > 0) {
+          const communities = new Array<DbCommunity>()
+
+          for (let i = 0; i < json.apiVersions.length; i++) {
+            const apiVersion = json.apiVersions[i]
+            let community = await DbCommunity.findOne({
```
ulfgebhardt, line 77: can we do this without reading first? just writing into database and overriding if needed?
Maybe our ORM will be overwhelmed with this, but SQL wise:
`REPLACE INTO communities VALUES (a, b, c)`
Is this possible when there is an entry in the database with a fictional field d which we do not define but don't want to override when writing?

backend/src/federation/index.ts
```diff
@@ -112,8 +157,7 @@ export const startDHT = async (
         socket.on('open', function () {
           // noiseSocket fully open with the other peer
           // console.log("writing to socket");
-          socket.write(Buffer.from(`${nodeRand}`))
-          socket.write(Buffer.from(JSON.stringify(nodeAPI)))
+          socket.write(Buffer.from(JSON.stringify(apiList)))
```
ulfgebhardt, line 160: consider:
`foreach api do send(json{api.v, api.u} + \n)`

database/entity/0055-add_communities_table/Community.ts
```diff
@@ -0,0 +1,25 @@
...
+  id: number
+
+  @Column({ name: 'public_key', type: 'binary', length: 64, default: null, nullable: true })
+  publicKey: Buffer
+
+  @Column({ name: 'api_version', length: 10, nullable: false })
+  apiVersion: string
+
+  @Column({ name: 'endpoint', length: 255, nullable: false })
+  endPoint: string
```
ulfgebhardt, line 15: either:

```suggestion
  @Column({ name: 'end_point', length: 255, nullable: false })
  endPoint: string
```

or:

```suggestion
  @Column({ name: 'endpoint', length: 255, nullable: false })
  endpoint: string
```

database/migrations/0055-add_communities_table.ts
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * This migration creates the `community` and 'communityfederation' tables in the `apollo` database (`gradido_community`).
+ */
+
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+export async function upgrade(queryFn: (query: string, values?: any[]) => Promise<Array<any>>) {
+  await queryFn(`
+    CREATE TABLE IF NOT EXISTS communities (
```
ulfgebhardt, line 11: I am for

```suggestion
    CREATE TABLE communities (
```

since this will fail if the table already exist - and we do expect it not to exist?!

database/migrations/0055-add_communities_table.ts
```diff
@@ -0,0 +1,28 @@
...
+      updated_at datetime(3),
+      PRIMARY KEY (id),
+      UNIQUE KEY public_api_key (public_key, api_version)
+    ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
+  `)
+}
+
+export async function downgrade(queryFn: (query: string, values?: any[]) => Promise<Array<any>>) {
+  // write downgrade logic as parameter of queryFn
+  await queryFn(`DROP TABLE IF EXISTS communities;`)
```
ulfgebhardt, line 27:
```suggestion
  await queryFn(`DROP TABLE communities;`)
```

We know it exists if this migration has run?!

backend/src/federation/index.ts
```diff
@@ -18,12 +15,31 @@ const POLLTIME = 20000
...
+
+type CommunityApi = {
+  api: string
+  url: string
+}
+type CommunityApiList = {
+  apiVersions: CommunityApi[]
+}
+
+const prepareCommunityApiList = (): CommunityApiList => {
```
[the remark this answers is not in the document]
clauspeterhuebner: The first try was to send each apiversion per own socket.write invocation. But this causes on the other side several asynchronous invocations of socket.on in parallel. If each socket.on invokes a write access to insert/update the community table, we have to synchronize them on our own.
If we send the apiversion per List, it is easier to recognize and filter on the expected data object "apiversions" and have nothing to synchronize on writing all api versions sequencially

backend/src/federation/index.ts
```diff
@@ -18,12 +15,31 @@ const POLLTIME = 20000
...
-  API_1_00: `${nodeURL}/api/1_00/`,
-  API_1_01: `${nodeURL}/api/1_01/`,
-  API_2_00: `${nodeURL}/graphql/2_00/`,
+const nodeURL = CONFIG.FEDERATION_COMMUNITY_URL || 'not configured'
+
+enum ApiVersionType {
+  V1 = 'v1',
+  V1_1 = 'v1_1',
+  V2 = 'v2',
+}
```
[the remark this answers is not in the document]
clauspeterhuebner: In the prototype implementation i tried to give the graphQL resolvers and their api a version information, like we discussed in the past.
A graphql-schema can't handle sub-schemata for resolvers and models for different endpoint.

At the end there will exist only one graphql-schema and per import i add a schema.ts out of the federation directory, which only contain the path definitions of the federation resolvers and model.
To insert a version information in the endpoint, which has an assoziation with the implementation i created in the federation directory the following structure:
backend/src/federation/graphql/v0
- arg
- model
- resolver
backend/src/federation/client/v0
- *Client.ts files

the resolver classes in v0/resolver has a prefix "v0_" at each public api method and each client in the v0 directory incokes a request by using the "v0_" prefix. 
Only such a prefix at the method-names were used by the graphql-schema compilation and leads to a versioning in the url.

That's the reason, why we should define the enum-values like the prefix, we will use for the method-names later.

backend/src/federation/index.ts
```diff
@@ -18,12 +15,31 @@ const POLLTIME = 20000
...
+
+type CommunityApi = {
+  api: string
+  url: string
+}
+type CommunityApiList = {
+  apiVersions: CommunityApi[]
+}
```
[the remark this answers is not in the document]
clauspeterhuebner: Not correct.
with this type the JSON string looks like:
{"apiVersions":[{"api":"v1","url":"http://localhost:4003/graphql"},{"api":"v1_1","url":"http://localhost:4003/graphql"},{"api":"v2","url":"http://localhost:4003/graphql"}]}

and without this type the JSON string looks like:
{[{"api":"v1","url":"http://localhost:4003/graphql"},{"api":"v1_1","url":"http://localhost:4003/graphql"},{"api":"v2","url":"http://localhost:4003/graphql"}]}

so it is easier to check only for "apiversions" as a own type than for an array of CommunityApi, which is not named in the JSON string

backend/src/federation/index.ts
clauspeterhuebner: I know about your requirement to use for the Insert/Update an optimized TypeORM solution.
i found these two interesting answers about this requirement:
- https://dev.to/rishit/optimizing-typeorm-tips-from-experience-part-1-dont-use-save-4ke9
- https://stackoverflow.com/questions/51583085/repository-save-uses-insert-instead-of-update

At the end the TypeORM needs a mechanism to check if the entity still exists to decide for insert or update.
This can on TypeORM level only be done by a leading read-access to have the correct entity-instances with their primary keys, After the attribut initializations calling save() the TypeORM can decide between insert and update base on the previous loaded entity isntances.
Please see the attached file for the SQL-Log for first INSERT and a follwong UPDATE per single save() invocation with a list of entities:
grafik

backend/src/federation/index.ts
```diff
@@ -18,12 +15,31 @@ const POLLTIME = 20000
...
+
+const prepareCommunityApiList = (): CommunityApiList => {
+  const apiEnumList = Object.values(ApiVersionType)
+  const communityApiArray = new Array<CommunityApi>()
+  apiEnumList.forEach((apiEnum) => {
```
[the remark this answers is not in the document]
clauspeterhuebner: Good hint! Many THX, so i learn more effient features of javascript...

backend/src/federation/index.ts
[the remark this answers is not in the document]
ulfgebhardt: Why would we need to syncronize them?
We can just override the existing data, no reading is needed from the table.
The database can solve the lock itself.

But I am fine with your solution if its easier.

backend/src/federation/index.ts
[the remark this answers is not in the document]
ulfgebhardt: See the next issue: https://github.com/gradido/gradido/issues/2428
We will be required to have our own apollo running for each endpoint/version, serving its own schema

backend/src/federation/index.ts
clauspeterhuebner: as mentioned above, sending each api with a separate socket.write will cause several asynchron and parallel socket.on invocations, which will then have to be synchronized for writing to the same database table.

So i prefere the solution writing the whole api-list per one socket.write invocation.

database/entity/0055-add_communities_table/Community.ts
clauspeterhuebner: if i have the choice, then i prefere the first suggestion... ;-)

database/migrations/0055-add_communities_table.ts
clauspeterhuebner: yes that's right. Is changed

database/migrations/0055-add_communities_table.ts
clauspeterhuebner: yes that's right. Is changed

backend/.env.dist
```diff
@@ -66,3 +66,4 @@ EVENT_PROTOCOL_DISABLED=false
 # on an hash created from this topic
 # FEDERATION_DHT_TOPIC=GRADIDO_HUB
 # FEDERATION_DHT_SEED=64ebcb0e3ad547848fef4197c6e2332f 
+# FEDERATION_COMMUNITY_URL=http://localhost:4000/graphql
```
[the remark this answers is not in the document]
clauspeterhuebner: yes i changed it to your suggested endpoint

backend/src/federation/index.ts
```diff
@@ -36,18 +52,47 @@ export const startDHT = async (
     logger.info(`keyPairDHT: publicKey=${keyPair.publicKey.toString('hex')}`)
     logger.debug(`keyPairDHT: secretKey=${keyPair.secretKey.toString('hex')}`)
 
+    const apiList = prepareCommunityApiList()
```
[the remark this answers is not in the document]
clauspeterhuebner: done

backend/src/federation/index.ts
```diff
@@ -18,12 +15,31 @@ const POLLTIME = 20000
 const SUCCESSTIME = 120000
 const ERRORTIME = 240000
 const ANNOUNCETIME = 30000
-const nodeRand = between(1, 99)
-const nodeURL = `https://test${nodeRand}.org`
-const nodeAPI = {
-  API_1_00: `${nodeURL}/api/1_00/`,
-  API_1_01: `${nodeURL}/api/1_01/`,
-  API_2_00: `${nodeURL}/graphql/2_00/`,
+const nodeURL = CONFIG.FEDERATION_COMMUNITY_URL || 'not configured'
```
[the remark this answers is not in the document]
clauspeterhuebner: done

backend/src/federation/index.ts
```diff
@@ -36,18 +52,47 @@ export const startDHT = async (
...
 
     server.on('connection', function (socket: any) {
       // noiseSocket is E2E between you and the other peer
       // pipe it somewhere like any duplex stream
-      logger.info(`Remote public key: ${socket.remotePublicKey.toString('hex')}`)
+      logger.info(`server on... with Remote public key: ${socket.remotePublicKey.toString('hex')}`)
       // console.log("Local public key", noiseSocket.publicKey.toString("hex")); // same as keyPair.publicKey
 
-      socket.on('data', (data: Buffer) => logger.info(`data: ${data.toString('ascii')}`))
+      socket.on('data', async (data: Buffer) => {
```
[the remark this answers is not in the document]
clauspeterhuebner: surrounded it with a try catch and logger.error output

backend/src/federation/index.ts
```diff
@@ -36,18 +52,47 @@ export const startDHT = async (
...
+        logger.info(`data: ${data.toString('ascii')}`)
+        const json = JSON.parse(data.toString('ascii'))
```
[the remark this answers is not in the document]
clauspeterhuebner: you are right to say it is the attack vector here on receiving data on socket.
That's the reason to define the acceptance of data on a dedicated type of data like it is done directly after parsing the data as a JSON object and check for the array of "apiVersions" containing the attributes "api" and "url"

Sending the list of apiVersions per separate socket.writes will cause asynchronous and parallel invocations of socket.on and we have to synchronize the write access on the same database table. For sure it could be done by the database itself, but on my opinion avoiding proactive will be a better solution than resolving it afterwards...

backend/src/federation/index.ts
```diff
@@ -36,18 +32,60 @@ export const startDHT = async (
...
-
+      socket.on('data', async (data: Buffer) => {
+        try {
+          logger.info(`data: ${data.toString('ascii')}`)
+          const json = JSON.parse(data.toString('ascii'))
+          if (
+            json.apiVersions &&
+            Object.prototype.toString.call(json.apiVersions) === '[object Array]' &&
+            json.apiVersions.length > 0 &&
+            Object.prototype.toString.call(json.apiVersions[0].api) === '[object String]' &&
```
ulfgebhardt, line 60: Why not use `typeof var`?
https://stackoverflow.com/questions/4059147/check-if-a-variable-is-a-string-in-javascript

backend/src/federation/index.ts
```diff
@@ -36,18 +32,60 @@ export const startDHT = async (
...
+            Object.prototype.toString.call(json.apiVersions[0].url) === '[object String]'
+          ) {
+            const communities = new Array<DbCommunity>()
+
+            for (let i = 0; i < json.apiVersions.length; i++) {
+              const apiVersion = json.apiVersions[i]
+              let community = await DbCommunity.findOne({
```
ulfgebhardt, line 67: What about this?
https://stackoverflow.com/questions/46745688/typeorm-upsert-create-if-not-exist

backend/src/federation/index.ts
```diff
@@ -36,18 +32,60 @@ export const startDHT = async (
...
+            }
+
+            await DbCommunity.save(communities)
+            logger.debug(`federation communities stored: ${JSON.stringify(communities)}`)
+          }
+        } catch (e) {
+          logger.error(`Error on receiving data from socket: ${JSON.stringify(e)}`)
+        }
+      })
       // process.stdin.pipe(noiseSocket).pipe(process.stdout);
```
ulfgebhardt, line 89: remove

backend/src/federation/index.ts
ulfgebhardt: The unsyncronized thing will happen with multiple communities anyway.
I do not believe thats a problem tho.

I agree that we keep it like this for now.

database/entity/0055-add_communities_table/Community.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  @Column({ name: 'last_announced_at', type: 'datetime', nullable: false })
+  lastAnnouncedAt: Date
+
+  @Column({ name: 'created_at', default: () => 'CURRENT_TIMESTAMP', nullable: false })
+  createdAt: Date
+
+  @Column({ name: 'updated_at', type: 'datetime', nullable: true, default: null })
+  updatedAt: Date | null
```
ulfgebhardt, line 24: use the proper typeorm constructs please: https://stackoverflow.com/questions/64941148/node-js-add-created-at-and-updated-at-in-entity-of-typeorm

backend/src/federation/index.ts
```diff
@@ -36,18 +32,60 @@ export const startDHT = async (
...
 
-      socket.on('data', (data: Buffer) => logger.info(`data: ${data.toString('ascii')}`))
-
+      socket.on('data', async (data: Buffer) => {
+        try {
+          logger.info(`data: ${data.toString('ascii')}`)
+          const json = JSON.parse(data.toString('ascii'))
+          if (
+            json.apiVersions &&
+            Object.prototype.toString.call(json.apiVersions) === '[object Array]' &&
```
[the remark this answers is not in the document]
clauspeterhuebner: Many THX, good to know.
I was slightly disapointed yesterday about the typescript/javascript typesafetiness and found my solution in different foren.
But yours is definitly nicer...
done

backend/src/federation/index.ts
clauspeterhuebner: From some answers out of the foren it was not clear what the typeof operator will return for a String: 'object' or as expected 'string' 
The Object.prototype.toString.call solution seems to be working correct and returns always [object string] for both 'string' and 'String'.
I've tried it now and changed it to the "typeof var" solution

backend/src/federation/index.ts
clauspeterhuebner: As i understand the answers in your link the upsert will not properly work for all databases and uses the .onConflict wich is marked as deprecated.
There is an alternative suggestion using the QueryBuilder with .onUpdate. This will create one Statement and work on mariadb with a SELECT-Statement afterwards for reading the changed entity again:
INSERT INTO `communities`
(
	`id`, 
	`public_key`, 
	`api_version`, 
	`end_point`, 
	`last_announced_at`, 
	`created_at`, 
	`updated_at`
)
VALUES (DEFAULT, ?, ?, ?, ?, DEFAULT, DEFAULT) 
ON DUPLICATE KEY UPDATE 
	`end_point` = VALUES(`end_point`), 
	`last_announced_at` = VALUES(`last_announced_at`) 
-- PARAMETERS: 
[
	"6c554f16417caa1ecc2099a2c3692553a8c0fa4d6ffdf4a4c319513181d174f0",
	"v1_0",
	"http://localhost:4003/api",
	"2022-11-30T20:15:03.894Z"
]

**But this .onUpdate solution is marked as deprecated too!**

So which solution will satisfy you:
- the **deprecated** one with an INSERT ON DUPLICATE- and a following SELECT-Statement
- or the current regular valid one of TypeORM-Entity with a leading SELECT- and an INSERT- or UPDATE-Statement

backend/src/federation/index.ts
clauspeterhuebner: done

database/entity/0055-add_communities_table/Community.ts
clauspeterhuebner: done

backend/src/federation/index.ts
```diff
@@ -36,19 +32,61 @@ export const startDHT = async (
     logger.info(`keyPairDHT: publicKey=${keyPair.publicKey.toString('hex')}`)
     logger.debug(`keyPairDHT: secretKey=${keyPair.secretKey.toString('hex')}`)
 
+    const apiList = {
+      apiVersions: Object.values(ApiVersionType).map(function (apiEnum) {
+        return { api: apiEnum, url: CONFIG.FEDERATION_COMMUNITY_URL }
+      }),
```
ulfgebhardt, line 38: why do you need the subfield apiVersions. I still do not understand this?
Why dont we use `apiList = []` the apiList to be an array?

```
const apiList = [
{ api: 'v1_0', url: 'https://localhost:4000/api/v1_0/' },
{ api: 'v1_1', url: 'https://localhost:4000/api/v1_1/' },
{ api: 'v2_0', url: 'https://localhost:4000/api/v2_0/' },
]
```

If I read [this](https://www.w3schools.com/JS/js_json_arrays.asp) correctly `'["Ford", "BMW", "Fiat"]'` is a valid json string,

backend/src/federation/index.ts
```diff
@@ -36,19 +32,61 @@ export const startDHT = async (
     logger.info(`keyPairDHT: publicKey=${keyPair.publicKey.toString('hex')}`)
     logger.debug(`keyPairDHT: secretKey=${keyPair.secretKey.toString('hex')}`)
 
+    const apiList = {
+      apiVersions: Object.values(ApiVersionType).map(function (apiEnum) {
+        return { api: apiEnum, url: CONFIG.FEDERATION_COMMUNITY_URL }
```
ulfgebhardt, line 37: Deliver the "correct" URL which should include the version number.
```suggestion
        return { api: apiEnum, url: CONFIG.FEDERATION_COMMUNITY_URL + apiEnum + '/' }
```

backend/src/federation/index.ts
```diff
@@ -36,19 +32,61 @@ export const startDHT = async (
...
+      socket.on('data', async (data: Buffer) => {
+        try {
+          logger.info(`data: ${data.toString('ascii')}`)
+          const json = JSON.parse(data.toString('ascii'))
+          if (
+            json.apiVersions &&
+            Array.isArray(json.apiVersions) &&
+            json.apiVersions.length > 0 &&
+            typeof json.apiVersions[0].api === 'string' &&
+            typeof json.apiVersions[0].url === 'string'
```
ulfgebhardt, line 61: this males no sense, put this check in the loop below

```suggestion
```

backend/src/federation/index.ts
```diff
@@ -36,19 +32,61 @@ export const startDHT = async (
...
-
-      // process.stdin.pipe(noiseSocket).pipe(process.stdout);
+      socket.on('data', async (data: Buffer) => {
+        try {
+          logger.info(`data: ${data.toString('ascii')}`)
+          const json = JSON.parse(data.toString('ascii'))
+          if (
+            json.apiVersions &&
+            Array.isArray(json.apiVersions) &&
+            json.apiVersions.length > 0 &&
```
ulfgebhardt, line 59: This check can be reduced to `isArray(json)` if no senseless field `apiVersions` is submitted.
The array length check is not needed at all - if the length is zero the for loop is executed zero times.

backend/src/federation/index.ts
```diff
@@ -112,8 +150,7 @@ export const startDHT = async (
         socket.on('open', function () {
           // noiseSocket fully open with the other peer
           // console.log("writing to socket");
```
ulfgebhardt, line 152: remove comment

ulfgebhardt (commented):
I think you are very close!

backend/src/federation/index.ts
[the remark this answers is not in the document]
clauspeterhuebner: No it does not work, even setting the '...=> {api: apiEnum,...' is not resolvable for compiler.

clauspeterhuebner: As in a previous version i created the types

type CommunityApi = {
  api: string
  url: string
}
type CommunityApiList = {
  apiVersions: CommunityApi[]
}
to have structured objects. To benefit of the CommunityApi-Type with only one attribute of type CommunityApi[] base on an easier check of the parsed JSON-string after received data. Without the CommunityApiList-Type you have to check on an array of ComunityApi-Type entries with several attributes.

I reintroduced the types back, because on my opinion working with more type-safetiness in the code increases the quality more than focus on simply reducing the amount of lines of code.

backend/src/federation/index.ts
clauspeterhuebner: the concatination for a complete usable url should not be done at receiving time, instead it will be done in the client class at the point of time where it is realy used for sending a request.

backend/src/federation/index.ts
clauspeterhuebner: removed and unnecessary by reintroducing the type definitions, see above

backend/src/federation/index.ts
clauspeterhuebner: removed and unnecessary by reintroducing the type definitions, see above

backend/src/federation/index.ts
clauspeterhuebner: done

backend/src/federation/index.ts
```diff
@@ -18,37 +13,141 @@ const POLLTIME = 20000
...
     logger.debug(`keyPairDHT: secretKey=${keyPair.secretKey.toString('hex')}`)
 
+    const ownApiVersions = Object.values(ApiVersionType).map(function (apiEnum) {
+      const comApi: CommunityApi = {
+        api: apiEnum,
+        url: CONFIG.FEDERATION_COMMUNITY_URL
+          ? (CONFIG.FEDERATION_COMMUNITY_URL.endsWith('/')
+              ? CONFIG.FEDERATION_COMMUNITY_URL
+              : CONFIG.FEDERATION_COMMUNITY_URL + '/') + apiEnum
+          : 'not configured',
```
ulfgebhardt, line 80: this part should be done in the config.
`not configured` is not a good value to be transmitted.

backend/.env.template
```diff
@@ -59,3 +59,5 @@ EVENT_PROTOCOL_DISABLED=$EVENT_PROTOCOL_DISABLED
 # Federation
 FEDERATION_DHT_TOPIC=$FEDERATION_DHT_TOPIC
 FEDERATION_DHT_SEED=$FEDERATION_DHT_SEED
+FEDERATION_DHT_TEST_SOCKET=$FEDERATION_DHT_TEST_SOCKET
```
ulfgebhardt, line 62: Why do this on the same branch when its clearly out of scope from the original ticket?

clauspeterhuebner: without a test infrastructure of the DHT hyperswarm and socket communication, on my opinion a valid interims solution to test different data constellations by configuring one instance of the swarm in DHT_TEST_SOCKET-Mode

ulfgebhardt (commented):
@clauspeterhuebner how are things? Will you remove the testsocket - do you think its sufficiently covered with the unit tests for now? Whats the exact plan?

ulfgebhardt (approved):

Issue #2436, opened 2022-11-29
🐛 [Bug] Data inconsistency - user & user_contacts


ulfgebhardt:
## 🐛 Bugreport

Data inconsistency - user & user_contacts

- There is live users with deleted user contacts (when restoring users)
- There is live user contacts with deleted users (on production data 2 occurrences)

`SELECT email FROM users LEFT JOIN user_contacts ON users.email_id = user_contacts.id WHERE users.deleted_at IS NOT NULL AND  user_contacts.deleted_at IS NULL;`

`SELECT email FROM users LEFT JOIN user_contacts ON users.email_id = user_contacts.id WHERE users.deleted_at IS NULL AND  user_contacts.deleted_at IS NOT NULL;`

- [ ] Make sure to fix delete & restore logic for users to include the user contacts
- [ ] Research if we can define an automatisation with typeorm (when marking as deleted on this model also mark as deleted on connected other model)
- [ ] Write migration to ensure consistency of data (undelete user_contacts where needed, delete user_contacts where needed)

PR #2444, opened 2022-11-30
fix(backend): delete / undelete email contact as well

Mogge:
## 🍰 Pullrequest

soft delete / restore email contact when deleting / undeleting an user

### Issues
- relates #2436 

### Todo

- [X] None

### DIFF
--- a/backend/src/graphql/resolver/AdminResolver.test.ts
+++ b/backend/src/graphql/resolver/AdminResolver.test.ts
@@ -366,6 +366,19 @@ describe('AdminResolver', () => {
             expect(new Date(result.data.deleteUser)).toEqual(expect.any(Date))
           })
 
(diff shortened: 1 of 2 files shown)


backend/src/graphql/resolver/AdminResolver.test.ts
```diff
@@ -489,6 +502,15 @@ describe('AdminResolver', () => {
                 }),
               )
             })
+
+            it('has deleted at set to null in users and user contacts', async () => {
```
ulfgebhardt, line 506:
```suggestion
            it('has deleted_at set to null in users and user contacts', async () => {
```

ulfgebhardt (changes requested):

ulfgebhardt (approved):

PR #2451, opened 2022-12-01
fix(database): consistent deleted at bewteen users and user contacts

### LINKED ISSUE #2436: 🐛 [Bug] Data inconsistency - user & user_contacts

Mogge:
### Issues
- fixes #2436

### Todo

- [X] None


database/migrations/0055-consistent_deleted_users.ts
```diff
@@ -0,0 +1,26 @@
...
+    AND user_id IN (SELECT id FROM users WHERE deleted_at IS NOT NULL);`)
+
+  for (let i = 0; i < contactsToFix.length; i++) {
+    const deletedAt = new Date(contactsToFix[i].deleted_at)
+      .toISOString()
+      .slice(0, 19)
+      .replace('T', ' ')
+
+    await queryFn(`
+      UPDATE user_contacts SET deleted_at = '${deletedAt}' WHERE id = ${contactsToFix[i].id};`)
```
ulfgebhardt, line 20: `UPDATE user_contacts SET uc.deletedAt = u.deletedAt WHERE ....

Also remove the for`

PR #2452, opened 2022-12-01
fix(backend): run all timers for high values

Mogge:
## 🍰 Pullrequest
It looks like `setTimeout` hast problems with higher values in jest. Therefore run two timers.

### Issues

- None

### Todo

- [X] None

### DIFF
--- a/backend/src/graphql/resolver/util/creations.test.ts
+++ b/backend/src/graphql/resolver/util/creations.test.ts
@@ -170,8 +170,11 @@ describe('util/creation', () => {
         const targetDate = new Date(now.getFullYear(), now.getMonth() + 1, 0, 23, 0, 0)
 
         beforeAll(() => {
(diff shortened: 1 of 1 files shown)


Elweyn (approved):
Works on my local machin

backend/src/graphql/resolver/util/creations.test.ts
```diff
@@ -225,8 +228,10 @@ describe('util/creation', () => {
           })
 
           it('has the clock set correctly', () => {
+            const targetMonth = nextMonthTargetDate.getMonth() + 1
+            const targetMonthString = (targetMonth < 10 ? '0' : '') + String(targetMonth)
```
ulfgebhardt, line 232: check if this is a relevant thing to avoid this ternary https://stackoverflow.com/questions/2998784/how-to-output-numbers-with-leading-zeros-in-javascript

ulfgebhardt (approved):

PR #2458, opened 2022-12-06
feat(backend): semaphore to lock transaction table

ulfgebhardt:
## 🍰 Pullrequest

semaphore to lock transaction table

### Issues

- None

### Todo

- [x] build test cases to prove its working


backend/src/util/validate.ts
```diff
@@ -24,7 +24,7 @@ async function calculateBalance(
   time: Date,
   transactionLink?: dbTransactionLink | null,
 ): Promise<{ balance: Decimal; decay: Decay; lastTransactionId: number } | null> {
-  const lastTransaction = await Transaction.findOne({ userId }, { order: { balanceDate: 'DESC' } })
+  const lastTransaction = await Transaction.findOne({ userId }, { order: { id: 'DESC' } })
```
ulfgebhardt, line 27: @Mogge please check this - is this valid?

Mogge: OK

backend/src/graphql/resolver/TransactionLinkResolver.ts
```diff
@@ -309,6 +312,7 @@ export class TransactionLinkResolver {
         throw new Error(`Creation from contribution link was not successful. ${e}`)
       } finally {
         await queryRunner.release()
+        releaseLock()
       }
       return true
```
Mogge, line 317: Is this return statement in the right place? In the transaction resolver it is inside the try block

ulfgebhardt: that doesn't matter in my opinion. It was unclear to me if a `finally` block is executed after a `return` statement. This is the case.
In this particular part of the code its the other case - the finally block is definitely executed, because the return comes after it.

My opinion is that this code (return after finally) is better for readability and therefore to be preferred over the other variant.
I would like to keep it as it is, since its no impact on how things are executed, but has the least code impact.

backend/src/graphql/resolver/ContributionResolver.ts
```diff
@@ -639,10 +642,10 @@ export class ContributionResolver {
       })
     } catch (e) {
       await queryRunner.rollbackTransaction()
-      logger.error(`Creation was not successful: ${e}`)
```
Mogge, line 642: Why remove this?
Better: `logger.error('Creation was not successful:', e)`

Mogge (dismissed):
Please restore the error log

ulfgebhardt:
> Please restore the error log

Done

backend/src/graphql/resolver/TransactionLinkResolver.test.ts
```diff
@@ -23,6 +23,11 @@ import { User } from '@entity/User'
 import { UnconfirmedContribution } from '@model/UnconfirmedContribution'
 import Decimal from 'decimal.js-light'
 import { GraphQLError } from 'graphql'
+import { TRANSACTIONS_LOCK } from '@/util/TRANSACTIONS_LOCK'
+
+// mock semaphore to allow use fake timers
+jest.mock('@/util/TRANSACTIONS_LOCK')
+TRANSACTIONS_LOCK.acquire = jest.fn().mockResolvedValue(jest.fn())
```
clauspeterhuebner, line 30: Why is the TRANSACTION_LOCK mocked here, when it is never used in this file?
Or i can't see or understand the usage of it in line 193 with the statement:

setTimeout(jest.fn(), 1000 * 60 * 60 * 24)

clauspeterhuebner (approved):
good work!
In the ContributionResolver i'am not sure if we need to set the lock even before getting and validating the users creations... but if it works, it's fine

backend/src/graphql/resolver/TransactionLinkResolver.test.ts
ulfgebhardt: The problem is that [here](https://github.com/gradido/gradido/pull/2458/files/7023fdba2a43826c2d8a4c9a2642ab64b38acde4#diff-d177c02b13a87d9ee1ffa3a6a55fc6412101c3de894bcd480bca3e493eda47d1L171) `redeemTransaction` is called, which results in a lock acquire. We want to mock this in the test environment unless we wat to test the semaphore itself - or thats how I understood @Mogge

ulfgebhardt:
> good work! In the ContributionResolver i'am not sure if we need to set the lock even before getting and validating the users creations... but if it works, it's fine

I think this is a valid argument - the way its implemented there could be two concurrent confirm actions. But since this is an admin thing, I rather move this problem into the future ;)

PR #2460, opened 2022-12-12
refactor(backend): seed contributions as user

Mogge:
## 🍰 Pullrequest
To get rid of flaky test, get the contribution id as mutation result and not by database query after mutation. Uses `createContribution` instad of `adminCreateContribution` to create the contributions.

### Issues

- None

### Todo

- [X] None

### DIFF
--- a/.github/workflows/lint_pr.yml
+++ b/.github/workflows/lint_pr.yml
@@ -29,6 +29,9 @@ jobs:
             admin
             database
             release
+            federation
+            workflow
+            docker
             other
           # Configure that a scope must always be provided.
           requireScope: true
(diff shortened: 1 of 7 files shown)


ulfgebhardt:
@Mogge I believe the test is still flaky -> at least the failing one. I don't think this is related to anything I did - the test itself has no prerequisites but fails because it cannot be guaranteed which of the two calls will be successful - it should be either one, but which ones is unsure.

Please have a close look here and tell me if its AdminResolver related.

Tirokk (approved):
Looks good to me and works, as far I see. 🚀💫

PR #2465, opened 2022-12-14
feat(backend): setup unit tests for federation

Mogge:
## 🍰 Pullrequest
Setup unit tests for federation.
Events can be triggered as shown. Different event arguments can be applied to simulate different cases. 

### Issues

- None

### Todo

- [X] None

### DIFF
--- a/backend/src/federation/index.test.ts
+++ b/backend/src/federation/index.test.ts
@@ -1,0 +1,162 @@
+import { startDHT } from './index'
+import DHT from '@hyperswarm/dht'
+import CONFIG from '@/config'
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import { logger } from '@test/testSetup'
+
+CONFIG.FEDERATION_DHT_SEED = '64ebcb0e3ad547848fef4197c6e2332f'
+
+jest.mock('@hyperswarm/dht')
+jest.useFakeTimers()
+
+const TEST_TOPIC = 'gradido_test_topic'
+
+const keyPairMock = {
+  publicKey: Buffer.from('publicKey'),
+  secretKey: Buffer.from('secretKey'),
+}
+
+const serverListenSpy = jest.fn()
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const serverEventMocks: { [key: string]: any } = {}
+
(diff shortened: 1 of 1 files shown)


backend/src/federation/index.test.ts
```diff
@@ -0,0 +1,162 @@
...
+          describe('on data', () => {
+            beforeEach(() => {
+              socketEventMocks.data(Buffer.from('some-data'))
+            })
+
+            it('can be triggered', () => {
+              expect(true).toBe(true)
+            })
+          })
+        })
```
ulfgebhardt, line 158: I think this is the crucial part where you recieve data and we handle it. We would have several test scenarios where we recieve valid or invalid data - this should be the main test case.

Will this work like above?

`expect(socketOnMocks[data]).toBeCalledWith(Buffer.from('some-data'))`

backend/src/federation/index.test.ts
```diff
@@ -0,0 +1,162 @@
...
+
+        describe('socket events', () => {
+          describe('on data', () => {
+            beforeEach(() => {
+              socketEventMocks.data(Buffer.from('some-data'))
+            })
+
+            it('can be triggered', () => {
+              expect(true).toBe(true)
```
ulfgebhardt, line 155: this would be `expect(theDHT).toThrowSomeError("invalid format")`, correct?
The call above will trigger our code?!

Mogge: The call triggers the code. I did this to see, that there are no errors bu the test implementation. The code executed should be tested as needed. E.g. if the on data event returns proper data, a new entry should be seen in the database.

Mogge: E.g.
```
          describe('on data', () => {
            beforeEach(() => {
              socketEventMocks.data('invalid data type')
            })

            it('throws an error', () => {
              expect(logger.error).toBeCalledWith('Error on receiving data from socket: ...')
            })
          })
```

clauspeterhuebner:
It looks quite good for me as far as i can evaluate these mocking- and test-framework things...

But i miss the data receive tests, which are part of the current index.ts implementation with active TestMode.
These implicite tests should on Ulf's opinion NOT part of the implementation and be shifted in unit tests before he will accept the PR.

PR #2468, opened 2022-12-15
feat(backend): test semaphore

Mogge:
## 🍰 Pullrequest
Create a lot of traffic on the transaction table at the same time, see that there are no errors.

### Issues

- None

### Todo

- [X] None


ulfgebhardt:
I "fixed" it. I think the remaining errors show that the semaphore is working properly, but that the time we calculate `now()` must be inside the semaphore lock - thats just a guess for now tho.

All remaining failing tests are caused by `from < to` errors. It seems to affect `confirmContribution` and `redeemTransactionLink`

PR #2471, opened 2022-12-16
fix(admin): reduce triggers of success toast on deleted user form to exactly one

### LINKED ISSUE #2435: 🐛 [Bug] When undeleting an user - two success toasts
## 🐛 Bugreport

When undeleting an user - two success toasts

image

- [x] Why?
- [x] Change this to one toast only

mahula:
## 🍰 Pullrequest
In the admin frontend's user search, when undeleting a user the succes toast was triggered twice

- in the compoment `UndeletedUserFormular`
  Here the success toast was inconsistently triggered only for undeletion, but not for deletion of a user, so it has been removed here to fix the issue.
- on the page `UserSearch`
  Here the success toast is triggered more globally for admin actions like user deletion/undeletion in the user entry.

### Issues
- fixes #2435 2435

### DIFF
--- a/admin/src/components/DeletedUserFormular.vue
+++ b/admin/src/components/DeletedUserFormular.vue
@@ -65,7 +65,6 @@ export default {
           },
         })
         .then((result) => {
-          this.toastSuccess(this.$t('user_recovered'))
           this.$emit('updateDeletedAt', {
             userId: this.item.userId,
             deletedAt: result.data.unDeleteUser,


ulfgebhardt (commented):
I approve this, jsut give us a bit more time till we are sure we don't need to have to do another deploy. Great work!

ulfgebhardt (approved):

PR #2506, opened 2023-01-03
feat(backend): add hideAmountGDD & hideAmountGDT to users table.

### LINKED ISSUE #2463: 🚀 [Feature] Store Flags to show GDD & GDT in Database
## 🚀 Feature

In order to allow users to present the Gradido-Software it must be possible to mask the GDD and GDT values to not reveal the corresponding account balance during presentation. This must be a persistent option so that even when switching devices the balance is not revealed.

Backend:
- [ ] implement two config flags(boolean) `show_gdd` and `show_gdt`
- [ ] allow the user to set the config flag via graphql-api (might be already there as universal config thing)
- [ ] Default: both values are shown

Frontend:
- [ ] retrieve the config when retrieving the user object in the frontend
- [ ] have two store values (or one for all configs) in the frontend
- [ ] make sure that the retrieved config values correspond to the show/hide of the gdt/gdd values
- [ ] trigger the setConfig call when toggling the corresponding eye symbol
- [ ] Have a new section in the user settings showing the two configs and allowing to change them

Elweyn:
### Issues

- fixes #2463

### Todo

- [X] None


Mogge (approved):
Very good PR! Migration works in both directions. Everything runs as expected! Thanx

PR #2526, opened 2023-01-09
fix(frontend): leaves are over the user symbol

### LINKED ISSUE #2513: 🐛 [Bug] [Design] Leaves are over the user symbol
## 🐛 Bugreport

Leaves are over the user symbol covering it:
image

ogerly:
## 🍰 Pullrequest
leaves are over the user symbol

### Issues
- fixes #2513

 
Bildschirmfoto vom 2023-01-09 11-41-39

### DIFF
--- a/frontend/src/layouts/DashboardLayout.vue
+++ b/frontend/src/layouts/DashboardLayout.vue
@@ -177,7 +177,7 @@ export default {
           } = result
           this.GdtBalance =
             transactionList.balance.balanceGDT === null
-              ? 0
+              ? null
               : Number(transactionList.balance.balanceGDT)
           this.transactions = transactionList.transactions
           this.balance = Number(transactionList.balance.balance)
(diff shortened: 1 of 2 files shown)


frontend/src/layouts/DashboardLayout.vue
```diff
@@ -177,7 +177,7 @@ export default {
           } = result
           this.GdtBalance =
             transactionList.balance.balanceGDT === null
-              ? null
+              ? 0
```
Mogge, line 180: Do we really want this? If for some reason the GDT server is not working. this may confuse the users a lot. `null` renders as `-`

Mogge (dismissed):

frontend/src/layouts/DashboardLayout.vue
ogerly: then we must test this times through what really happens if the GDT server times really does not go. I had yesterday the case that the GDT server did not go and **null** was detected as an error. 
I have but later noticed. so I have changed this to **0**. I change this back again.

ulfgebhardt (approved):

PR #2548, opened 2023-01-13
refactor(frontend): slots for right sidebar and header

Mogge:
### Issues

- None

### Todo

- [X] None


frontend/src/layouts/templates/ContentHeader.vue
```diff
@@ -1,116 +1,15 @@
...
-  props: {
-    balance: { type: Number, required: true },
-    GdtBalance: { type: Number, required: true },
-    totalUsers: { type: Number, required: true },
-  },
-
   computed: {
     path() {
-      return this.$route.path
+      return this.$route.path.replace(/^\//, '')
```
mahula, line 12: The unit test suite fails related to the `replace` method:

`test/testSetup.js:99
  throw new Error(w);
        ^
Error: Error in nextTick: "Error: Error in render: "TypeError: Cannot read properties of undefined (reading 'replace')""
`

mahula: More precise, the error ist related to the [time mock in DashboardLayout.spec.js](https://github.com/gradido/gradido/blob/031da1b028a5ee63fe2fd4d1f8be28c821d4c925/frontend/src/layouts/DashboardLayout.spec.js#L91).

mahula (approved):
Good refactoring towards slot utilization in the frontend.

PR #2554, opened 2023-01-16
refactor(backend): properly define custom Error class for logging errors

ulfgebhardt:
## 🍰 Pullrequest

`LogError` class instead of `logger.error`

### Issues

- None

### Todo

- [X] None

### DIFF
--- a/backend/src/server/LogError.ts
+++ b/backend/src/server/LogError.ts
@@ -1,0 +1,9 @@
+import { backendLogger as logger } from './logger'
+
+export default class LogError extends Error {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  constructor(msg: string, ...details: any[]) {
+    super(msg)
+    logger.error(msg, ...details)
+  }
+}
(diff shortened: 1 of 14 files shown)


backend/src/server/LogError.ts
```diff
@@ -0,0 +1,9 @@
+import { backendLogger as logger } from './logger'
+
+export default class LogError extends Error {
```
Mogge, line 3: Did you think about modifying the Error class itself? So we avoid those terrible imports.

Mogge (changes requested):
I understand the idea, but I totally disagree with the implementation.
 * We cannot log detailed information without exposing it to the client
 * Why not override the error class itself with an optional parameter for the log?

backend/src/server/LogError.ts
ulfgebhardt: I thought about it - but its advised never to do that, even tho possible. This is the proper way to do it. Further more its not required to expose anything to the client.

We can log any data without exposing it to the client.

ulfgebhardt:
closing this in favor of https://github.com/gradido/gradido/pull/2640

PR #2569, opened 2023-01-18
Fix pagination frontend

Elweyn:
## 🍰 Pullrequest

Pagination of Gdd- GdtList could not be shown.

### Issues

- None

### Todo

- [X] Make pagination config in Transaction.vue and give it through to GddTransactionList & GdtTransactionList

### DIFF
--- a/frontend/src/pages/Transactions.vue
+++ b/frontend/src/pages/Transactions.vue
@@ -79,13 +79,18 @@ export default {
       this.$emit('update-transactions', pagination)
     },
   },
-  created() {
+  mounted() {
     if (this.gdt) {
       this.updateGdt()
     }
   },
   watch: {
     currentPage() {
+      if (this.gdt) {
+        this.updateGdt()
+      }
+    },
+    gdt() {
       if (this.gdt) {
         this.updateGdt()
       }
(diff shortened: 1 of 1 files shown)


frontend/src/pages/Transactions.vue
```diff
@@ -79,7 +80,7 @@ export default {
       this.$emit('update-transactions', pagination)
     },
   },
-  created() {
+  mounted() {
```
Mogge, line 83: Why this? Fetch data after mounted?

frontend/src/pages/Transactions.vue
```diff
@@ -15,6 +15,7 @@
         :transactionLinkCount="transactionLinkCount"
         :transactions="transactions"
         :showPagination="true"
+        :pageSize="pageSize"
```
Mogge, line 18: Default is 25, so this should not change the behavior at all

frontend/src/pages/Transactions.vue
```diff
@@ -90,6 +91,11 @@ export default {
         this.updateGdt()
       }
     },
+    gdt() {
```
Mogge, line 94: Is this in the other PR?

Mogge (approved):
It is better to have it.

PR #2636, opened 2023-01-30
refactor(backend): refine assembly of error message in user resolver

mahula:
## 🍰 Pullrequest
The user resolver's error message for `forgotPAssword` requests in case the email cannot be resend contained the unit of time twice.


backend/src/graphql/resolver/EmailOptinCodes.test.ts
```diff
@@ -96,7 +96,7 @@ describe('EmailOptinCodes', () => {
         mutate({ mutation: forgotPassword, variables: { email: 'peter@lustig.de' } }),
       ).resolves.toMatchObject({
         data: null,
-        errors: [new GraphQLError('email already sent less than 10 minutes minutes ago')],
+        errors: [new GraphQLError('email already sent less than 10 minutes ago')],
```
Mogge, line 99: Good catch!

Mogge (approved):

PR #2647, opened 2023-02-01
fix(frontend): simple disabled function on submit send

### LINKED ISSUE #2590: 🐛 [Bug] To long text in send form has send button still enabled
## 🐛 Bugreport

Validation works, but button is active when text is too long.

Offenbar ist der Button aktiv markiert, auch wenn die Nachricht zu lang ist.
image
image

ogerly:
## 🍰 Pullrequest
simple disabled function on submit button in send form

### Issues
- fixes #2590 

https://user-images.githubusercontent.com/1324583/216031904-995d4116-02eb-48c8-b6a3-bd3f100c5f49.mp4


frontend/src/components/GddSend/TransactionForm.vue
```diff
@@ -177,6 +177,17 @@ export default {
     },
   },
   computed: {
+    disabled() {
```
Mogge, line 180: Can we have a solution using the validator?

ogerly: I have tried it but since the inputs are components. this seems somehow not to go. I would need help.

Mogge (dismissed):

Mogge (approved):

PR #2651, opened 2023-02-01
feat(federation): implement a graphql endpoint to answer getpublickey-request

clauspeterhuebner:
## 🍰 Pullrequest

Implement in the federation-Modul:
- remove all TestResolvers for multiple graphql endpoint of api-version tests
- add new graphql endpoint with api version 1_0 for request getPublicKey
  - without Inputparameters
  - read homecommunity entry with mathching api version from communities table
  - if missing home-community entry throw error (entring the home-community should be part of dht-node at starttime)
  - return the own publicKey from homecommunity-entry

### Issues

- None

### Todo

- [X] remove all TestResolvers for multiple graphql endpoint of api-version tests
- [ ] add new graphql endpoint with api version 1_0 for request getPublicKey
     * without Inputparameters
     * read homecommunity entry with mathching api version from communities table
     * if missing home-community entry throw error (entring the home-community should be part of dht-node at starttime)
     * return the own publicKey from homecommunity-entry


database/migrations/0060-update_communities_table.ts
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * This migration creates the `community` and 'communityfederation' tables in the `apollo` database (`gradido_community`).
+ */
+
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+export async function upgrade(queryFn: (query: string, values?: any[]) => Promise<Array<any>>) {
+  await queryFn(
+    'ALTER TABLE `communities` ADD COLUMN `foreign` tinyint(4) NOT NULL DEFAULT 0 AFTER `id`;',
```
Mogge, line 11:
```suggestion
    'ALTER TABLE `communities` ADD COLUMN `foreign` tinyint(4) NOT NULL DEFAULT 1 AFTER `id`;',
```

database/entity/0060-update_communities_table/Community.ts
```diff
@@ -0,0 +1,51 @@
...
+  CreateDateColumn,
+  UpdateDateColumn,
+} from 'typeorm'
+
+@Entity('communities')
+export class Community extends BaseEntity {
+  @PrimaryGeneratedColumn('increment', { unsigned: true })
+  id: number
+
+  @Column({ name: 'foreign', type: 'bool', nullable: false, default: false })
```
Mogge, line 15:
```suggestion
  @Column({ name: 'foreign', type: 'bool', nullable: false, default: true })
```

Mogge (changes requested):
Better to have default to `true` on field `foreign` as we want to have only one community that is not foreign, I guess

Issue #2487, opened 2022-12-26
🔧 [Refactor] Error logger to log the complete object


ulfgebhardt:
## 🔧 Refactor ticket

The error logger currently does something like
```
logger.error(`error blabla ${error}`)
```
This will result in an error message like `error blabla [object][object]`.
Simply by using the console.log mechanic the value of the error messages is improved alot, because the full object is logged:
```
logger.error('error blabla', error)
```

Issue #1870, opened 2022-05-04
💥 [DevOps] Backend: 80% Coverage


ulfgebhardt:
## 💥 DevOps ticket

Backend: 80% Coverage

PR #2662, opened 2023-02-03
refactor(backend): use LogError on contributionLinkResolver

ulfgebhardt:
## 🍰 Pullrequest

fix logger on contributionLinkResolver

### Issues

- contributes to #2487 
- contributes to #1870 

### Todo

- [X] None


backend/src/graphql/resolver/ContributionLinkResolver.ts
```diff
@@ -39,35 +40,30 @@ export class ContributionLinkResolver {
     }: ContributionLinkArgs,
   ): Promise<ContributionLink> {
     isStartEndDateValid(validFrom, validTo)
+    // TODO: this should be enforced by the schema.
```
Mogge, line 43: It is. If you call the API without `name`, you get a schema error.
The test is the same as `name.length === 0` and there for obsolete as we test that length must be greater than `x`

backend/src/graphql/resolver/ContributionLinkResolver.ts
```diff
@@ -39,35 +40,30 @@ export class ContributionLinkResolver {
...
-      const msg = `The value of 'name' with a length of ${name.length} did not fulfill the requested bounderies min=${CONTRIBUTIONLINK_NAME_MIN_CHARS} and max=${CONTRIBUTIONLINK_NAME_MAX_CHARS}`
-      logger.error(`${msg}`)
-      throw new Error(`${msg}`)
+    if (name.length < CONTRIBUTIONLINK_NAME_MIN_CHARS) {
+      throw new LogError('The value of name is too short', name.length)
     }
+    if (name.length > CONTRIBUTIONLINK_NAME_MAX_CHARS) {
+      throw new LogError('The value of name is too long', name.length)
+    }
+    // TODO: this should be enforced by the schema.
```
Mogge, line 53: It is. See above. `'' ? true : false  -> false`

Mogge (changes requested):
Please kick those to dos.

backend/src/graphql/resolver/ContributionLinkResolver.ts
ulfgebhardt: image

the whole thing requires refactoring - lets talk about solutions

Mogge: Yes, we can do this. My point is, that the test `(!name)` can only be true if the name is an empty string and therefore useless, as we test later for a minimum length

Mogge: How can we localize error messages thrown by this validator?

ulfgebhardt: > How can we localize error messages thrown by this validator?

We do not localize Error messages - but I am sure there is a way.

ulfgebhardt: I see your point and will remove the check & test

ulfgebhardt: Solved with your suggestion and created Issue #2672 for the validator part.

Mogge (approved):

PR #2665, opened 2023-02-03
feat(federation): harmonize and sync modules and data of federation

clauspeterhuebner:
### Issues

- None

### Todo

- [X] change database communities-table for homeCommunity-Entries
- [X] add dht-node federation-config properties for URL and default Port of Api-Version Endpoints
- [X] in dht-node insert HomeCommunity-Entry for each own supported Api-Version before starting dht-announcement
- [X] in dht-node change Api-Version to pure number without leading 'v', as expected for the Request getPublicKey in modul backend and federation
- [X] adapt and change communities-data send by dht-node to match the expected data by the backend-graphql-client from database for validating communities


database/migrations/0060-update_communities_table.ts
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * This migration creates the `community` and 'communityfederation' tables in the `apollo` database (`gradido_community`).
+ */
+
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+export async function upgrade(queryFn: (query: string, values?: any[]) => Promise<Array<any>>) {
+  await queryFn(
+    'ALTER TABLE `communities` MODIFY COLUMN `last_announced_at` datetime(3) AFTER `end_point`;',
```
ulfgebhardt, line 11: In your other PR you have the same migration but without `last_announced_at` - something seems fishy here.

clauspeterhuebner: its not fishy than its more tricky to keep several PR in sync without knowing which one will reach an approve to be earlier than the other...
the first merged PR wins and the follower PRs have to be adapted on the masters state...

dht-node/src/dht_node/index.ts
```diff
@@ -184,3 +187,38 @@ export const startDHT = async (topic: string): Promise<void> => {
     logger.error('DHT unexpected error:', err)
   }
 }
+
+async function writeHomeCommunityEnries(pubKey: any): Promise<CommunityApi[]> {
+  const homeApiVersions: CommunityApi[] = Object.values(ApiVersionType).map(function (apiEnum) {
+    const port =
+      Number.parseInt(CONFIG.FEDERATION_COMMUNITY_API_PORT) + Number(apiEnum.replace('_', ''))
```
Elweyn, line 194: I don't understand why you want to add the API version to the port! Or is it the concept that we call different GraphQL endpoints for different API versions?

dht-node/src/dht_node/index.ts
```diff
@@ -184,3 +187,38 @@ export const startDHT = async (topic: string): Promise<void> => {
...
+    const comApi: CommunityApi = {
+      api: apiEnum,
+      url: CONFIG.FEDERATION_COMMUNITY_URL + ':' + port.toString() + '/api/',
+    }
+    return comApi
+  })
+  try {
+    // first remove privious existing homeCommunity entries
```
Elweyn, line 202: You don't remove the homeCommunity here but the foreignCommunities

dht-node/src/dht_node/index.ts
Elweyn: So for version 1_0 we would get port 5010? 
This seems not right for me! Why don't we use the version in the endpoint url instead of making some magic on ports

Elweyn: Cause with this we would not use the structure you used in the Test where you showed that we can have different calls per version!
In my opinion the endpoint url should resemble to this:
- http://localhost:5000/api/v1_0
- http://localhost:5000/api/v2_0
- http://localhost:5000/api/1_0
- http://localhost:5000/api/2_0

dht-node/src/dht_node/index.test.ts
```diff
@@ -710,17 +719,17 @@ describe('federation', () => {
                   Buffer.from(
                     JSON.stringify([
                       {
-                        api: 'v1_0',
-                        url: 'http://localhost:4000/api/v1_0',
+                        api: '1_0',
+                        url: 'http://localhost:5001/api/',
```
Elweyn, line 723: Like written in the other file I prefer the logic with the API version in the URL instead of making some port magic

Elweyn (changes requested):

dht-node/.env.dist
```diff
@@ -1,3 +1,5 @@
+CONFIG_VERSION=v2.2023-02-03
```
ulfgebhardt, line 1: please not have the config version in here, we removed that

dht-node/src/dht_node/index.test.ts
```diff
@@ -44,10 +44,11 @@ const lookupResultMock = {
   token: Buffer.from(TEST_TOPIC),
   from: {
     id: Buffer.from('somone'),
+    foreign: true,
```
ulfgebhardt, line 47: is this part of the DHT library? Because this is DHT-related?
Isn't `foreign` our new database key?

dht-node/src/dht_node/index.ts
clauspeterhuebner: this implementation mirrors the requirements and results from our last meeting about this topic.
the pattern for an graphql-endpoint looks like:
<host>:<port>/api/<apiversion>/

<apiversion> = no leading 'V', only numbers and underscore, <major-version>_<minor-version>

apiversion should mirror the port-number

the port follows the rules:
- baseport of federation is 5000
- till 5009 is reserverd for DHT
- version V1_0 => 5010
- version V1_1 => 5011
- version V2_0 => 5020
- ...

dht-node/src/dht_node/index.test.ts
clauspeterhuebner: as answered on your comment before, the endpoint, which will be used to communitcate with the other community will follow the pattern \<host\>:\<port\>/api/<version-nr>/
the url in the database of a foreign community defines this part \<host\>:\<port\>/api/
the api in the database of the foreign community will be appended on the url.
so the resulting endpoint which the grphql-client will invoke looks like
\<host\>:\<port\>/api/\<version-nr\>/

the magic on the port is a requirement from Ulf and Moritz...

dht-node/src/dht_node/index.ts
clauspeterhuebner: why do you think, that instead of home-community-entries the foreign-community-entries are removed?
the home-community-entries have the foreign-flag = false.

dht-node/.env.dist
clauspeterhuebner: deleted

dht-node/src/dht_node/index.ts
```diff
@@ -184,3 +187,38 @@ export const startDHT = async (topic: string): Promise<void> => {
...
+    const homeComs = await DbCommunity.find({ foreign: false })
+    if (homeComs.length > 0) {
+      await DbCommunity.remove(homeComs)
+    }
```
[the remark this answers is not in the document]
clauspeterhuebner: changed to a single queryBuilder statement

dht-node/src/dht_node/index.ts
Elweyn: > the port follows the rules:
>
> - baseport of federation is 5000
> - till 5009 is reserverd for DHT
> - version V1_0 => 5010
> - version V1_1 => 5011
> - version V2_0 => 5020
> - ...

OK I understand what you are saying, I think that it will be difficult to setup every endpoint.

The test project was with one endpoint that would be called with the following url:
- 1_0: ``https://localhost:5000/api/1_0``
- 1_1: ``https://localhost:5000/api/1_1``
- 2_0: ``https://localhost:5000/api/2_0``
But now you want to initialize them for each version, the unit test say that it would give this urls:
- 1_0: ``https://localhost:5001/api/``
- 1_1: ``https://localhost:5002/api/``
- 2_0: ``https://localhost:5003/api/``

clauspeterhuebner: yes that's correct, the testdata will not exact match the data, which the dht-node will write in database.
But all the TestResolvers for all three apiversions will be removed with the next federation implementation step. 
They where only implemented to ensure a technical solution for a versionized graphql-endpoint, because graphql for itself will not support versioning of endpoints.
So on my opinion i will not change these unit tests anymore for code, which will be removed. Can you agree?

dht-node/src/dht_node/index.ts
```diff
@@ -184,3 +187,35 @@ export const startDHT = async (topic: string): Promise<void> => {
...
+      const homeCom = new DbCommunity()
+      homeCom.foreign = false
+      homeCom.apiVersion = homeApi.api
+      homeCom.endPoint = homeApi.url
+      homeCom.publicKey = pubKey.toString('hex')
+
+      // this will NOT update the updatedAt column, to distingue between a normal update and the last announcement
+      await DbCommunity.insert(homeCom)
+      logger.info(`federation home-community inserted successfully: ${JSON.stringify(homeCom)}`)
+    })
```
Mogge, line 215:
```suggestion
    })
    return homeApiVersions
```

dht-node/src/dht_node/index.ts
```diff
@@ -184,3 +187,35 @@ export const startDHT = async (topic: string): Promise<void> => {
...
+  } catch (err) {
+    throw new Error(`Federation: Error writing HomeCommunity-Entries: ${err}`)
+  }
+
+  return homeApiVersions
```
Mogge, line 220:
```suggestion
```

dht-node/src/dht_node/index.ts
```diff
@@ -31,13 +31,16 @@ export const startDHT = async (topic: string): Promise<void> => {
...
+    /*
     const ownApiVersions = Object.values(ApiVersionType).map(function (apiEnum) {
       const comApi: CommunityApi = {
         api: apiEnum,
-        url: CONFIG.FEDERATION_COMMUNITY_URL + apiEnum,
+        url: CONFIG.FEDERATION_COMMUNITY_URL,
       }
       return comApi
     })
+    */
```
Mogge, line 43: Do we still need this comment?

Mogge (dismissed):
Some small things. Looks good for me

dht-node/src/dht_node/index.ts
clauspeterhuebner: removed

dht-node/src/dht_node/index.ts
clauspeterhuebner: the signature of this methode needs a return value of type Promise<CommunityApi[]>
if the return statement is shifted in the try-catch-statement the compiler shows missing return value error

dht-node/src/dht_node/index.ts
clauspeterhuebner: the signature of this methode needs a return value of type Promise<CommunityApi[]>
if the return statement is shifted in the try-catch-statement the compiler shows missing return value error

ulfgebhardt (approved):

PR #2667, opened 2023-02-06
refactor(other): don't rebuild modul if unit test file has been changed

Elweyn:
## 🍰 Pullrequest

As developer I don't want to rebuild modules if the test file has been changed. This PR fixes it

### Issues

- None

### Todo

- [X] Frontend: Block rebuild of docker (nodemon) when unit-test file
- [X] Admin: Block rebuild of docker (nodemon) when unit-test file
- [X] Backend: Block rebuild of docker (nodemon) when unit-test file


mahula (approved):
Good thing to do.

PR #2668, opened 2023-02-06
refactor(backend): get last transaction by only one function

### LINKED ISSUE #2658: 🔧 [Refactor] Last transaction & Sort Order
## 🔧 Refactor ticket

Last transaction & Sort Order

- Unify the `getLastTransaction` logic into one function
- Fix Sort Order: Order by date desc + id desc

Mogge:
## 🍰 Pullrequest
have under resolver a access layer folder. There you find the function to get the last transaction to a given user, with optional relations.

I tried this with production data, and the contributions that cannot be confirmed are confirmed correctly with this fix

### Issues

- fixes #2658

### Todo

- [X] None


backend/src/graphql/resolver/BalanceResolver.ts
```diff
@@ -15,6 +15,8 @@ import { calculateDecay } from '@/util/decay'
 import { RIGHTS } from '@/auth/RIGHTS'
 import { GdtResolver } from './GdtResolver'
 
+import { getLastTransaction } from './accessLayer/getLastTransaction'
```
ulfgebhardt, line 18: please use the utils folder - making new folders without the proper refactor is making our code base impossible to maintain

ulfgebhardt (changes requested):
statistic resolver also have an occurence

backend/src/graphql/resolver/accessLayer/getLastTransaction.ts
```diff
@@ -0,0 +1,14 @@
+import { Transaction as DbTransaction } from '@entity/Transaction'
```
ulfgebhardt, line 1: this file requires a test

backend/src/graphql/resolver/BalanceResolver.ts
Mogge: I do not like the repository construction. I wanted to place more stuff in the new folder in further refactoring

backend/src/graphql/resolver/accessLayer/getLastTransaction.ts
Mogge: Right!

backend/src/graphql/resolver/BalanceResolver.ts
ulfgebhardt: The repository construct is the one we have + the utils folder. I don't think its wise to make a third structure within this PR.

ulfgebhardt: [This](https://medium.com/the-crowdlinker-chronicle/best-way-to-inject-repositories-using-typeorm-nestjs-e134c3dbf53c) seems like an interesting read to possibly reduce
```
import { UserRepository } from '@repository/User'
const userRepository = getCustomRepository(UserRepository)
const someValue = await userRepository.someQuery({..})
```
to 
```
const someValue = await userEntity.someQuery({..})
```

But hey, i dont know

ulfgebhardt: Read more here:

image

https://stackoverflow.com/questions/71557301/how-to-workraound-this-typeorm-error-entityrepository-is-deprecated-use-repo

I think it is to summarize: TypeORM seems to have detected its shitty design and now might offer ways to overcome shortcommings.
I strongly believe we will be better off using their mechanics then our custom stuff.

ulfgebhardt (approved):

PR #2669, opened 2023-02-06
refactor(backend): use LogError on contributionResolver

ulfgebhardt:
## 🍰 Pullrequest

fix logger on contributionResolver

### Issues

- contributes to #2487 
- contributes to #1870 
- based on #2663

### Todo

- [X] None

PR #2673, opened 2023-02-07
refactor(backend): use LogError on TransactionLinkResolver

ulfgebhardt:
## 🍰 Pullrequest

fix logger on TransactionLinkResolver

### Issues

- contributes to #2487 
- contributes to #1870 
- based on #2669

### Todo

- [X] None

### DIFF
--- a/backend/src/graphql/resolver/TransactionLinkResolver.test.ts
+++ b/backend/src/graphql/resolver/TransactionLinkResolver.test.ts
@@ -24,6 +24,7 @@ import { UnconfirmedContribution } from '@model/UnconfirmedContribution'
 import Decimal from 'decimal.js-light'
 import { GraphQLError } from 'graphql'
 import { TRANSACTIONS_LOCK } from '@/util/TRANSACTIONS_LOCK'
+import { logger } from '@test/testSetup'
 
 // mock semaphore to allow use fake timers
 jest.mock('@/util/TRANSACTIONS_LOCK')
@@ -61,6 +62,7 @@ describe('TransactionLinkResolver', () => {
         })
 
         it('throws error when link does not exists', async () => {
+          jest.clearAllMocks()
           await expect(
             mutate({
               mutation: redeemTransactionLink,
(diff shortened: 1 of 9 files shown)


Mogge (changes requested):
I see files I already have reviewed. Please clean up

backend/src/graphql/resolver/TransactionLinkResolver.test.ts
```diff
@@ -398,242 +532,210 @@ describe('TransactionLinkResolver', () => {
...
+          await mutate({
+            mutation: login,
+            variables: { email: 'peter@lustig.de', password: 'Aa12345_' },
           })
+        })
+
+        afterAll(async () => {
+          await cleanDB()
+          resetToken()
+        })
```
Elweyn, line 565: I would expect this to be at line 486 without the login of the admin

Elweyn (dismissed):

backend/src/graphql/resolver/TransactionLinkResolver.test.ts
```diff
@@ -50,7 +52,75 @@ afterAll(async () => {
 })
 
 describe('TransactionLinkResolver', () => {
+  describe('createTransactionLink', () => {
```
Mogge, line 55: Test for unauthenticated first?

Mogge (changes requested):
Missing authenticated test :(

backend/src/graphql/resolver/TransactionLinkResolver.test.ts
ulfgebhardt: This test suit seems to do that only for admin queries. I'd like to leave it that way.

Mogge: May be it seems like this, but it is not like this. We test `redeemTransactionLink`  which needs no authentication. Now you test `createTransactionLink` which needs authentication

Mogge (approved):

backend/src/graphql/resolver/TransactionLinkResolver.test.ts
ulfgebhardt: Documented in #2704

PR #2691, opened 2023-02-10
refactor(backend): event protocol rework

### LINKED ISSUE #1870: 💥 [DevOps] Backend: 80% Coverage

ulfgebhardt:
## 🍰 Pullrequest

Rework the event protocol.

The main changes are:
- Introduce an universal `EVENT()` function
- Have specialized functions for each event `EVENT_SOME_TYPE()` with specific what data is required for that event

image

image

This should be fixed with the deny test

### Issues

- fixes #1870 

### Todo

- [ ] Ensure every event is tested for to be in the database


backend/src/graphql/resolver/UserResolver.ts
```diff
@@ -819,10 +809,8 @@ export class UserResolver {
     if (!emailSent) {
       logger.info(`Account confirmation link: ${activationLink}`)
     } else {
-      const event = new Event()
-      const eventSendConfirmationEmail = new EventSendConfirmationEmail()
-      eventSendConfirmationEmail.userId = user.id
-      await writeEvent(event.setEventSendConfirmationEmail(eventSendConfirmationEmail))
+      // TODO: this event is used twice, why?
```
Mogge, line 812: This is admin sending confirmation email

backend/src/graphql/resolver/UserResolver.ts
```diff
@@ -333,8 +327,9 @@ export class UserResolver {
         timeDurationObject: getTimeDurationObject(CONFIG.EMAIL_CODE_VALID_TIME),
       })
       logger.info(`sendAccountActivationEmail of ${firstName}.${lastName} to ${email}`)
-      eventSendConfirmEmail.userId = dbUser.id
-      writeEvent(event.setEventSendConfirmationEmail(eventSendConfirmEmail))
+
+      // TODO: this event is used twice, why?
```
Mogge, line 331: This is registration email

Mogge (changes requested):
Please no to dos. If you do not want the same event, create a new one for admin send activation mail

Elweyn (commented):
Good changes. 
Would only need to discuss if we want a better way of distinction between the following two:
- ``EVENT`` => Creating an DbEvent
- ``EVENT_*`` => Creating an EVENT & using save on the created object.

Maybe have like before a write function: ``write(EVENT_*())``
Or rename the functions: ``saveEVENT_*``

backend/src/graphql/resolver/ContributionResolver.ts
```diff
@@ -730,12 +692,11 @@ export class ContributionResolver {
...
     const res = await contributionToUpdate.save()
 
-    const event = new Event()
-    const eventAdminContributionDeny = new EventAdminContributionDeny()
-    eventAdminContributionDeny.userId = contributionToUpdate.userId
-    eventAdminContributionDeny.amount = contributionToUpdate.amount
-    eventAdminContributionDeny.contributionId = contributionToUpdate.id
-    await writeEvent(event.setEventAdminContributionDeny(eventAdminContributionDeny))
+    await EVENT_ADMIN_CONTRIBUTION_DENY(
+      contributionToUpdate.userId,
```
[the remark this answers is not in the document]
ulfgebhardt: I changed this to `moderator.id`.

I believe that we need a better concept for the events overall - something like "We always log the account interacted with" - "if its a moderator, we log him aswell" - "we can link events" - stuff like that need a better concept, but I rather not do that in this PR.

Lets have a discussion on how we should proceed here.

My suggestion is:
- clarify if `EVENT` and `EVENT_SOME_TYPE` can have distinct different behaviour.
- merge this a good step into the right direction and have things visualized as a next step down that road, so investigation becomes easier.

backend/src/graphql/resolver/ContributionResolver.ts
```diff
@@ -730,12 +692,11 @@ export class ContributionResolver {
...
     const res = await contributionToUpdate.save()
 
-    const event = new Event()
-    const eventAdminContributionDeny = new EventAdminContributionDeny()
-    eventAdminContributionDeny.userId = contributionToUpdate.userId
-    eventAdminContributionDeny.amount = contributionToUpdate.amount
-    eventAdminContributionDeny.contributionId = contributionToUpdate.id
-    await writeEvent(event.setEventAdminContributionDeny(eventAdminContributionDeny))
+    await EVENT_ADMIN_CONTRIBUTION_DENY(
+      moderator.id,
```
Mogge, line 696: I do not agree with this change. Yes, we do not have events of this kind so far. But the user stored in the database should be the one the action affects.

Mogge: On admin delete we have stored the affected user as well. For consistency reasons, I want to see the user here as well

Mogge (changes requested):

Mogge (approved):

PR #2706, opened 2023-02-14
refactor(backend): combine logic for `listTransactionLinks` & `listTransactionLinksAdmin`

ulfgebhardt:
## 🍰 Pullrequest

Combine list transaction links and list transaction links admin

### Issues

- based on #2705

### Todo

- [X] None


ogerly (commented):
nice

backend/src/graphql/resolver/TransactionLinkResolver.ts
```diff
@@ -342,43 +318,66 @@ export class TransactionLinkResolver {
...
+}
 
-    return {
-      linkCount: count,
-      linkList: transactionLinks.map((tl) => new TransactionLink(tl, new User(user))),
-    }
+const transactionLinkList = async (
+  { currentPage = 1, pageSize = 5, order = Order.DESC }: Paginated,
+  filters: TransactionLinkFilters | null,
+  userId: number,
```
Mogge, line 357:
```suggestion
  user: DbUser,
```

backend/src/graphql/resolver/TransactionLinkResolver.ts
```diff
@@ -342,43 +318,66 @@ export class TransactionLinkResolver {
...
-    const [transactionLinks, count] = await DbTransactionLink.findAndCount({
-      where,
-      withDeleted: filters ? filters.withDeleted : false,
-      order: {
-        createdAt: order,
-      },
-      skip: (currentPage - 1) * pageSize,
-      take: pageSize,
-    })
+    return transactionLinkList(paginated, filters, userId)
```
Mogge, line 350:
```suggestion
    return transactionLinkList(paginated, filters, getUser(context))
```

backend/src/graphql/resolver/TransactionLinkResolver.ts
```diff
@@ -342,43 +318,66 @@ export class TransactionLinkResolver {
...
   }
 
+  @Authorized([RIGHTS.LIST_TRANSACTION_LINKS])
+  @Query(() => [TransactionLink])
+  async listTransactionLinks(
+    @Args()
+    paginated: Paginated,
+    @Ctx() context: Context,
+  ): Promise<TransactionLinkResult> {
+    const user = getUser(context)
```
Mogge, line 328:
```suggestion
```

backend/src/graphql/resolver/TransactionLinkResolver.ts
```diff
@@ -342,43 +318,66 @@ export class TransactionLinkResolver {
...
+    return transactionLinkList(
+      paginated,
+      {
+        withDeleted: false,
+        withExpired: true,
+        withRedeemed: false,
+      },
+      user.id,
```
Mogge, line 336:
```suggestion
      getUser(context),
```

backend/src/graphql/resolver/TransactionLinkResolver.ts
```diff
@@ -342,43 +318,66 @@ export class TransactionLinkResolver {
...
-    return {
-      linkCount: count,
-      linkList: transactionLinks.map((tl) => new TransactionLink(tl, new User(user))),
-    }
+const transactionLinkList = async (
+  { currentPage = 1, pageSize = 5, order = Order.DESC }: Paginated,
+  filters: TransactionLinkFilters | null,
+  userId: number,
+): Promise<TransactionLinkResult> => {
+  const user = await DbUser.findOneOrFail({ id: userId })
```
Mogge, line 359:
```suggestion
```

backend/src/graphql/resolver/TransactionLinkResolver.ts
```diff
@@ -342,43 +318,66 @@ export class TransactionLinkResolver {
...
+  const { withDeleted, withExpired, withRedeemed } = filters || {
+    withDeleted: false,
+    withExpired: false,
+    withRedeemed: false,
+  }
+  const [transactionLinks, count] = await DbTransactionLink.findAndCount({
+    where: {
+      userId,
```
Mogge, line 367:
```suggestion
      userId: user.id,
```

Mogge (dismissed):
Nice refactor. Please do not query the user, as it is already queried in the context. You broke the frontend :-1:  Fix `frontend/src/components/Transactions/TransactionLinkSummary.vue`  .

backend/src/graphql/resolver/TransactionLinkResolver.ts
ulfgebhardt: The admin provides the user id and is not limited to the user from context - thats the acutal refactor - thats I am doing here - maybe i dont get your suggestion - but i think it break the functionality.

ulfgebhardt:
> TransactionLinkSummary.vue

Good catch regarding the Frontend - I think we should discuss this:
```
apolloQueryMock.mockResolvedValue({
            data: {
              listTransactionLinks: [
                  {
                    amount: '76',
                    link: 'http://localhost/redeem/ce28664b5308c17f931c0367',
                    createdAt: '2022-03-16T14:22:40.000Z',
                    holdAvailableAmount: '5.13109484759482747111',
                    id: 87,
                    memo:
                      'Hat jemand die Nummer von der Hexe aus Schneewittchen? Ich bräuchte mal ein paar Äpfel.',
                    redeemedAt: null,
                    validUntil: '2022-03-30T14:22:40.000Z',
                  },
             ...
             ]
             }})
```

I explicitly did not touch this test file to ensure that things are working as expected - but as far as I understand our test concept, we cannot detect if an API change has an inpact on our frontend, since we mock the results.

I'd like to discuss this and if we can have a solution, so our tests would detect this breaking change @Mogge @mahula  .

Mogge (changes requested):
Frontend still broken. Can you test this on your machine? `message | 'Cannot query field "links" on type "TransactionLink". Did you mean "link"?'`

Mogge:
Regarding the schema tests in the frontend. In older tests, we do not test scheme aware, we just mock responses. In newer tests, where we use an apollo provider with a mocked client (e.g. `admin/src/pages/Overview.spec.js`, the tests are schema aware and find such problems as in this case. So with some more refactoring, we should achieve the goal to have schema aware unit tests in both frontends.

backend/src/graphql/resolver/TransactionLinkResolver.ts
```diff
@@ -342,43 +318,34 @@ export class TransactionLinkResolver {
...
-      },
-      skip: (currentPage - 1) * pageSize,
-      take: pageSize,
-    })
-
-    return {
-      linkCount: count,
-      linkList: transactionLinks.map((tl) => new TransactionLink(tl, new User(user))),
-    }
+    return transactionLinkList(paginated, filters, await DbUser.findOneOrFail({ id: userId }))
```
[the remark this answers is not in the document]
Mogge: Otherwise we could expose details of our user table

ulfgebhardt:
I actually did not break the frontend - my blind changes in the second attempt were correct; I broke the backend tho. Here I had the `@Query` decorator not adjusted to the new return type of the `listTransactionLinks` function. You can see the change in https://github.com/gradido/gradido/pull/2706/commits/75d23c5dafea30967d0a5bbf498c014beab86ee3

image

This cannot be type-checked by typescript and is not validated within the `typegraphql` framework on build time, but on runtime - which is the 400 error you encountered.

image

There are linting options tho according to my research: https://github.com/borremosch/eslint-plugin-type-graphql/blob/HEAD/docs/rules/invalid-decorated-type.md

I now tested this in the frontend and have a correct list displayed in both frontends:

image
image

I beleive this is therefore correctly working now.

ulfgebhardt:
> Regarding the schema tests in the frontend. In older tests, we do not test scheme aware, we just mock responses. In newer tests, where we use an apollo provider with a mocked client (e.g. `admin/src/pages/Overview.spec.js`), the tests are schema aware and find such problems as in this case. So with some more refactoring, we should achieve the goal to have schema aware unit tests in both frontends.

As stated in my last comment the actual error was not a client side schema issue, but a server side schema issue and I believe it could not be detected on frontend tests even if they were schema aware. My first problem was a client-schema-problem tho and could have been detected that way.
I believe using the apollo-mock-client is very wise, but agree that we might rethink our seed & mock data concept before we start defining alot of data objects in our code. I made the following issue to address this shortcomings: https://github.com/gradido/gradido/issues/2736

Furthermore I want to apologize - since I did not run this, I externalized the effort in getting stable software to you @Mogge instead of doing the whole journey myself. I can see in this particular example that working blindly on the code without running it is not yet sufficient to ensure things are running correctly - and maybe never fully will. I promise to test things on a running instance in the future - especially when it comes to graphql related things. Sorry!

Mogge (approved):

PR #2727, opened 2023-02-21
feat(backend): alias in update user info

Mogge:
### Issues

- new defined rule-set find in #2874 

### Todo

- [X] None


Mogge:
I decided not to use the class validator. The overhead to create proper, easy to handle error messages is so high, that it is in no relation to the advantages we may gain.

backend/src/graphql/resolver/UserResolver.test.ts
```diff
@@ -1134,6 +1134,129 @@ describe('UserResolver', () => {
...
+
+        describe('valid alias', () => {
+          it('updates the user in DB', async () => {
+            await mutate({
+              mutation: updateUserInfos,
+              variables: {
+                alias: 'bibiBloxberg',
+              },
+            })
+            await expect(User.findOne()).resolves.toEqual(
```
Elweyn, line 1251: Are you sure that we will always only have one entry in ``users``?
Please search for the user that should have the alias setten to prevent eventuell flakyness.
For example like this:
```suggestion
            await expect(User.findOne({ firstname: 'Bibi', lastname: 'Bloxberg' })).resolves.toEqual(
```

Elweyn (dismissed):

backend/src/graphql/arg/UpdateUserInfosArgs.ts
```diff
@@ -1,4 +1,5 @@
 import { ArgsType, Field } from 'type-graphql'
+// import { Length } from 'class-validator'
```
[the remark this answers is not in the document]
Elweyn: Have only seen the comment you did on the PR later.
Nice that you tried it, but if you prefer the error handling with the throw error mechanismus then lets remove this.

backend/src/graphql/resolver/UserResolver.test.ts
```diff
@@ -1134,6 +1134,129 @@ describe('UserResolver', () => {
...
+              mutation: login,
+              variables: {
+                email: 'bibi@bloxberg.de',
+                password: 'Aa12345_',
+              },
+            })
+          })
+
+          afterAll(async () => {
+            const [user] = await User.find({ id: peter.id })
```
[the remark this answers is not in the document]
Mogge: Response is an array of users with find: `[user1, user2, user3, ...]`

backend/src/graphql/resolver/UserResolver.test.ts
Mogge: Yes, the test is setup to have only one user. Therefore the workaround with the existing alias

backend/src/graphql/resolver/UserResolver.ts
```diff
@@ -536,6 +537,23 @@ export class UserResolver {
...
     }
 
+    if (alias) {
+      if (alias.length < 5) {
+        throw new LogError('Given alias is too short', alias)
+      }
+      if (alias.length > 20) {
+        throw new LogError('Given alias is too long', alias)
+      }
+      if (!alias.match(/^[0-9A-Za-z]+$/)) {
```
clauspeterhuebner, line 547: whats the meaning of the ending $-character in the exp?
missing allowed - and _ character.

clauspeterhuebner (changes requested):
some questions

backend/src/graphql/arg/UpdateUserInfosArgs.ts
```diff
@@ -8,6 +9,10 @@ export default class UpdateUserInfosArgs {
   @Field({ nullable: true })
   lastName?: string
 
+  @Field({ nullable: true })
+  //  @Length(5, 20)
```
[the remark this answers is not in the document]
Mogge: Yes

backend/src/graphql/resolver/UserResolver.ts
Mogge: $ is end of string

backend/src/graphql/resolver/UserResolver.test.ts
```diff
@@ -1134,6 +1134,129 @@ describe('UserResolver', () => {
...
+          })
+        })
+
+        describe('invalid characters', () => {
+          it('throws and logs an error', async () => {
+            await expect(
+              mutate({
+                mutation: updateUserInfos,
+                variables: {
+                  alias: 'no_underscore',
```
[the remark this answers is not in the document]
Mogge: See my new regex

backend/src/graphql/resolver/UserResolver.ts
```diff
@@ -536,6 +537,23 @@ export class UserResolver {
       userEntity.lastName = lastName
     }
 
+    if (alias) {
+      if (alias.length < 5) {
+        throw new LogError('Given alias is too short', alias)
+      }
+      if (alias.length > 20) {
```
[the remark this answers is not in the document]
clauspeterhuebner: Base on previous discussions about this topic and the usecase document, an alias will
- be an additional unique key, which can be used outside the community to identify an user
- a human readable unique key
- be defined as an alphanummeric string plus the character - and _ with a length between 5 and 20
- not support the character @ so it will be impossible to enter an email as an alias
- be shorter for predefinable system/community relevant identifiers
- support a blacklist to avoid capturing of predefinable strings

The basic idea is to have different (human readable and cryptical) user identifiers, which can be used crossover communities.
To find and access an user all identifiers could be used. But to store user information as a reference in business data e.g. in the receiving part of a cross-community-transaction the community-uuid + the users-uuid should be used and neither the alias nor the email. 
The currently used userid will only reference to internal users and should be migrated to the two columns community-uuid and user-uuid. 
With both uuids you can always identify an user crossover communities, even if the theoretical case exists that two users from differen communities will have the same uuid.

clauspeterhuebner:
please find in the decription of this PR a new definition for alias-rules and first elements of blacklist

backend/src/graphql/resolver/UserResolver.ts
```diff
@@ -536,6 +537,23 @@ export class UserResolver {
...
+        throw new LogError('Given alias is too long', alias)
+      }
+      if (!alias.match(/^[0-9A-Za-z]+$/)) {
+        throw new LogError('Invalid characters in alias', alias)
+      }
+      const aliasInUse = await DbUser.find({ alias })
```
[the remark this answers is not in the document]
clauspeterhuebner: as discussed in our last meeting on tuesday we convert and store the alias in lowercase, same as with email.
The alias-rules should be describe in any label/message/texts in the UI for the user

clauspeterhuebner (dismissed):

backend/src/graphql/resolver/UserResolver.ts
```diff
@@ -523,6 +525,11 @@ export class UserResolver {
       user.lastName = lastName
     }
 
+    if (alias) {
+      await validateAlias(alias)
+      user.alias = alias
+    }
+
```
ulfgebhardt, line 532:
```suggestion
    if (alias && await validateAlias(alias)) {
      user.alias = alias
    }

```

backend/src/graphql/resolver/util/validateAlias.ts
```diff
@@ -0,0 +1,37 @@
...
+  if (!alias.match(/^[0-9A-Za-z]([_-]?[A-Za-z0-9])+$/))
+    throw new LogError('Invalid characters in alias', alias)
+  if (reservedAlias.includes(alias.toLowerCase())) throw new LogError('Alias is not allowed', alias)
+  const aliasInUse = await DbUser.find({
+    where: { alias: Raw((a) => `LOWER(${a}) = "${alias.toLowerCase()}"`) },
+  })
+  if (aliasInUse.length !== 0) {
+    throw new LogError('Alias already in use', alias)
+  }
+}
```
ulfgebhardt, line 37:
```suggestion
export const validateAlias = async (alias: string): Promise<boolean> => {
  if (alias.length < 3) throw new LogError('Given alias is too short', alias)
  if (alias.length > 20) throw new LogError('Given alias is too long', alias)
  /* eslint-disable-next-line security/detect-unsafe-regex */
  if (!alias.match(/^[0-9A-Za-z]([_-]?[A-Za-z0-9])+$/))
    throw new LogError('Invalid characters in alias', alias)
  if (reservedAlias.includes(alias.toLowerCase())) throw new LogError('Alias is not allowed', alias)
  const aliasInUse = await DbUser.find({
    where: { alias: Raw((a) => `LOWER(${a}) = "${alias.toLowerCase()}"`) },
  })
  if (aliasInUse.length !== 0) {
    throw new LogError('Alias already in use', alias)
  }
  return true
}
```

ulfgebhardt, line 37:
```suggestion
export const validateAlias = async (alias: string): Promise<boolean> => {
  if (alias.length < 3) throw new LogError('Given alias is too short', alias)
  if (alias.length > 20) throw new LogError('Given alias is too long', alias)
  /* eslint-disable-next-line security/detect-unsafe-regex */
  if (!alias.match(/^[0-9A-Za-z]([_-]?[A-Za-z0-9])+$/))
    throw new LogError('Invalid characters in alias', alias)
  if (reservedAlias.includes(alias.toLowerCase())) throw new LogError('Alias is not allowed', alias)
  const aliasInUse = await DbUser.find({
    where: { alias: Raw((a) => `LOWER(${a}) = "${alias.toLowerCase()}"`) },
  })
  if (aliasInUse.length !== 0) {
    throw new LogError('Alias already in use', alias)
  }
  return true
}
```

ulfgebhardt (commented):
I like the whole PR, one thing I want you to think about is the `await validateAlias()` which fails since it can throw - this is not immediately obvious - hence I suggest to let `validateAlias()` return a boolean in the success case. This will make sure the reader will understand that only if this function is successful the alias is set since its part of the if statement.
If you disagree I will also approve, this is just a suggestion

backend/src/graphql/resolver/util/validateAlias.ts
ulfgebhardt: I changed result type to `Promise<boolean>` and added `return true`

ulfgebhardt (approved):

ulfgebhardt (approved):

PR #2745, opened 2023-02-26
refactor(backend): eslint-plugin-type-graphql + fixes

### LINKED ISSUE #2735: 💥 [DevOps] Eslint for typegraphql
## 💥 DevOps ticket

We need `typegraphql` linting to ensure return types of functions are correctly defined. Otherwise we can get runtime errors. See for discussion and details: https://github.com/gradido/gradido/pull/2706

Implement this:
https://www.npmjs.com/package/eslint-plugin-type-graphql

Especially this: https://github.com/borremosch/eslint-plugin-type-graphql/blob/HEAD/docs/rules/invalid-decorated-type.md

ulfgebhardt:
## 🍰 Pullrequest

This is the same as #2744 but includes all fixes required to have the lint green again

### Issues

- based on #2744
- fixes #2735 

### Todo

- [ ] test with production data locally


backend/src/graphql/model/Balance.ts
```diff
@@ -19,14 +19,14 @@ export class Balance {
   @Field(() => Decimal)
   balance: Decimal
 
-  @Field(() => Number, { nullable: true })
+  @Field(() => Int, { nullable: true })
```
Mogge, line 22: This is not an integer. Should be Numer

backend/src/graphql/model/GdtEntryList.ts
```diff
@@ -16,15 +19,15 @@ export class GdtEntryList {
...
-  @Field(() => Number)
+  @Field(() => Int)
   count: number
 
   @Field(() => [GdtEntry], { nullable: true })
-  gdtEntries?: GdtEntry[]
+  gdtEntries: GdtEntry[] | null
 
-  @Field(() => Number)
+  @Field(() => Int)
```
Mogge, line 28: Float?

backend/src/graphql/model/GdtEntryList.ts
```diff
@@ -16,15 +19,15 @@ export class GdtEntryList {
...
   gdtSum: number
 
-  @Field(() => Number)
+  @Field(() => Int)
```
Mogge, line 31: What is this?

backend/src/graphql/resolver/GdtResolver.ts
```diff
@@ -35,7 +39,7 @@ export class GdtResolver {
   }
 
   @Authorized([RIGHTS.GDT_BALANCE])
-  @Query(() => Number)
+  @Query(() => Int, { nullable: true })
```
Mogge, line 42: No Int- Number or Float

backend/src/graphql/resolver/TransactionLinkResolver.ts
```diff
@@ -128,7 +128,7 @@ export class TransactionLinkResolver {
   @Authorized([RIGHTS.QUERY_TRANSACTION_LINK])
   @Query(() => QueryLinkResult)
   async queryTransactionLink(@Arg('code') code: string): Promise<typeof QueryLinkResult> {
-    if (code.match(/^CL-/)) {
+    if (/^CL-/.exec(code)) {
```
Mogge, line 131: Why this change?

backend/src/graphql/resolver/TransactionLinkResolver.ts
```diff
@@ -178,7 +178,7 @@ export class TransactionLinkResolver {
     const clientTimezoneOffset = getClientTimezoneOffset(context)
     const user = getUser(context)
 
-    if (code.match(/^CL-/)) {
+    if (/^CL-/.exec(code)) {
```
Mogge, line 181: Same

backend/src/graphql/resolver/UserResolver.ts
```diff
@@ -272,22 +276,22 @@ export class UserResolver {
...
     dbUser.lastName = lastName
     dbUser.language = language
-    dbUser.publisherId = publisherId
+    dbUser.publisherId = publisherId || 0
     dbUser.passwordEncryptionType = PasswordEncryptionType.NO_PASSWORD
-    logger.debug('new dbUser=' + dbUser)
+    logger.debug('new dbUser', dbUser)
     if (redeemCode) {
-      if (redeemCode.match(/^CL-/)) {
+      if (/^CL-/.exec(redeemCode)) {
```
Mogge, line 283: Why this? I just do not like this expressions

backend/src/password/EncryptorUtils.ts
```diff
@@ -9,7 +12,7 @@ const sodium = require('sodium-native')
 
 // We will reuse this for changePassword
 export const isValidPassword = (password: string): boolean => {
-  return !!password.match(/^(?=.*[a-z])(?=.*[A-Z])(?=.*[0-9])(?=.*[^a-zA-Z0-9 \\t\\n\\r]).{8,}$/)
+  return !!/^(?=.*[a-z])(?=.*[A-Z])(?=.*[0-9])(?=.*[^a-zA-Z0-9 \\t\\n\\r]).{8,}$/.exec(password)
```
Mogge, line 15: once more

Mogge (changes requested):
Please check the GDT amount carefully

backend/src/graphql/model/Balance.ts
Mogge: Or float

ulfgebhardt: Float should be right! Good catch!

backend/src/graphql/resolver/TransactionLinkResolver.ts
ulfgebhardt: This is suggested by the typescript-eslint check. I am not a big fan of this - we could disable this check, but this is an equivalent

backend/src/graphql/model/GdtEntryList.ts
ulfgebhardt: According to our documentation its a float!

backend/src/graphql/resolver/TransactionLinkResolver.ts
ulfgebhardt: The rule in question is this https://github.com/typescript-eslint/typescript-eslint/blob/v4.33.0/packages/eslint-plugin/docs/rules/prefer-regexp-exec.md

Money quote:
```
This rule is aimed at enforcing a consistent way to apply regular expressions to strings.
[...]
`RegExp#exec` may also be slightly faster than `String#match`; this is the reason to choose it as the preferred usage.
```

In my opinion we can disable this rule and revert related changes.

`@typescript-eslint/prefer-regexp-exec: false`

I think readability beats performance in this case - the performance is the task of the engine/compiler and not me - I have to ensure maintainability and therefore readability is king.

What you think @Mogge ?

backend/src/graphql/resolver/TransactionLinkResolver.ts
ulfgebhardt: The rule in question is this https://github.com/typescript-eslint/typescript-eslint/blob/v4.33.0/packages/eslint-plugin/docs/rules/prefer-regexp-exec.md

Money quote:
```
This rule is aimed at enforcing a consistent way to apply regular expressions to strings.
[...]
`RegExp#exec` may also be slightly faster than `String#match`; this is the reason to choose it as the preferred usage.
```

In my opinion we can disable this rule and revert related changes.

`@typescript-eslint/prefer-regexp-exec: false`

I think readability beats performance in this case - the performance is the task of the engine/compiler and not me - I have to ensure maintainability and therefore readability is king.

What you think @Mogge ?

Mogge (dismissed):

Mogge (approved):

PR #2763, opened 2023-02-28
fix(backend): set email tls true in test

Mogge:
## 🍰 Pullrequest
If in local environment `CONFIG.EMAIL_TLS` is `false` the test fails without any reasons. Therefore force it to be `true`

### Issues

- None

### Todo

- [X] None


mahula (approved):
Good catch.

Elweyn (approved):
Good catch!

PR #2770, opened 2023-03-04
refactor(frontend): creations generated by link (FR) + (NL)

BerndHueckstaedt:
### Issues

- None

### Todo

- [X] None


Elweyn (approved):
Not sure if "créations générées par lien" is the right description but I actually doesnt have a better alternativ

PR #2780, opened 2023-03-06
feat(backend): events for contribution links

ulfgebhardt:
## 🍰 Pullrequest

events for contribution links

### Issues

- None

### Todo

- [X] None

Issue #2722, opened 2023-02-18
🚀 [Feature] Events: refactoring and capturing


ulfgebhardt:
## 🚀 Feature

Events

- [x] Refactor existing EventProtocol + Database
- [x] Events for every mutation on the database + test
- [x] Adjust Epic name

PR #2783, opened 2023-03-07
feat(backend): events for contribution messages

ulfgebhardt:
### Issues

- based on #2780 
- contributes to #2722

### Todo

- [X] None


backend/src/graphql/resolver/ContributionMessageResolver.ts
```diff
@@ -142,17 +152,23 @@ export class ContributionMessageResolver {
...
         language: contribution.user.language,
-        senderFirstName: user.firstName,
-        senderLastName: user.lastName,
+        senderFirstName: moderator.firstName,
+        senderLastName: moderator.lastName,
         contributionMemo: contribution.memo,
       })
       await queryRunner.commitTransaction()
+      await EVENT_ADMIN_CONTRIBUTION_MESSAGE_CREATE(
+        { id: contribution.userId } as DbUser,
```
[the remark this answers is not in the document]
Elweyn: So you hack the need to query it!?

backend/src/graphql/resolver/ContributionMessageResolver.ts
```diff
@@ -57,6 +62,11 @@ export class ContributionMessageResolver {
         await queryRunner.manager.update(DbContribution, { id: contributionId }, contribution)
       }
       await queryRunner.commitTransaction()
+      await EVENT_CONTRIBUTION_MESSAGE_CREATE(
+        user,
+        { id: contributionMessage.contributionId } as DbContribution,
```
Elweyn, line 67: Like at the bottom it feels strange to define it like this. At line 62 it seem to use the same object.

Elweyn (approved):
Seems to be a good PR; 
I just feel a bit uneasy on the actual business logic use of EVENT_ADMIN_CONTRIBUTION_MESSAGE_CREATE and EVENT_CONTRIBUTION_MESSAGE_CREATE.

backend/src/graphql/resolver/ContributionMessageResolver.ts
clauspeterhuebner, line 161: the userId is overwritten by the moderatorId in line 136 and cause here a wrong event entry

backend/src/graphql/resolver/ContributionMessageResolver.ts
```diff
@@ -112,18 +122,18 @@ export class ContributionMessageResolver {
...
-        contribution.user.emailContact = await UserContact.findOneOrFail({
+        contribution.user.emailContact = await DbUserContact.findOneOrFail({
           where: { id: contribution.user.emailId },
         })
       }
       contributionMessage.contributionId = contributionId
       contributionMessage.createdAt = new Date()
       contributionMessage.message = message
-      contributionMessage.userId = user.id
+      contributionMessage.userId = moderator.id
```
clauspeterhuebner, line 136: if you overwrite the userId with the moderatorId it will cause an inconsistence to the user-object relation.
And below for writing the event it will create a wrong event-entry.

clauspeterhuebner (dismissed):

backend/src/graphql/resolver/ContributionMessageResolver.ts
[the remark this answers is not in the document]
ulfgebhardt: > So you hack the need to query it!?

Yes, I interpret this object ( `{ id: contribution.userId }` )  as DbUser. This just effects the compile time, not the runtime - but provides typesecurity, which I explicitly hack in this case, because I know this is the `userId` and I do not want to query the full User Object from the database explicitly.

This approach combines the benefits of typechecks and performance with the downside of a bit more code.

ulfgebhardt: This is not correct - like 136 is about `contributionMessage` not `contribution.userId `. The `contribution.userId` is not touched.

`contributionMessage.userId = moderator.id`

backend/src/graphql/resolver/ContributionMessageResolver.ts
ulfgebhardt: image

There is no change here - I just rename `user` to `moderator` for consistency and readability. There is no logic change here. I decided that this was a small enough refactor to be reviewable with big impact.

Issue #2787, opened 2023-03-07
🔧 [Refactor] Fix frontend & admin for Release


ulfgebhardt:
### Fix frontend for Release:

- [x] #2796
![image](https://user-images.githubusercontent.com/1238238/223446489-4a747f8b-5423-4574-8613-aec45567cf00.png)
![image](https://user-images.githubusercontent.com/1238238/223446559-4a8d3130-aad4-4ee7-8ec0-808e38d92104.png)

#### Menu- & Logo-Abstände => https://github.com/gradido/gradido/pull/2793

- [x] Logo Abstände
`<a href="/admin/" class="navbar-brand active" target="_self" style="margin-top: -7px"><img src="img/brand/gradido_logo_w.png" alt="..." class="navbar-brand-img"></a>`
Hack: auf brand `style="margin-top: -7px"`
- [x] Menu Abstände
`<nav class="navbar p-3 navbar-dark bg-success navbar-expand-md">[...]</nav>`
remove `p3`
- [x] Menu `Schöpfungen` -> remove p1
![image](https://user-images.githubusercontent.com/1238238/223454265-2e35fd5e-898a-47df-a7df-94abdd60214f.png)
![image](https://user-images.githubusercontent.com/1238238/223454331-0161fa4b-64ee-4fcf-aa51-a18cd0be17ad.png)
![image](https://user-images.githubusercontent.com/1238238/223454198-cfe3f3ec-d6e2-4c19-b466-3a0466d87cf6.png)

#### tab `offen`
  - [x] E-Mail empty
![image](https://user-images.githubusercontent.com/1238238/223447841-1506a2bd-bcd2-4ea9-93fa-11d99dd4db53.png)
  - [x] Bearbeiten -> Chat
![image](https://user-images.githubusercontent.com/1238238/223448250-dda61dd5-07d9-4bb4-bfd3-c935754e13c7.png)
  - [x] Symbol für offen im tab oben
![image](https://user-images.githubusercontent.com/1238238/223447969-7cd18bb3-5d02-4ec1-885c-b85e08b8d8d5.png)
![image](https://user-images.githubusercontent.com/1238238/223448187-795d2074-a765-40a9-8e15-c95a191a6598.png)
  - [x] Reject -> Deny oder umgedreht
![image](https://user-images.githubusercontent.com/1238238/223449425-0a11fcb0-bdef-4b9f-abc2-6fcdb4986775.png)

#### tab `bestätigt`
  - [x] Chat is missing
[...]

PR #2795, opened 2023-03-08
fix(backend): add relation messages to database query

### LINKED ISSUE #2741: 🐛 [Bug] messagesCount always 0
## 🐛 Bugreport
in the admin area, there is no more visual representation when a moderator has written. it is because all messagesCount 0. if I write a message to a user then there is a number. but if I reload ... the value is again 0. 
but there are messages in it. 

https://user-images.githubusercontent.com/1324583/221191003-33c77568-2128-49bb-8b90-67a4eabd4047.mp4

Elweyn:
### Issues

- fixes #2741
- relates #2787

### Todo

- [X] Show chat symbol


Mogge (commented):
Why the changes in the unit test?

Elweyn:
> Why the changes in the unit test?

@Mogge cause the logic of objectContaining makes it very difficult to debug the error, so I changed the order to exactly the same order that is given back from the query.
Maybe we should refactor the querried datas an check that there is exactly the object we hope.

ulfgebhardt (approved):

PR #2811, opened 2023-03-10
fix(admin): admin open contribution edit button

### LINKED ISSUE #2803: 🐛 [Bug][Admin] open contribution missing edit button
## 🐛 Bugreport

Open contribution misses the edit button

Elweyn:
### Issues

- fixes #2803 

### Todo

- [X] None


ogerly (dismissed):

backend/src/graphql/resolver/ContributionResolver.ts
```diff
@@ -314,41 +314,28 @@ export class ContributionResolver {
...
-    }
-    if (emailContact.deletedAt || emailContact.user.deletedAt) {
-      throw new LogError('User was deleted', email)
-    }
 
     const moderator = getUser(context)
 
     const contributionToUpdate = await DbContribution.findOne({
       where: { id, confirmedAt: IsNull(), deniedAt: IsNull() },
+      relations: ['user'],
```
Mogge, line 326: @ulfgebhardt I need the user relation only for the event here. Why did you insist on a user object?

Mogge: Furthermore. This is a contradiction to the graphql idea. I should only query the user, if I really need it. No I am forced by the event to query the user always

ulfgebhardt: `{ id: myuserId } as DBUser`

admin/src/components/EditCreationFormular.vue
```diff
@@ -96,41 +96,31 @@ export default {
...
       rangeMin: 0,
-      rangeMax: 1000,
-      selected: '',
+      selected: this.selectedComputed,
+      userId: this.item.userId,
     }
   },
   methods: {
     submitCreation() {
+      // console.log('submitCreation', this.selected)
```
ulfgebhardt, line 111: remove

backend/src/graphql/resolver/ContributionResolver.test.ts
```diff
@@ -2080,15 +2077,14 @@ describe('ContributionResolver', () => {
           // stephen@hawking.uk: [1000, 1000, 1000] - deleted
           // garrick@ollivander.com: [1000, 1000, 1000] - not activated
 
-          describe('user for creation to update does not exist', () => {
+          describe.skip('user for creation to update does not exist', () => {
```
ulfgebhardt, line 2080: why skip? If the user status is no criteria anymore lets remove the test?
Is there implications for the operation if the user was deleted/does not exist?

backend/src/graphql/resolver/ContributionResolver.test.ts
```diff
@@ -2106,15 +2102,14 @@ describe('ContributionResolver', () => {
             })
           })
 
-          describe('user for creation to update is deleted', () => {
+          describe.skip('user for creation to update is deleted', () => {
```
ulfgebhardt, line 2105: why skip?

backend/src/graphql/resolver/ContributionResolver.test.ts
```diff
@@ -2158,15 +2152,14 @@ describe('ContributionResolver', () => {
             })
           })
 
-          describe('user email does not match creation user', () => {
+          describe.skip('user email does not match creation user', () => {
```
ulfgebhardt, line 2155: why skip?

backend/src/graphql/resolver/ContributionResolver.ts
ulfgebhardt, line 326:
```suggestion
```

backend/src/graphql/resolver/ContributionResolver.ts
```diff
@@ -372,10 +359,8 @@ export class ContributionResolver {
     result.memo = contributionToUpdate.memo
     result.date = contributionToUpdate.contributionDate
 
-    result.creation = await getUserCreation(emailContact.user.id, clientTimezoneOffset)
-
     await EVENT_ADMIN_CONTRIBUTION_UPDATE(
-      emailContact.user,
+      contributionToUpdate.user,
```
ulfgebhardt, line 363:
```suggestion
      { id: contributionToUpdate.userId } as DbUser,
```

backend/src/graphql/resolver/ContributionResolver.test.ts
Mogge: Same. It does not matter if the user exists or is deleted, as I do not query the user anymore

backend/src/graphql/resolver/ContributionResolver.test.ts
ulfgebhardt, line 2080: I would define that you can update an contribution even tho an user does not exist or is deleted. Therefore I would remove the two tests completely.

admin/src/components/CreationFormular.vue
```diff
@@ -129,21 +125,21 @@ export default {
...
+      userId: this.item.userId,
     }
   },
   methods: {
     updateRadioSelected(name) {
       // do we want to reset the memo everytime the month changes?
       this.text = this.$t('creation_form.creation_for') + ' ' + name.short + ' ' + name.year
       this.rangeMin = 0
-      this.rangeMax = name.creation
+      this.rangeMax = Number(name.creation)
```
ulfgebhardt, line 136: I stumbled over this function - main reason was the parameter `name` which is not correct in my opinion - I think it should be `selected` or `option`. It is a bit out of scope, but this refactoring would increase readability alot I think. If you don't want to do this, please just close.

ulfgebhardt (changes requested):
image
image
image

I cannot test this, since I can only edit Contributions in the Admin interface I created there. But apparently there is a side effect of your code that affects `adminCreateContribution` 
image

So besides minor code rather questions then suggestions this must be fix or explained why this behaviour is like that. Furthermore the tests should reflect this problem with a failure.

Mogge (commented):
I have no idea where all the eslint changes are coming from ...

ulfgebhardt (approved):
❤️

PR #2827, opened 2023-03-12
refactor(backend): eslint plugin import + fixes

ulfgebhardt:
### Issues

- None

### Todo

- [X] None


backend/src/password/EncryptorUtils.ts
```diff
@@ -1,13 +1,13 @@
...
 /* eslint-disable @typescript-eslint/no-unsafe-assignment */
+import { User } from '@entity/User'
 import CONFIG from '@/config'
 import LogError from '@/server/LogError'
 import { backendLogger as logger } from '@/server/logger'
-import { User } from '@entity/User'
 import { PasswordEncryptionType } from '@enum/PasswordEncryptionType'
 
-// eslint-disable-next-line @typescript-eslint/no-var-requires
+// eslint-disable-next-line @typescript-eslint/no-var-requires, import/no-commonjs
```
Mogge, line 10: Can this be done by an import?

backend/src/event/EVENT_ACTIVATE_ACCOUNT.ts
```diff
@@ -1,5 +1,6 @@
 import { User as DbUser } from '@entity/User'
 import { Event as DbEvent } from '@entity/Event'
+/* eslint-disable-next-line import/no-cycle */
```
Mogge, line 3: disable this rule?

backend/src/password/EncryptorUtils.ts
ulfgebhardt: I'd like to investigate this in another PR - there was a problem with `import { ... } from ' sodium-native' ` - thats why its required.
I made an Issue: https://github.com/gradido/gradido/issues/2903

backend/src/event/EVENT_ACTIVATE_ACCOUNT.ts
ulfgebhardt: Solution I implemented now: disable rule

Elweyn (commented):
Let's define which rules we find good and which one we dislike.

backend/src/event/EVENT_ACTIVATE_ACCOUNT.ts
ulfgebhardt: Better solution proposed in https://github.com/gradido/gradido/pull/2905

Mogge (approved):

PR #2837, opened 2023-03-15
feat(frontend): send coins via gradido ID

Mogge:
### Issues

- None

### Todo

- [X] None


frontend/src/graphql/mutations.js
```diff
@@ -144,6 +144,7 @@ export const createContributionMessage = gql`
 export const login = gql`
   mutation($email: String!, $password: String!, $publisherId: Int) {
     login(email: $email, password: $password, publisherId: $publisherId) {
+      gradidoID
       email
```
clauspeterhuebner, line 148: do we need the email as a result of this mutation anymore, when we get the gradidoID now?

backend/src/graphql/resolver/UserResolver.test.ts
```diff
@@ -2199,6 +2205,125 @@ describe('UserResolver', () => {
...
+        })
+      })
+
+      describe('identifier is not found', () => {
+        it('throws and logs "No user found to given identifier" error', async () => {
+          await expect(
+            query({
+              query: userQuery,
+              variables: {
+                identifier: '00000000-0000-0000-0000-000000000000',
```
clauspeterhuebner, line 2265: please see my comments in backend/src/graphql/resolver/util/findUserByIdentifier.ts how to check and verify a correct uuid of version-4.
Then this testdata of a uuid will fail.

backend/src/graphql/resolver/UserResolver.test.ts
```diff
@@ -2199,6 +2205,125 @@ describe('UserResolver', () => {
...
+              },
+            }),
+          ).resolves.toEqual(
+            expect.objectContaining({
+              errors: [new GraphQLError('No user found to given identifier')],
+            }),
+          )
+          expect(logger.error).toBeCalledWith(
+            'No user found to given identifier',
+            '00000000-0000-0000-0000-000000000000',
```
clauspeterhuebner, line 2275: please see my comments in backend/src/graphql/resolver/util/findUserByIdentifier.ts how to check and verify a correct uuid of version-4.
Then this testdata of a uuid will fail.

clauspeterhuebner (changes requested):

backend/src/graphql/resolver/util/findUserByIdentifier.ts
```diff
@@ -0,0 +1,36 @@
+import { User as DbUser } from '@entity/User'
+import { UserContact as DbUserContact } from '@entity/UserContact'
+import LogError from '@/server/LogError'
+
+export const findUserByIdentifier = async (identifier: string): Promise<DbUser | null> => {
```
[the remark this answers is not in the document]
Mogge: I am not going to move this file that far away from the resolver.

frontend/src/components/GddSend/TransactionForm.vue
```diff
@@ -131,54 +140,69 @@ export default {
...
       radioSelected: this.selected,
+      userName: '',
     }
   },
   methods: {
     onValidation() {
       this.$refs.formValidator.validate()
     },
     onSubmit() {
+      if (this.gradidoID) this.form.identifier = this.gradidoID
```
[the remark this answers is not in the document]
Mogge: This is the computed `property `computedID` of the component. See below in that file

frontend/src/components/GddSend/TransactionForm.vue
```diff
@@ -131,54 +140,69 @@ export default {
...
+      error({ message }) {
+        this.toastError(message)
+      },
     },
   },
   computed: {
     disabled() {
       if (
-        this.form.email.length > 5 &&
+        this.form.identifier.length > 5 &&
```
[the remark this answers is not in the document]
Mogge: Out of scope of this PR. I agree, this is horrible code. Welcome to the frontend

frontend/src/graphql/mutations.js
Mogge: Next step. Remove the email from the graphql User

backend/src/graphql/resolver/UserResolver.test.ts
Mogge: :+1:

backend/src/graphql/resolver/TransactionResolver.ts
```diff
@@ -302,10 +301,9 @@ export class TransactionResolver {
   @Authorized([RIGHTS.SEND_COINS])
   @Mutation(() => Boolean)
   async sendCoins(
-    @Args() { email, amount, memo }: TransactionSendArgs,
+    @Args() { identifier, amount, memo }: TransactionSendArgs,
     @Ctx() context: Context,
   ): Promise<boolean> {
-    logger.info(`sendCoins(email=${email}, amount=${amount}, memo=${memo})`)
```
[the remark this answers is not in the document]
Mogge: Right

frontend/src/components/GddSend/TransactionForm.vue
```diff
@@ -50,16 +50,24 @@
...
                         />
                       </div>
+                      <div v-else-if="gradidoID" class="mb-4">
+                        <b-row>
+                          <b-col>{{ $t('form.recipient') }}</b-col>
+                        </b-row>
+                        <b-row>
+                          <b-col class="font-weight-bold">{{ userName }}</b-col>
+                        </b-row>
+                      </div>
```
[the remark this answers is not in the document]
Mogge: I do not understand your question

clauspeterhuebner (dismissed):
On my opinion if we never start to collect utility and db-access logic as dao-classes, we will never reach a kind of access-layer.
As mentioned the findUserByIdentifier is a prime example for such a dao class. Where you locate these doa-file is not as important than it will be created and named as dao.
but with a package "util" no one will recognize the goal of a future access-layer...

Only a hint from my side...

backend/src/graphql/resolver/TransactionResolver.test.ts
```diff
@@ -81,7 +80,7 @@ describe('send coins', () => {
         await mutate({
           mutation: sendCoins,
           variables: {
-            email: 'wrong@email.com',
+            identifier: 'wrong@email.com',
```
ulfgebhardt, line 83: the tests should reflect the new capabilities of the identifier and use gradido-id-identifier.
I suggest to predefine uuids for our seeds like peter-lustig: 1111-1111111-1111 or simillar.

backend/src/graphql/resolver/UserResolver.ts
```diff
@@ -824,6 +825,13 @@ export class UserResolver {
 
     return true
   }
+
+  @Authorized([RIGHTS.USER])
+  @Query(() => User, { nullable: true })
+  async user(@Arg('identifier') identifier: string): Promise<User | null> {
+    const user = await findUserByIdentifier(identifier)
+    return user ? new User(user) : null
+  }
```
ulfgebhardt, line 834: I suggest to always return a user or throw.
```suggestion
  @Query(() => User)
  async user(@Arg('identifier') identifier: string): Promise<User> {
    return new User(await findUserByIdentifier(identifier))
  }
```

backend/src/graphql/resolver/util/findUserByIdentifier.ts
```diff
@@ -0,0 +1,35 @@
+import { User as DbUser } from '@entity/User'
+import { UserContact as DbUserContact } from '@entity/UserContact'
+import LogError from '@/server/LogError'
+import { validate, version } from 'uuid'
+
+export const findUserByIdentifier = async (identifier: string): Promise<DbUser | null> => {
+  let user: DbUser | undefined
```
ulfgebhardt, line 7: I am a big fan of success case last and error cases before, but here it clearly has disadvantages.

I suggest to treat this like a switch case and return immediately after you have found an user. This will make it easy to reduce the result to `Promise<DbUser>` and simplify the resolver function like suggested above.

backend/src/graphql/resolver/util/findUserByIdentifier.ts
```diff
@@ -0,0 +1,35 @@
...
+    }
+    user = userContact.user
+    user.emailContact = userContact
+  } else {
+    // last is alias when implemented
+    throw new LogError('Unknown identifier type', identifier)
+  }
+
+  return user
+}
```
ulfgebhardt, line 35:
```suggestion
export const findUserByIdentifier = async (identifier: string): Promise<DbUser> => {
  // I would love to see a comment here what each if does, it would help my future self to read this
  // uuid identifier
  if (validate(identifier) && version(identifier) === 4) {
    const user = await DbUser.findOne({ where: { gradidoID: identifier }, relations: ['emailContact'] })
    if (!user) {
      throw new LogError('No user found to given identifier', identifier)
    }
    // this exposes the user email!
    return user
  // I would love to see a comment here what each if does, it would help my future self to read this
  // email identifier
  } else if (/^.{2,}@.{2,}\..{2,}$/.exec(identifier)) {
    const userContact = await DbUserContact.findOne(
      {
        email: identifier,
        emailChecked: true,
      },
      { relations: ['user'] },
    )
    if (!userContact) {
      throw new LogError('No user with this credentials', identifier)
    }
    if (!userContact.user) {
      throw new LogError('No user to given contact', identifier)
    }
    return userContact.user
    // this like would expose the users email - since its know for the user, since he queried this user by email, its not very dangerous, but lets leave out the email completely? Or is this to fast?
    // user.emailContact = userContact
  }

  throw new LogError('Unknown identifier type', identifier)
}
```

ulfgebhardt (changes requested):
image
The backend crashes, I cannot test.

ulfgebhardt (approved):
Sending an User via Identifier works form the transaction list. But I do not have the option to manually input this yet:

image

I believe we can merge this, but should think about how our strategy to allow users send to an identifier aswell instead of an EMail.

PR #2881, opened 2023-03-23
doc(backend): alias rules and conventions

### LINKED ISSUE #2874: 🚀 [Feature] rules and blacklisting for alias

clauspeterhuebner:
### Issues

- None

### Todo

- [X] None


ulfgebhardt (approved):
I think the blacklist might be incomplete, but we will figure that out when implementing this what else to ignore I beleive.

PR #2886, opened 2023-03-24
feat(backend): add event for subscribe and unsubscribe

Elweyn:
### Issues

- bases on #2885
- fixes #2785

### Todo

- [X] Add event for UNSUBSCRIBE_NEWSLETTER
- [X] Add event for SUBSCRIBE_NEWSLETTER

PR #2887, opened 2023-03-24
feat(backend): add fields to subscriber

### LINKED ISSUE #2876: 🚀 [Feature] KlickTipp: klicktipp controller add fields to a user
## 🚀 Feature
As marketing person I need to set some fields to my user to make email campaigns.

## 🤖 Todos
- KlickTipp api has updateSubscriber call were you can update the fields.
  - [ ] Add method to the KlickTippController that use the api call

Elweyn:
### Issues

- based on #2886
- fixes #2876

### Todo

- [X] None


backend/src/apis/KlicktippController.ts
```diff
@@ -75,9 +99,25 @@ export const tagUser = async (email: string, tagIds: string): Promise<boolean> =
...
   return ''
 }
+
+export const addFieldsToSubscriber = async (
+  email: string,
+  fields: any = {},
+  newemail = '',
+  newsmsnumber = '',
+) => {
+  const isLogin = await loginKlicktippUser()
```
Mogge, line 118: This is true in case of no klicktipp active, therefore the following code is called

Elweyn: Yeah I corrected this!

PR #2904, opened 2023-03-29
fix(other): delete node_modules and /tmp/yarn--* on start.sh

ulfgebhardt:
## 🍰 Pullrequest

delete node_modules and /tmp/yarn--* on start.sh

This should fix the error we currently have on production regarding an error on user search (corrupted node_modules) and make the cron job obsolete which clears the tmp folder.

### Issues

- None

### Todo

- [ ] Remove all related cron jobs


clauspeterhuebner (approved):
i preferre to keep the cronjob to ensure low amount of filesystem entries in the /tmp folder.
this solution in the start.sh will only support the cronjob, because i think we had not as much deployments in production that in /tmp exists so many entries. This problem is well known by yarn, but still no prospective solution till now.

PR #2921, opened 2023-04-05
refactor(backend): eslint disable tests typesafer

ulfgebhardt:
### Issues

- None

### Todo

- [X] None


backend/src/graphql/resolver/TransactionResolver.test.ts
```diff
@@ -1,12 +1,11 @@
-/* eslint-disable @typescript-eslint/no-unsafe-call */
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
mahula, line 1: Why does this one stay here?

ulfgebhardt: image

I just investigated the connection stuff - for this test there is seed data variables which I did not adjust/research in and therefore the eslint ignore needs to be in place to pass the test

mahula (approved):

PR #2925, opened 2023-04-05
fix(frontend): moderator id missing

Mogge:
## 🍰 Pullrequest
No more warnings on console when running community spec

### Issues

- None

### Todo

- [X] None


frontend/src/pages/Community.spec.js
```diff
@@ -72,6 +72,7 @@ describe('Community', () => {
                 messagesCount: 0,
                 deniedAt: null,
                 deniedBy: null,
+                moderatorId: 69,
```
Elweyn, line 75: Does this moderatorId make sence?
In my opinion we can have a contributionListItem without moderatorId, in case of state === 'IN_PROGRESS' or 'PENDING'.

Elweyn (commented):
Nice! XD
Only a remark that we maybe want an Item without moderatorId.

frontend/src/pages/Community.spec.js
Mogge: Of course it can be null

Elweyn (approved):

ulfgebhardt (approved):

PR #2901, opened 2023-03-29
feat(admin): visualize the federation

ogerly:
## 🍰 Pullrequest
visualise the federation

- sorted by publicKeys
        oreign: 'ASC',
        createdAt: 'DESC',
        lastAnnouncedAt: 'DESC',

- Displayed 
       verified
      url
      publicKey (0, 26)
      lastAnnouncedAt
      createdAt

The table can be reloaded by hand
 
### Issues
- fixes #2595 
- fixes #2833 
- fixes #1788

image

https://user-images.githubusercontent.com/1324583/228506019-4f5b0eb4-c098-49ef-9c97-37e0cbd17b66.mp4

PR #2931, opened 2023-04-06
fix(federation): correct export of the community url

Elweyn:
## 🍰 Pullrequest

When I look at the Instances page I see the url of the federated communities but there is two times 'api/' 

### Issues

- based on: #2901

### Todo

- [X] None


mahula (dismissed):
Good catch.
I approve.

mahula (approved):

PR #2933, opened 2023-04-11
fix(frontend): no receiver on send by link

### LINKED ISSUE #2932: 🐛 [Bug]
## 🐛 Bugreport
In the transaction list
if you click on a user and then switch to "Send Gradidos via link", the recipient is still visible.
This should be hidden at this point, because links are not sent to the recipient by the system.

send_link_recipient_displayed_bug

Mogge:
### Issues
- fixes #2932 

### Todo

- [X] None


mahula (approved):
This is fixed.
I approve.

PR #2968, opened 2023-05-04
[Obsolete] test(federation): federation tests

ulfgebhardt:
## 🍰 Pullrequest

This PR implements tests to
- check for key comparison failure
- check the success case when validating a community

Furthermore it
- solves a problem with invisible characters caused by buffer size (key length was assumed as half as it is in database)

### Issues

- based on https://github.com/gradido/gradido/pull/2967

### Todo

- [x] The tests on validateCommunities are only based on log outputs not on actual behaviour of the software. I think this is bad practice but would like a second opinion @Mogge (if so make an issue) https://github.com/gradido/gradido/issues/3017


mahula (approved):

ulfgebhardt:
This PR is obsolete, but shows a problem with the current database design. The publicKey has a lenth of 32 bytes, not 64. The database design is wrong and has to be corrected

ulfgebhardt:
closing this as most of this is already part of the code merge from claus peter

PR #2955, opened 2023-04-25
feat(federation): x com 3 introduce business communities

clauspeterhuebner:
### Issues

- None

### Todo

- [x] introduce treatement of .env.devop file to handle DHT_TOPIC, DHT_SEED and KeyPair as separated secrets on production
  - [x] removed the devop solution based on disagreement in the team
- [x] on start DHT the home community entries have to be inserted/updated in database (communities + federated_communties)  

is part of issue #2956 
--------------------------------------------------------------------------
- [ ] create new graphql-handshake to get public-info (name, description, creation_date) of remote community 
**Community-UUID is NO public info and MUST be exchanged in encrypted manner per authentication process later**
  - [ ] graphql-client in backend-module for requesting public community infos
  - [ ] graphql-resolver in federation-module to process the request getPublicCommunityInfo()
- [ ] introduce backend-logic to insert a business-community entry as soon as a federated community is marked as verified_at
- [ ] update the existing business community entry with the recieved community infos

PR #2970, opened 2023-05-05
[Invalid] feat(backend): expose federation keypair to backend

ulfgebhardt:
## 🍰 Pullrequest

expose federation keypair to backend so we can use it to encrypt and sign things.

This is done without importing the dht library, but uses the existing sodium-native lib to generate the seed.

### Issues

- releates https://github.com/gradido/gradido/pull/2975 

### Todo

- [X] None


Elweyn (approved):
Seems good for me, but @clauspeterhuebner please look also at this!

ulfgebhardt:
> defining the key FEDERATION_DHT_SEED additionally in the backend .env-files will cause conflicts between the dht-node definition and possibly different definition here. As in the requirement of this PR described, it is the goal to offer the generated keypair in all modules, which need them:
> 
> * dht-node-modul to take part of the dht-hyperswarm
> * backend-modul to en/crypt and sign future handshake data
> * federation-modul to en/crypt and sign future handshake data
> * devop requirement to separate these Admin-Configs from Developer/Application-Configs
> 
> The same here as mentioned several times: if we want to work as a team, we have to communicate before spending efforts in dupplicate solutions if you had took a look in the #2955 there exists still an implemented solution about this topic, which i discussed with you and Moritz in our technical architecture meeting two week ago.

I dont understand how you provide the private key to the backend in the linked pr #2955 - i need the private key for the upcomming encryption for authentication -> as descirbed in the PRs description

ulfgebhardt:
This PR is invalid - we will solve this differently. I keep this open till the problem is solved

ulfgebhardt:
Superseded by https://github.com/gradido/gradido/pull/3024

PR #2975, opened 2023-05-05
refactor(backend): replace jasonwebtoken with jose

ulfgebhardt:
## 🍰 Pullrequest

As I consider to use JWE for the authentication towards the other communities i switch the `jsonwebtoken` for `jose` which allows JWT and JWE. It is unclear up to now if JWT is sufficient or JWE is required to archieve our goals. The library looks promising tho - hence this test

### Issues

- None

### Todo

- [X] None


Mogge (approved):
Works!

PR #2980, opened 2023-05-10
refactor(backend): eslint security

ulfgebhardt:
## 🍰 Pullrequest

This PR introduces the `eslint-security` plugin, which does code analysis to warn about potential security issues.

While the PR fixes one security issue, it it ignores the remaining 9 since its either not a security concern or its not an easy fix.
The EMail Regex is a potential security flaw we have in our code(could run for long and if utilized correctly it could lead to an ddos attack) - lacking of an immediate solution I decided to ignore this issue.

The benefit is that, when merged, developers will be warned about potential security flaws they introduce with their code.

### Issues

- based on #2979 

### Todo

- [X] None

### DIFF
--- a/backend/src/webhook/elopage.ts
+++ b/backend/src/webhook/elopage.ts
@@ -115,6 +115,7 @@ export const elopageWebhook = async (req: any, res: any): Promise<void> => {
   ) {
     const email = loginElopageBuy.payerEmail
 
+    // eslint-disable-next-line security/detect-unsafe-regex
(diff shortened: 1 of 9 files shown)


backend/src/webhook/elopage.ts
```diff
@@ -115,6 +115,7 @@ export const elopageWebhook = async (req: any, res: any): Promise<void> => {
   ) {
     const email = loginElopageBuy.payerEmail
 
+    // eslint-disable-next-line security/detect-unsafe-regex
```
Elweyn, line 118: Could be a ticket

Elweyn (approved):

PR #2983, opened 2023-05-11
fix(frontend): date fns locales

Mogge:
## 🍰 Pullrequest
There are warnings that the locale en does not exist in the date fns package in the wallet and the 
admin interface. Fixed it

### Issues

- None

### Todo

- [X] None


ulfgebhardt (approved):
Build the thing, could not find error on build regarding missing export `en` anymore - software works(frontend & admin) as expected, language can be changed - things seem to work.

PR #2988, opened 2023-05-16
feat(backend): send coins via alias

Mogge:
## 🍰 Pullrequest
Allows to send coins with alias as identifier

Watch out: __based on #2984__

### Issues

- is there one?

### Todo

- [X] None


backend/src/graphql/resolver/TransactionResolver.ts
```diff
@@ -323,6 +323,7 @@ export class TransactionResolver {
     }
 
     // TODO this is subject to replay attacks
+    // --- WHY?
```
ulfgebhardt, line 326: This todo is still valid.

If you catch the request bob sends to the server and just send it again, the transaction is executed again.
We actually had this in our application when we didnt block the send button before executing the transaction. A very fast double klick had the transaction executed twice (tho this is not a replay attack).

A possible solution is to request a new transactionRequestToken, which can be used only once for a transaction.
If the transaction token is used, you cannot reuse it and have to request a new one to send another transaction.

This is overhead for very little benefit, but potential high security value. Therefore we skipped this in the past and decided against an implementation.

The semaphore does not solve this issue, since the server receives both calls and just waits (using the semaphore) before executing the second one - so this is not solved by semaphore(tho a semaphore for the transactionRequestToken would probably be needed)

Feel free to remove the `TODO` and make an issue for it instead.

backend/src/graphql/resolver/util/validateAlias.ts
```diff
@@ -3,6 +3,9 @@ import { User as DbUser } from '@entity/User'
 
 import { LogError } from '@/server/LogError'
 
+// eslint-disable-next-line security/detect-unsafe-regex
+export const validAliasRegex = /^(?=.{3,20}$)[a-zA-Z0-9]+(?:[_-][a-zA-Z0-9]+?)*$/
```
ulfgebhardt, line 7: i would write this in caps since its a constant, but thats a minor non blocking thing
```suggestion
export const VALID_ALIAS_REGEX = /^(?=.{3,20}$)[a-zA-Z0-9]+(?:[_-][a-zA-Z0-9]+?)*$/
```

ulfgebhardt (commented):
Besides the `// --- WHY?` I am all cool with this.

backend/src/graphql/resolver/TransactionResolver.ts
ulfgebhardt: Remove the `Why?`

ulfgebhardt (approved):

PR #2993, opened 2023-05-18
refactor(federation): refactor get public key query

ulfgebhardt:
## 🍰 Pullrequest

This PR simplifies the getPublicKey query to just return a string. This simplifies the query, but is a breaking change. Therefor a (temporary) API Version 2_0 is created which reflects this properly.

### Issues

- based on https://github.com/gradido/gradido/pull/2992

### Todo

- [ ] Test this properly


clauspeterhuebner (changes requested):
I will not agree with changing the versioning of backend/federation/client from directory-based to file-based, because:
- it differs to the versioning in federation/graphql/api
- the version dependent grouping must be searched in different directories per file-postfix
- keeping all version specific api methods in one client-file will not support parallel team working on separate business cases of the same version
- using inheritance is no argument to change versioning from directory-based to file-based

deployment/bare_metal/.env.dist
```diff
@@ -68,7 +68,7 @@ FEDERATION_COMMUNITY_API_PORT=5000
 
 FEDERATION_CONFIG_VERSION=v1.2023-01-09
 # comma separated list of api-versions, which cause starting several federation modules
-FEDERATION_COMMUNITY_APIS=1_0,1_1,2_0
+FEDERATION_COMMUNITY_APIS=1_0,1_1
```
clauspeterhuebner, line 71: it doesn't make sense to keep only the version 1_1 next to 1_0 even version 2_0 has real differences but not only a different version-nr.

federation/src/graphql/api/1_1/resolver/PublicKeyResolver.ts
```diff
@@ -0,0 +1,8 @@
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import { Resolver } from 'type-graphql'
+// eslint-disable-next-line camelcase
+import { PublicKeyResolver as PublicKeyResolver_1_0 } from '../../1_0/resolver/PublicKeyResolver'
```
clauspeterhuebner, line 4: i thought relative paths are not allowed in imports?

clauspeterhuebner (changes requested):
As i mentioned several times and in the last review:

I will not agree with changing the versioning of backend/federation/client from directory-based to file-based, because:
* it differs to the versioning in federation/graphql/api
* the version dependent grouping must be searched in different directories per file-postfix
* keeping all version specific api methods in one client-file will not support parallel team working on separate business cases of the same version
* using inheritance is no argument to change versioning from directory-based to file-based

deployment/bare_metal/.env.dist
Elweyn: Don't we have the version number test covered through the ``1_0`` and we tests the minor version upgrade with the ``1_1``? And we had in the proof that ``2_0`` would work.
For me this ``1_0,1_1`` solution would be good!

federation/src/graphql/api/1_1/resolver/PublicKeyResolver.ts
Elweyn: Let's create a Ticket for this!

federation/src/graphql/schema.ts
```diff
@@ -2,15 +2,15 @@ import { GraphQLSchema } from 'graphql'
 import { buildSchema } from 'type-graphql'
 
 // import isAuthorized from './directive/isAuthorized'
-import DecimalScalar from './scalar/Decimal'
-import Decimal from 'decimal.js-light'
+// import DecimalScalar from './scalar/Decimal'
+// import Decimal from 'decimal.js-light'
```
Elweyn, line 6: Why do you comment this? Do we need it in the future?
If not:
```suggestion
```

federation/src/graphql/schema.ts
```diff
@@ -2,15 +2,15 @@ import { GraphQLSchema } from 'graphql'
...
 import { getApiResolvers } from './api/schema'
 
 const schema = async (): Promise<GraphQLSchema> => {
   return await buildSchema({
     resolvers: [getApiResolvers()],
     // authChecker: isAuthorized,
-    scalarsMap: [{ type: Decimal, scalar: DecimalScalar }],
+    // scalarsMap: [{ type: Decimal, scalar: DecimalScalar }],
```
Elweyn, line 13: Same question will we need it in the future?
Else
```suggestion
```

Elweyn (changes requested):

deployment/bare_metal/.env.dist
clauspeterhuebner: A single version-number for its own will not ensure the treatement of several different versions.
Now to remove the version 2_0, which really differes from 1_0 and 1_1 (the version 1_1 has no differenc against the version 1_0, it simply shows a solution how to avoid duplicate code if there are no changes against a previous version) seems not like an intelligent solution, because it is impossible to test against real different apis and not only different version-numbers.

federation/src/graphql/api/1_1/resolver/PublicKeyResolver.ts
clauspeterhuebner: good idea

Elweyn: - [ ] 🚀 [Feature] Federation: introduce linting rules

Elweyn: #3064

Elweyn (dismissed):

Elweyn (approved):

clauspeterhuebner:
not agreed refactoring, so the PR is closed

PR #2994, opened 2023-05-19
feat(frontend): preserve email after login

Mogge:
### Issues

- None

### Todo

- [X] None


Elweyn (approved):
Like @ulfgebhardt said in the meeting we don't have the email when coming back from the admin interface!
I make a ticket for this scenario but like we said is OK for now!

ulfgebhardt (commented):
As prognosed the email is missing when I come from the admin interface
image

PR #3019, opened 2023-05-30
refactor(federation): removed (potential) duplicate db query

ulfgebhardt:
## 🍰 Pullrequest

Removed (potential) duplicate db query.

This simplifies the code, removes one potential DB Query while it has the same outcome.

### Issues

- None

### Todo

- [X] None


clauspeterhuebner (changes requested):
the motivation of this PR is not correct.
the resultset of both queries is different:
- the first one search for an existing entry with the same public key
- the second one search for an existing entry of a home-community but with different publicKey
Depending on the resultset an update (with keeping the id/pubKey/uuid) or an insert (new id/pubkey/uui) will be done.
That's a behaviour of running an instance without a seed like we do during development.

ulfgebhardt:
You need to explain this to me - the code does the same as before in my opinion

clauspeterhuebner:
For sure we can talk about this topic, but please take a deeper look at the comments and the existing logic.
Imaging - like it is commented - a home-community-entry exists, but it has a different publickey than expected (possible by starting dht without a seed).

On my opinion it will be more productiv not to focus on refactoring existing code before collecting some experiences by working and understanding the logic behind the code. 
This seems to be a good example that your first view is focused to reduce perceived doublicate or inefficient code without a full and deeper understanding of the logic by working with it. Afterwards with real experiences every optimization is wellcome.

clauspeterhuebner (approved):

PR #3022, opened 2023-05-30
refactor(federation): remove function getSeed

ulfgebhardt:
## 🍰 Pullrequest

remove function getSeed, since its only used once and a simple ternary.

### Issues

- None

### Todo

- [X] None


clauspeterhuebner (approved):
From my point of view the previous modularity with the separated function getSeed() was more readable than the current construct. At least a comment, what happens in this construct will be nice.
But its your turn.

ulfgebhardt:
I included a comment

clauspeterhuebner (approved):

PR #3023, opened 2023-05-30
refactor(federation): expose private key to write home community

ulfgebhardt:
## 🍰 Pullrequest

Expose private key to writeHomeCommunity functionality. This is required to write the key to the database.

### Issues

- based on #3022 
- based on #3019 

### Todo

- [X] None


clauspeterhuebner (dismissed):
Ok, here i can see your kind of working.
I would never extend an api in backend for write-access without an existing entity, which supports this column.
So for me the headline of the PR is a little confusing... But its ok to see this other kind of working.

dht-node/src/dht_node/index.ts
```diff
@@ -29,14 +24,20 @@ export type CommunityApi = {
   url: string
 }
 
+type KeyPair = { publicKey: Buffer; secretKey: Buffer }
+
 export const startDHT = async (topic: string): Promise<void> => {
   try {
     const TOPIC = DHT.hash(Buffer.from(topic))
-    const keyPair = DHT.keyPair(getSeed())
+    const keyPair = DHT.keyPair(
```
clauspeterhuebner, line 32: it seems not uptodate with the master, the comment about the logic is missing here

clauspeterhuebner (approved):

dht-node/src/dht_node/index.ts
ulfgebhardt: image

Elweyn (approved):

clauspeterhuebner (approved):

PR #3026, opened 2023-05-30
refactor(database): fix database public key lengths

ulfgebhardt:
## 🍰 Pullrequest

Corrects the key length of the public key to 32 instead of 64 bytes, since thats the length generated by the DHT library.

### Issues

- None

### Todo

- [ ] Has to be tested


clauspeterhuebner (dismissed):
What about the test-cases, which uses publicKeys with the full lenght of 64?

clauspeterhuebner (approved):

Mogge (approved):

Elweyn (approved):

PR #3035, opened 2023-06-07
refactor(backend): removed klicktipp middleware and replace it with a function

ulfgebhardt:
## 🍰 Pullrequest

removed klicktipp middleware and replace it with a function.

Further refactoring would make the klicktipp subscribe state as a own entity resolving with its own resolver if requested. Since we do not use graphql in that way, we have to explicitly define the klicktipp state on a user when requested manually.

### Issues

- None

### Todo

- [x] Test if klicktipp subscribe state is still shown properly


clauspeterhuebner (commented):
Is this PR coordinated with @Elweyn klicktipp-topic?
an uncoordinated refactoring on a WIP-topic is not nice and no team-working.

Mogge (approved):

Elweyn (approved):
Thanks for the refactoring!

PR #3044, opened 2023-06-09
refactor(dht): eslint dht import

ulfgebhardt:
## 🍰 Pullrequest

Enables all eslint import rules

### Issues

- based on https://github.com/gradido/gradido/pull/3042

### Todo

- [X] None


.github/workflows/test_dht_node.yml
```diff
@@ -51,7 +51,7 @@ jobs:
         uses: actions/checkout@v3
       
       - name: Lint
-        run: cd dht-node && yarn && yarn run lint
+        run: cd database && yarn && cd ../backend && yarn && yarn run lint
```
Elweyn, line 54:
```suggestion
        run: cd dht-node && yarn && yarn run lint && cd ../database && yarn && yarn run lint && cd ../backend && yarn && yarn run lint
```

Elweyn (changes requested):
Not totally sure if the suggestion is appropriate but I think that if you want to lint you lint everything: dht, backend and database

.github/workflows/test_dht_node.yml
ulfgebhardt: This line makes sure the database has all required `node_modules` and built files so testing/linting works properly.

Only the current project (e.g. `dht_node`) is linted.

ulfgebhardt, line 54:
```suggestion
        run: cd database && yarn && cd ../dht-node && yarn && yarn run lint
```

ulfgebhardt: To you are right - there was a mistake in this like - it should be `cd ../dht-node`

Elweyn (approved):

PR #3060, opened 2023-06-14
refactor(backend): call yarn upgrade and fix all tests and jest warnings

einhornimmond:
## 🍰 Pullrequest
With a call of 
```bash
yarn install --audit
```
there where 58 vulnerabilities found:
Severity: 6 Moderate | 39 High | 13 Critical

So this is the result of 
```bash
yarn upgrade
```
and fixing tests and linter which weren't longer compatible with updated node modules

[ ] fix End-To-End Test (not longer compatible with updated modules I think)

### DIFF
--- a/backend/src/graphql/resolver/ContributionMessageResolver.ts
+++ b/backend/src/graphql/resolver/ContributionMessageResolver.ts
@@ -56,7 +56,9 @@ export class ContributionMessageResolver {
       contributionMessage.isModerator = false
       await queryRunner.manager.insert(DbContributionMessage, contributionMessage)
 
-      if (contribution.contributionStatus === ContributionStatus.IN_PROGRESS) {
+      if (
+        (contribution.contributionStatus as ContributionStatus) === ContributionStatus.IN_PROGRESS
+      ) {
         contribution.contributionStatus = ContributionStatus.PENDING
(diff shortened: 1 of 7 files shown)


backend/src/graphql/resolver/ContributionMessageResolver.ts
```diff
@@ -56,7 +56,9 @@ export class ContributionMessageResolver {
       contributionMessage.isModerator = false
       await queryRunner.manager.insert(DbContributionMessage, contributionMessage)
 
-      if (contribution.contributionStatus === ContributionStatus.IN_PROGRESS) {
+      if (
```
Mogge, line 59: Why?

backend/src/graphql/resolver/ContributionMessageResolver.ts
```diff
@@ -137,10 +139,11 @@ export class ContributionMessageResolver {
       contributionMessage.isModerator = true
       await queryRunner.manager.insert(DbContributionMessage, contributionMessage)
 
+      const contributionStatusEnum = contribution.contributionStatus as ContributionStatus
```
Mogge, line 142: Why?

Mogge (changes requested):
You changed the `yarn.lock` by doing things on your machine ...
This is not helpful on an open source project. If you feel the need to update packages, please do this in the `package.json`

PR #2991, opened 2023-05-17
[POC] Authorization

ulfgebhardt:
## 🍰 Pullrequest

[POC] Authorization

This is pseudo code to show the get go of the idea for authorization and the basis for authentication (or vice versa?).

This code does not compile or lint - it is to have a basis for discussion if this is the right strategy.

### Issues

- None

### Todo

- [X] None

PR #3066, opened 2023-06-16
feat(federation): authentication

### LINKED ISSUE #3065: 🚀 [Feature] Federation: Authenticate Federation and Backend per Keys
## 🚀 Feature

- [ ] Add public key to the ``community`` and ``federatedCommunity`` tables #3026 
- [ ] Add private key to ``community`` and ``federatedCommunity`` tables #3023 #3024 
- [ ] Authenticate between Federation and Backend

ulfgebhardt:
## 🍰 Pullrequest

This implements the authentication of a client as base for the authorization of a client towards the federation endpoint.
The server verifies the sent token, decrypts it and responds with his own token reencrypting the provided nonce. The client then verifies this nonce after another decryption.

### Issues

- replaces https://github.com/gradido/gradido/pull/2991
- fixes #3065 

### Todo

- [x] Get things running with typechecks, linting etc
- [x] Validate response on client
- [x] Assign authenticated middleware to getPublicKey request
- [x] Test functionality
- [ ] Write Test


federation/src/auth/JWE.ts
```diff
@@ -0,0 +1,103 @@
...
+const crypto_scalarmult_curve25519_BYTES = 32
+
+export const verifyToken = async (
+  token: string,
+  keyPair: { publicKey: Buffer; privateKey: Buffer },
+  nonce: number | null = null,
+) => {
+  const pubKeyX = Buffer.alloc(crypto_scalarmult_curve25519_BYTES)
+  const privKeyX = Buffer.alloc(crypto_scalarmult_curve25519_BYTES)
+  crypto_sign_ed25519_pk_to_curve25519(pubKeyX, keyPair.publicKey.subarray(0, 32))
```
clauspeterhuebner, line 19: Please use the constant definition from the used library instead of a hardcoded value of 32

federation/src/auth/JWE.ts
```diff
@@ -0,0 +1,103 @@
...
+  crypto_sign_ed25519_sk_to_curve25519(privKeyX, keyPair.privateKey.subarray(0, 32))
```
clauspeterhuebner, line 20: Please use the constant definition from the used library instead of a hardcoded value of 32

federation/src/auth/JWE.ts
```diff
@@ -0,0 +1,103 @@
...
+      }),
+    ),
+  )
+    .setProtectedHeader({ alg: 'EdDSA' })
+    .sign(key)
+  return await new CompactEncrypt(
+    new TextEncoder().encode(
+      JSON.stringify({
+        jws,
+        publicKey: keyPair.publicKey.subarray(0, 32),
```
clauspeterhuebner, line 97: Please use the constant definition from the used library instead of a hardcoded value of 32

federation/src/auth/JWE.ts
```diff
@@ -0,0 +1,103 @@
...
+      }),
+    ),
+  )
+    .setProtectedHeader({ alg: 'ECDH-ES', enc: 'A256GCM' })
+    .encrypt(foreignPub)
+}
```
clauspeterhuebner, line 103: If this file JWE.ts is the same as in backend/src/federation/auth, what do you think about the idea to create a kind of common modul, to insert there all general parts of code, which will be used from several moduls?
It allows us to keep the code only once, but use it multi in several moduls.
I think we still have or at least we will have in future more than this single file, which will be used from several moduls. One example for multi usage will be the DAO-layer classes, which could be usefull for backend- and federation-modul like it is for this JWE file too.

clauspeterhuebner (changes requested):
good work, but some small questions and comments

a general question about the authentification:
As i interprete your code, there is no prevention about a man-in-the-middle scenario like it is solved by the OAUTH2.0 + OIDC protocol, right?

federation/src/auth/JWE.ts
ulfgebhardt: I am a big fan of removing duplication, but we need to find a strategy to do this - its nothing we can just do - we need to decide an approach and do the corresponding refactoring.
I do not think it to be wise to do a rush job on a system which impacts alot of subsystems of ours.

For further read I'd like to link the following:
- https://dev.to/michalbryxi/share-common-code-with-yarn-workspaces-5g29

We have such a shared codebase with `database` with all its up and downsides.

Main problems I see are:
- Frontend and Backend use different javascript flavors (js, ts)
- Scope of a shared module must be well defined in order to prevent all packages to diffuse into the common module(s)
- Approach to use (workspaces, tsconfig includes, npm packages) must be chosen wisely based on shared package

We could make an epic for this - but I do not want to solve the problem in this PR.

federation/src/graphql/directive/isAuthorized.ts
```diff
@@ -0,0 +1,66 @@
+import { Community as dbCommunity } from '@entity/Community'
+import { AuthChecker } from 'type-graphql'
+
+import { verifyToken, generateToken } from '@/auth/JWE'
+import { RIGHTS } from '@/auth/RIGHTS'
+import { ROLE_UNAUTHORIZED, ROLE_AUTHORIZED } from '@/auth/ROLES'
+import { Context } from '@/server/context'
+import { LogError } from '@/server/LogError'
+
+export const isAuthorized: AuthChecker<Context> = async ({ context }, rights) => {
```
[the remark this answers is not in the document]
ulfgebhardt: The `@authorized` enforces an array of Rights - this cannot be changed and is part of the library definition.

federation/src/graphql/directive/isAuthorized.ts
```diff
@@ -0,0 +1,66 @@
...
+      where: { publicKey: debugComKey },
+    })
+    // todo do not respond with token below?
+    if (community) {
+      context.community = community
+      context.role = ROLE_AUTHORIZED // TODO here authorization has to be checked
+    }
+
+    // check for correct rights
+    const missingRights = (rights as RIGHTS[]).filter((right) => !context.role?.hasRight(right))
```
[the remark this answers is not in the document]
ulfgebhardt: See above

federation/src/server/context.ts
```diff
@@ -0,0 +1,30 @@
...
+    setHeaders: [],
+  }
+  if (authorization) {
+    context.token = authorization.replace(/^Bearer /, '')
+  }
+  return context
+}
+
+export const getUser = (context: Context): dbCommunity => {
+  if (context.community) return context.community
```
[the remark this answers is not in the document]
ulfgebhardt: The lint rule might not be implemented yet on federation, just the backend.

ulfgebhardt:
> good work, but some small questions and comments
> 
> a general question about the authentification: As i interprete your code, there is no prevention about a man-in-the-middle scenario like it is solved by the OAUTH2.0 + OIDC protocol, right?

There is a timestamp - which is currently not checked.
In general this technic is most likely atackable with a replay attack - repeat the sent package and have a valid request.

I believe it is sufficient for now and analog to our `frontend <-> backend` communication. So once we decide to do the research what to use - openId/OAuth/... we can properly implement it in between frontend and backend aswell as between servers.

The point of this approach is more how to do authentication/authorization rather then actual security. I want to prove that a server found via DHT/Federation can prove its identity. This is very important since I believe if we cannot prove authentication is possible, DHT/Federation will not work, since then a culprit can just create gradido on any community. Hence its crucial we can identify a source of coins and prevent them to enter our ecosystem.

This PR has proven that this is possible, even tho not 100% foolproof on the security side of things, but that topic is for another time. The advantage with this approach is no state management. Downside is potential man-in-the-middle replay attack and CPU usage.

federation/src/graphql/directive/isAuthorized.ts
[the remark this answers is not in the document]
clauspeterhuebner: that's ok for me, i only recognized the check for an array of rights even we have only one right per method. If this is part of the library api, we should not change it.

clauspeterhuebner:
this issue is solved by 
- #3505
- #3516
so this ticket is closed

PR #3082, opened 2023-06-27
refactor(other): refactor state to status

### LINKED ISSUE #2671: Could this be status? In the database it is called status, so it would be nice, not to change the name
we will have an extra adminListAllContribution in the near future ... there you could adjust this.               
              
              Could this be status? In the database it is called status, so it would be nice, not to change the name

_Originally posted by @Mogge in https://github.com/gradido/gradido/pull/2623#discussion_r1097605338_

Elweyn:
### Issues

- fixes #2671

### Todo

- [X] Change state to status


ulfgebhardt (approved):
Tested
- submit creation
- send question
- answer question
- confirm creation

All works flawlessly, please merge

Issue #3110, opened 2023-07-03
🚀 [Feature][DLT-Connector] trigger and send transaction-info to dlt-connector


Elweyn:
- [ ] database migration: 
   - new table dlt_transactions
      - id
      - transactions_id
      - message_id
      - community_balance
      - community_balance_date
      - created_at
      - verified_at
- [ ] send transaction-info to dlt-connector
      - create graphql client for send-request
      - take response of request and writ in dlt-transaction entry
- [ ] insert trigger on transaction creations
      - to create dlt_transactions entry
      - to send dlt-transaction-request
      - to calculate community_balance

PR #3132, opened 2023-07-07
feat(other): iota-tangle-connector sending transaction

einhornimmond:
## 🍰 Pullrequest

- graphql basic stuff copied from backend
  - updated to new graphql version
- graphql resolver for transactions
- pack transaction with protobuf  

### Issues
- relates #3110

### Todo

- [x] graphql resolver for transactions
- [x] test resolver


dlt-connector/src/proto/TransactionBody.ts
```diff
@@ -0,0 +1,19 @@
+import { Decimal } from 'decimal.js-light'
+import { TransactionType } from '../graphql/enum/TransactionType'
+import { Field, Message } from '@apollo/protobufjs'
```
clauspeterhuebner, line 3: why using protobuf instead of type-graphql like the other graphql-models of the other modules?

dlt-connector/src/server/createServer.ts
```diff
@@ -0,0 +1,55 @@
...
+  })
+  await apollo.start()
+  app.use(
+    '/',
+    cors<cors.CorsRequest>(),
+    bodyParser.json(),
+    expressMiddleware(apollo, {
+      context: async ({ req }) => ({ token: req.headers.token }),
+    }),
+  )
```
clauspeterhuebner, line 49: what's the background for your changed creation-logic, because it looks quite different than the backend createServer?

dlt-connector/tsconfig.json
```diff
@@ -48,7 +48,13 @@
     "baseUrl": ".",                                 /* Base directory to resolve non-absolute module names. */
     "paths": {                                      /* A series of entries which re-map imports to lookup locations relative to the 'baseUrl'. */
       "@/*": ["src/*"],
+      "@arg/*": ["src/graphql/arg/*"],
+      "@enum/*": ["src/graphql/enum/*"],
+      "@input/*": ["src/graphql/input/*"],
+      "@resolver/*": ["src/graphql/resolver/*"],
+      "@scalar/*": ["src/graphql/scalar/*"],
       "@test/*": ["test/*"],
+      "@proto/*" : ["src/proto/*"],
```
clauspeterhuebner, line 57: why do you store the model for the graphql-api outside the graphql-directory?

clauspeterhuebner (approved):
...with several comments and little optional changes...

dlt-connector/src/graphql/input/TransactionInput.ts
```diff
@@ -0,0 +1,21 @@
+// https://www.npmjs.com/package/@apollo/protobufjs
+
+import { Decimal } from 'decimal.js-light'
+import { TransactionType } from '../enum/TransactionType'
+import { InputType, Field } from 'type-graphql'
+
+@InputType()
+export class TransactionInput {
+  @Field(() => TransactionType)
```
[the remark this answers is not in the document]
einhornimmond: No, without transaction type the transaction is invalid.

dlt-connector/src/graphql/input/TransactionInput.ts
```diff
@@ -0,0 +1,21 @@
...
+  @Field(() => Decimal)
+  amount: Decimal
+
+  @Field(() => Number)
+  created: number
+
+  // @protoField.d(4, 'string')
+  // @Field(() => Decimal)
+  // communitySum: Decimal
+}
```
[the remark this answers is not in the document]
einhornimmond: communitySum is the next additional attribute, but for the calculation I probably need a extra Pull-Request.
For the other attributes, I'm not sure when I will add which and which won't be added at all. 
I must think about it and maybe we must talk a bit about it.

dlt-connector/src/graphql/schema.ts
```diff
@@ -0,0 +1,13 @@
+import { Decimal } from 'decimal.js-light'
+import { GraphQLSchema } from 'graphql'
+import { buildSchema } from 'type-graphql'
+
+import { DecimalScalar } from './scalar/Decimal'
+import { TransactionResolver } from './resolver/TransactionsResolver'
+
+export const schema = async (): Promise<GraphQLSchema> => {
+  return buildSchema({
+    resolvers: [TransactionResolver],
```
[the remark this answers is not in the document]
einhornimmond: I use the current Apollo Version 4, not the deprecated Apollo Version 2 which is used from the rest project.
https://www.apollographql.com/docs/apollo-server/v2
Apollo Server 2 is officially deprecated, with end-of-life scheduled for 22 October 2023. Additionally, certain features are end-of-life as of 31 December 2023

In the current Apollo Version a string array is not longer allowed for BuildSchemaOptions.resolvers only NonEmptyArray<Function>
so your code don't longer work. I have copied schema.ts from backend and had to adjust it, because it don't longer work.

dlt-connector/src/server/createServer.ts
einhornimmond: I use the current Apollo Version 4, not the deprecated Apollo Version 2 which is used from the rest project.
https://www.apollographql.com/docs/apollo-server/v2
Apollo Server 2 is officially deprecated, with end-of-life scheduled for 22 October 2023. Additionally, certain features are end-of-life as of 31 December 2023

I have copied createServer.ts from federation and had to adjust it, because it don't longer work.

dlt-connector/src/proto/TransactionBody.ts
einhornimmond: I use protobuf for serializing the object before sending it to iota.
- signing is easier
- the iota message is shorter as if I would use json for example (it seems that the Proof-of-Work is harder as longer the message is)
- the iota message on the tangle is harder to read for people without the right software, the transaction from the test would look like this on tangle explorer: 08021202313018b4f5afa506
in json:
```json
{type: 'SEND', amount: '10', createdAt: 1688992436}
```

dlt-connector/tsconfig.json
einhornimmond: because this model represent the gradido transaction on the tangle, so it don't really belong to the graphql-api.

dlt-connector/src/proto/TransactionBody.ts
clauspeterhuebner: if an additional model will offer more advantages than the mapping to and from this model, than feel free to use it. 
But keep in mind, that an additional mapping and conversion between different models will cost additional resources - memory, cpu, garbagecollection,... - at runtime each time of processing.

Additional efforts during coding phase, because of unlikely inconvinient code, will cause these higher efforts only once for development. At runtime it could be more efficient than a simply looking code, which hides its complexity only from surface.

dlt-connector/src/server/createServer.ts
clauspeterhuebner: very nice, than you could collect experiences with this new version, so that your knowledge could help us for upgrading, which will coming soon , i gues...

clauspeterhuebner (approved):
i approve it again, but add some remarks too... ;-)

dlt-connector/src/graphql/input/TransactionInput.ts
```diff
@@ -0,0 +1,21 @@
+// https://www.npmjs.com/package/@apollo/protobufjs
+
+import { Decimal } from 'decimal.js-light'
+import { TransactionType } from '../enum/TransactionType'
+import { InputType, Field } from 'type-graphql'
+
+@InputType()
+export class TransactionInput {
```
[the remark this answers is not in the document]
einhornimmond: In graphql you cannot use ObjectType as Input:
https://graphql.org/learn/schema/#input-types
here is stated why: 
https://spec.graphql.org/October2016/#sec-Input-Objects
If you look in the backend code you will notice that the model Objects are only used as output and for input @ArgsType and @InputTypes in folder /arg are used.

Why I use input rather than arg?
Input is used for mutations, args for querys and I don't know if I even will have a single query.

clauspeterhuebner (approved):

Issue #3031, opened 2023-05-30
🚀 [Feature] role administration (frontend)


clauspeterhuebner:
## 🚀 Feature: introduce the "moderator"-role

- [ ] disable/hide the users-role GUI components for users not having ADMIN role
- [ ] only allow users-role GUI components fur ADMIN-users
- [ ]

Issue #3030, opened 2023-05-30
🚀 [Feature] role administration (backend)


clauspeterhuebner:
## 🚀 Feature: introduce the "moderator"-role

- [ ] adapt the database model with a relation of roles per user (use existing ENUM as db-values)
- [ ] modify the backend user-service to set/remove a user-role (only allowed for ADMIN-users)
- [ ] refactor the roles and rights definitions to introduce the role MODERATOR
- [ ] write (define a new type if missing) event if a users-role was changed
- [ ] adapt backend-logic on new MODERATOR-allowence distingue from ADMIN
  - [ ] ADMIN can do all of MODERATOR
  - [ ] ADMIN can do additionally:
    - [ ] set/remove a users-role
    - [ ] ?

PR #3141, opened 2023-07-10
feat(merge): Master moderator role select

ogerly:
## 🍰 Pullrequest
 merge  #3031 an #3030 

### Issues
- merge #3031 #3030


admin/src/pages/UserSearch.spec.js
```diff
@@ -208,25 +208,25 @@ describe('UserSearch', () => {
...
-          expect(wrapper.vm.searchResult.find((obj) => obj.userId === userId).isAdmin).toEqual(
-            expect.any(Date),
-          )
-        })
-      })
+      // describe('to admin', () => {
+      //   it('updates user role to admin', async () => {
+      //     await wrapper
+      //       .findComponent({ name: 'SearchUserTable' })
+      //       .vm.$emit('updateIsAdmin', userId, new Date())
```
clauspeterhuebner, line 215: update or setting a user role per setUserRole is now using a parameter of type string instead of a date.

admin/src/pages/UserSearch.spec.js
```diff
@@ -208,25 +208,25 @@ describe('UserSearch', () => {
...
+      //     expect(wrapper.vm.searchResult.find((obj) => obj.userId === userId).isAdmin).toEqual(
+      //       expect.any(Date),
```
clauspeterhuebner, line 217: the setUserRole directly returns the new role-name, so it could be optimized to avoid invoking an additional searchUser to check for the new role setting

admin/src/pages/UserSearch.spec.js
```diff
@@ -208,25 +208,25 @@ describe('UserSearch', () => {
...
-            .vm.$emit('updateIsAdmin', userId, null)
-          expect(wrapper.vm.searchResult.find((obj) => obj.userId === userId).isAdmin).toEqual(null)
-        })
-      })
+      // describe('to usual user', () => {
+      //   it('updates user role to usual user', async () => {
+      //     await wrapper
+      //       .findComponent({ name: 'SearchUserTable' })
+      //       .vm.$emit('updateIsAdmin', userId, null)
+      //     expect(wrapper.vm.searchResult.find((obj) => obj.userId === userId).isAdmin).toEqual(null)
```
clauspeterhuebner, line 227: the setUserRole directly returns the new role-name, so it could be optimized to avoid invoking an additional searchUser to check for the new role setting.
In case of setting the role=user the setUserRole have to be called with the role-parameter = null, because a user without a userrole will be treated as a normal user.

clauspeterhuebner (changes requested):
Please see my comments after a first review of your solution.

Mogge:
This was done in another PR

PR #3143, opened 2023-07-14
feat(backend):  use iota-tangle-connector

einhornimmond:
## 🍰 Pullrequest

Call DLT-Connector from backend when sending GDD to another account or confirm a contribution.

### Details
- change default port from DLT-Connector from 6000 (used by X-Windows) to 6010
- add option for backend to switch on/off using of dlt-connector and the target url
- add dltTransactionId and dltConfirmed fields to Transaction db model and migration for new db fields
- add apis/DltConnector with transmitTransaction function which will send the transaction as graphql request to the dlt-connector
- add test for apis/DltConnector 
- call transmitTransaction in TransactionResolver and ContributionResolver

### Issues
- relates #3110


database/entity/0069-add_dlt_message_id/Transaction.ts
```diff
@@ -0,0 +1,155 @@
...
+    name: 'linked_transaction_id',
+    type: 'int',
+    unsigned: true,
+    nullable: true,
+    default: null,
+  })
+  linkedTransactionId?: number | null
+
+  @Column({
+    name: 'dlt_transaction_id',
```
clauspeterhuebner, line 134: whats the business background of this id?
Is it a foreign key to somewhere?
What could the application do with this information later?

So i gues this will cause some conceptuell discussions

database/entity/Transaction.ts
```diff
@@ -1 +1 @@
-export { Transaction } from './0066-x-community-sendcoins-transactions_table/Transaction'
```
clauspeterhuebner, line 1: I think you base of work is outdated, because the master branch contains the following version:
0068-community_tables_public_key_length

clauspeterhuebner (changes requested):
I started with the review and saw some basic technical aspects we have to discuss:
- after a transaction write process you need to invoke a graphql-client like in backend/federeation/client
- your DltConnector seems not to implement a graphql-client, but a graphql-resolver as a mutation
- a graphqlresolver has to follow the graphql-spec by several formal aspects
  - naming patterns
  - path location
  - registration in schema
  - a resolver as inbound api MUST be safe per security technics per roles and rights
  - the data per graphql have to follow strict rules to be validate by the graphql-schema-validation
  - and further such topic

so let us meet for a discussion about conceptual and technical solutions

database/entity/0069-add_dlt_message_id/Transaction.ts
einhornimmond: It is for the iota message id. But as we have named the iota-connector dlt-connector I have also named this dlt_transaction_id rather than iota_transaction_id or iota_message_id.
The iota message id is the easiest way for checking if the transaction is on the tangle, or on a chronicle.
Currently I also plan to use this id for notification the backend if the transaction was 'confirmed' on the tangle.

einhornimmond:
> - after a transaction write process you need to invoke a graphql-client like in backend/federeation/client 
- I invoke graphql-client after transaction was written to db
- I have updated it so it will be only one instance
- Or do you mean with Versioning? 
  https://graphql.org/learn/best-practices/#versioning
  GraphQL suggest best to use a non-version API
- as dlt-connector and backend are in the same mono-repository versioning didn't make sense at all

> - the transaction writing is quite tricky, because of several used semaphores for synchronisation, which has to take in acount

And how exactly would the semaphore influence the usage of dlt-connector?

> - your DltConnector seems not to implement a graphql-client, but a graphql-resolver as a mutation

The dlt-connectore receive the request from backend, in order to do that it need a graphql-resolver
Client -> send  |  receive -> GraphQL Resolver

einhornimmond:
> - a graphqlresolver has to follow the graphql-spec by several formal aspects
> - naming patterns

### chat GPT: 
- The provided GraphQL schema appears to follow the naming patterns commonly used in GraphQL. Here are some observations about the schema:
- Type names: The type names (Decimal, Mutation, Query, TransactionInput, TransactionType, and TransmitTransactionResult) use PascalCase, which is a recommended convention for type names in GraphQL.
- Field names: The field names (amount, createdAt, type, and dltTransactionIdHex) use camelCase, which is also a commonly used convention for field names in GraphQL.
- Enum values: The enum values (CREATION, RECEIVE, and SEND) use uppercase with underscores, which is a common convention for enum values in GraphQL.
- Overall, the schema appears to adhere to the naming patterns commonly used in GraphQL. This promotes consistency and makes it easier for developers to understand and work with your schema.
- It's worth noting that naming conventions can vary between different projects and teams, so it's important to follow the conventions established within your own project or organization.

```gpl
# -----------------------------------------------
# !!! THIS FILE WAS GENERATED BY TYPE-GRAPHQL !!!
# !!!   DO NOT MODIFY THIS FILE BY YOURSELF   !!!
# -----------------------------------------------

"""The `Decimal` scalar type to represent currency values"""
scalar Decimal

type Mutation {
  transmitTransaction(data: TransactionInput!): TransmitTransactionResult!
}

type Query {
  version: String!
}

input TransactionInput {
  amount: Decimal!
  createdAt: Int!
  type: TransactionType!
}

"""Type of the transaction"""
enum TransactionType {
  CREATION
  RECEIVE
  SEND
}

type TransmitTransactionResult {
  dltTransactionIdHex: String!
}
```

einhornimmond:
> - a graphqlresolver has to follow the graphql-spec by several formal aspects 
> - path location

I orient myself with the paths of my TypeScript files for the resolver on the backend.

> - registration in schema

I use the schema.ts from backend as basis and updated to work with Apollo 4 which remove the attribute TypeNonEmptyArray<string> from resolvers

> - a resolver as inbound api MUST be safe per security technics per roles and rights

roles and rights didn't make sense in this case and hardly to setup without db access
The dlt-connector format and send the transaction to iota which is open for everyone, the authorization will be put into the transaction themself with signature

> - the data per graphql have to follow strict rules to be validate by the graphql-schema-validation

I added some validation using class-validator (notice: backend don't use this standard validation at all)

> - and further such topic

which topics exactly?

PR #3152, opened 2023-07-21
feat(backend): after transaction creations trigger to send them to dlt-connector

clauspeterhuebner:
### Issues

- None

### Todo

- [X] None


backend/src/graphql/resolver/TransactionLinkResolver.ts
```diff
@@ -306,6 +313,10 @@ export class TransactionLinkResolver {
       } finally {
         releaseLock()
       }
+      // trigger to send transaction via dlt-connector
+      sendTransactionsToDltConnector().catch(() => {
```
ulfgebhardt, line 317: This function `sendTransactionsToDltConnector` should also write the database


```
          const dltTx = DltTransaction.create()
          dltTx.transactionId = transaction.id
          await DltTransaction.save(dltTx)
```

You could do something like `sendTransactionsToDltConnector(dbTransaction)` this way its encapsulated.

Also consider to hook to the events.
Have the client be triggered by events (e.g. sendCoin Event) instead

database/entity/0070-add_dlt_transactions_table/Transaction.ts
```diff
@@ -0,0 +1,145 @@
+/* eslint-disable no-use-before-define */
+import { Decimal } from 'decimal.js-light'
+import { BaseEntity, Entity, PrimaryGeneratedColumn, Column, OneToOne, JoinColumn } from 'typeorm'
+import { DecimalTransformer } from '../../src/typeorm/DecimalTransformer'
+import { Contribution } from '../Contribution'
+import { DltTransaction } from './DltTransaction'
```
ulfgebhardt, line 6: This file should not be modified according to the migration.

backend/src/graphql/resolver/util/sendTransactionsToDltConnector.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export async function sendTransactionsToDltConnector(): Promise<void> {
+  const dltConnector = DltConnectorClient.getInstance()
+  if (dltConnector) {
+    const dltTransactions = await DltTransaction.find({
+      where: { messageId: IsNull() },
+      relations: ['transaction'],
+      order: { createdAt: 'ASC', id: 'ASC' },
+    })
+    for (const dltTx of dltTransactions) {
```
einhornimmond, line 15: Bad Idea. When 2 or more transactions called shortly after another while the first call of sendTransactionsToDltConnector isn't finished yet, some transactions could be transmitted multiply time to the dlt-connector. The call of transmitTransaction need ca. 1 seconds!
Better give the transaction ids as parameter maybe as array for send/receive set.
NodeJs is Single Threaded but async. After a transmitTransaction is fired and still not returned (io-wait) NodeJs goes one with another stuff, like getting the next transaction.

backend/src/graphql/resolver/TransactionLinkResolver.ts
clauspeterhuebner: That was my first idea too, to write the dlt_transaction-entry in this separated and async function, but i prefer here the fact to simply insert a new dlt-entry with the transactions.id in the frame of the transactions transaction. Because if the transactions transaction will be not commited, then the dlt-entry should also not exist. The async sendTransactionsToDltConnector searches only for new and still not successfully transferred dlt-entries. 

The idea to lean on hooking like events seems good, i will do some researches there.

database/entity/0070-add_dlt_transactions_table/Transaction.ts
clauspeterhuebner: I do not understand this argument.
the only change here is to add an optional OneToOne relation to the dlt_transactions table. So i gues the current changes will have no effect.
Which migration do you mean?

backend/src/graphql/resolver/util/sendTransactionsToDltConnector.ts
clauspeterhuebner: That's a good argument and have to be considered.
Please correct my statement, if i'm wrong: 
the send-request of a transaction to dlt-connector will insert the data in the iota-tangle and the insert will be confirmed with a messageId immediatly. 
This returned messageId will be send back in the response of the sendTransaction-request. 
The period of waiting for a iota-milestone is not part of this request-response-loop.

einhornimmond: It is not waiting on iota milestone, it is waiting on iota message id. 
For generating iota message id, a bit of mining (proof-of-work) is required.
So the call of the Iota message send function from @iota/client inside the binary lib (compiled from rust) launch x threads (x = cpu core count) to try out different nonce until the result reached the difficulty target set from iota network, put the transaction onto the tangle and return with the message id.

And even if it would return immediately, it would not mean that no time would pass. Since at least two network connections are involved, it takes at least a few ms. It may not be noticed for a long time, but the longer it is in operation, the higher the probability that it will happen. And to find this error then could be very difficult, because it is very hard to trigger.

backend/src/graphql/resolver/TransactionLinkResolver.ts
einhornimmond: When you call the function after queryRunner.commitTransaction() like I original did, you know that the transaction was written into db. Because if an error happen, code after queryRunner.commitTransaction() won't be called. Instead it will jump to the catch block.

backend/src/graphql/resolver/util/sendTransactionsToDltConnector.ts
clauspeterhuebner: You're right to say, that sending a transaction multiple times must be avoided.
On my opinion to avoid this a synchronysing monitor, which lock multi processing of the same code, will solve this possibility very easy.
That's the reason why i split the `dlt_transactions-entry creation` from ´dlt_transactions-entry sending`. The dlt_transaction creation must on my opinion be part of the regular transaction-creations, which have to be finished inside the regular business processing and in time with the user-response . All other actions like sending to dlt - and in the next step the calculations of community-balance - and processing the dlt-response could be splitted and be processed in the background and outside the users-business-process.

einhornimmond: The monitor prevents concurrent execution, that is correct.
If a second transaction is to be sent within the execution of the first transaction, the execution of sendTransactionsToDltConnector is skipped. That means the second transaction will not be sent to the dlt connector for the time being. Only when sendTransactionsToDltConnector is called again during the next transaction. 
Then two transactions are sent shortly after each other. Because you sort the data request by date, the order is still correct, which is good. 

I can't see any critical error anymore so it's ok for me. Problems will only occur if I consider the rule that the Iota milestone timestamp must not be more than 2 minutes away from the createAt of the transaction. This is how it is handled by Hedera.

einhornimmond (dismissed):

backend/src/graphql/resolver/TransactionLinkResolver.ts
clauspeterhuebner: shifted the dlt-transaction creation outside the transactions-writing transaction in the async send-method

backend/src/graphql/resolver/util/sendTransactionsToDltConnector.ts
clauspeterhuebner: Good hint about the max time from createdAt of transaction, but as you say it is handled in this way by Hedera. What about iota? Do we have to consider too or is on iota no limit?
I ask, because my vision is to send a balance_date with the transaction-data and this will be very older than the current time, because of the need to migrate the still existing transactions.

einhornimmond: Iota does not know our payload, so it does not matter for Iota. It is about our own validation logic. But we can also adapt this for the import of previous transactions. In my Gradido node I have done this so that the deviation is allowed until the first live transaction. So you can only import old transactions at the beginning, once the first live transaction has come in, no more.

einhornimmond (approved):

backend/src/graphql/resolver/util/sendTransactionsToDltConnector.ts
```diff
@@ -0,0 +1,82 @@
...
+            if (dltMessageId.length !== 32) {
+              logger.error(
+                'Error dlt message id is invalid: %s, should by 32 Bytes long in binary after converting from hex',
+                dltMessageId,
+              )
+              return
+            }
+            dltTx.messageId = dltMessageId.toString('hex')
+            dltTx.verified = true
+            dltTx.verifiedAt = new Date()
```
einhornimmond, line 39: Verify comes later. This sends the transaction to the iota tangle for the time being.
transmitTransaction

This is roughly how it will look with the verify. But I didn't want to pack everything into a PR, so this step comes later.
```suggestion
```
verifyTransaction

backend/.env.dist
```diff
@@ -21,6 +21,10 @@ KLICKTIPP_PASSWORD=secret321
 KLICKTIPP_APIKEY_DE=SomeFakeKeyDE
 KLICKTIPP_APIKEY_EN=SomeFakeKeyEN
 
+# DltConnector
+DLT_CONNECTOR=true
+DLT_CONNECTOR_URL=http://localhost:6000
```
einhornimmond, line 26:
```suggestion
DLT_CONNECTOR_URL=http://localhost:6010
```
Port 6000 is used under linux from x-windows

einhornimmond (dismissed):

einhornimmond (approved):

PR #3153, opened 2023-07-23
fix(admin): user role in admin interface

Elweyn:
### Issues

- None

### Todo

- [X] Fix Auth Guard to check on roles
- [X] Fix user search so that the current searches work
- [x] Fix not working components

### DIFF
--- a/frontend/src/components/Menu/Sidebar.vue
+++ b/frontend/src/components/Menu/Sidebar.vue
@@ -49,7 +49,7 @@
           </b-nav-item>
           <b-nav-item
             class="mb-3 text-light"
-            v-if="$store.state.roles.length > 0"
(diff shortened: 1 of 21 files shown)


frontend/src/components/Menu/Sidebar.vue
```diff
@@ -49,7 +49,7 @@
           </b-nav-item>
           <b-nav-item
             class="mb-3 text-light"
-            v-if="$store.state.roles.length > 0"
+            v-if="$store.state.roles && $store.state.roles.length > 0"
```
Elweyn, line 52: @Mogge roles are not commited to the store so this should always be false or do I miss something?
```suggestion
            v-if="$store.state.moderator?.roles?.length > 0"
```

mahula (approved):

PR #3168, opened 2023-08-07
fix(backend): update logfiles clearance-script

clauspeterhuebner:
### Issues

- None

### Todo

- [X] correct name pattern for file search
- [X] correct time pattern for file search


einhornimmond (changes requested):
## find
- -ctime: File's  status  was last changed less than n*24 hours ago.  
- -atime: File was last accessed less than n*24 hours  ago.
- -mtime: File's data was last modified less than n*24  hours  ago.

I think best would be using mtime, what do you think?

clauspeterhuebner:
it depends on the last kind of file access, which time is modified.
the `ctime`, which carries the last inode change of the file will not be touched as soon as the file was compressed, but the `atime` does
the `mtime` will always be touch as soon as the content of the file was changed
...so on my opinion `mtime` will not be the correct time

einhornimmond (approved):

PR #3174, opened 2023-08-11
feat(admin): contribution filtering by memo

einhornimmond:
## 🍰 Pullrequest

User search extended in the display of contributions in the admin interface so that memo texts can also be searched. This allows to search for tags, for example.
- add button for filtering out all contributions with hashtag
- a little bit of refactor in ContributionResolver to reduce lines and make code easier to read and to maintain


backend/src/graphql/arg/Paginated.ts
```diff
@@ -5,12 +5,12 @@ import { Order } from '@enum/Order'
...
 
-  @Field(() => Int, { nullable: true })
-  pageSize?: number
+  @Field(() => Int, { defaultValue: 3 })
+  pageSize: number
 
-  @Field(() => Order, { nullable: true })
-  order?: Order
+  @Field(() => Order, { defaultValue: Order.DESC })
+  order: Order
```
clauspeterhuebner, line 15: what does the new search feature "# in memo" has to do with pagination?
Is there a reason why to change the default settings of the class attributes?

clauspeterhuebner (approved):
very good work...

backend/src/graphql/arg/Paginated.ts
einhornimmond: It's not directly related to pagination, although it does share that. However, I have refactored the Pagination object a bit along with the findContributions function to be a bit more object oriented. 
Instead of setting the default values in the three resolver functions that use Pagination, they are now only set in one place in the Paginated object.
That means the nullable fields were always filled with default values anyway.

Translated with www.DeepL.com/Translator (free version)

PR #3182, opened 2023-08-19
feat(admin): update hashtag ui

einhornimmond:
### Issues

- None


clauspeterhuebner (approved):
...but there is no change about the sql-statement yet, right?
Does it make sense to add this change here too or is this still solved in an other PR?

i saw the SQL-statement changes in the next PR.
So it is OK...

PR #3207, opened 2023-09-13
feat(dlt): add community, use dlt-database

einhornimmond:
## 🍰 Pullrequest

- add Resolver and Controller for adding community to dlt-database
- connect to dlt-database
- test with dlt-database (using synchonize and dropSchema features from typeORM)


dlt-database/entity/index.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export const entities = [
+  AccountCommunity,
+  Account,
+  Community,
+  ConfirmedTransaction,
+  InvalidTransaction,
+  Migration,
+  TransactionRecipe,
+  User,
```
clauspeterhuebner, line 18: only a hint:
we use in the database the naming conventions for tables in plural, means:
table name = Users, Communities,...
entity name = User, Community,...

dlt-database/migrations/0002-refactor_add_community.ts
clauspeterhuebner, line 1: i thought you create you database per typeORM? Why now here using a database script for adaptions?

dlt-connector/src/controller/Community.ts
```diff
@@ -0,0 +1,36 @@
...
+  const result = await getDataSource().manager.find(Community, {
+    where: { iotaTopic },
+  })
+  return result.length > 0
+}
+
+export const create = (community: CommunityDraft, topic?: string): Community => {
+  const communityEntity = new Community()
+  communityEntity.iotaTopic = topic ?? iotaTopicFromCommunityUUID(community.uuid)
+  communityEntity.createdAt = new Date(community.createdAt)
```
clauspeterhuebner, line 25: the column `createdAt` annotated with `@CreateDateColumn` will set its value automatically. So please do not set it manually in the code. I don't know if the given value or the automatical value will be stored in database.

dlt-connector/src/controller/Community.test.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  afterAll(async () => {
+    await TestDB.instance.teardownTestDB()
+  })
+
+  describe('createCommunity', () => {
+    it('valid community', async () => {
+      const communityDraft = new CommunityDraft()
+      communityDraft.foreign = false
+      communityDraft.createdAt = '2022-05-01T17:00:12.128Z'
```
clauspeterhuebner, line 30: the column createdAt annotated with @CreateDateColumn will set its value automatically. So please do not set it manually in the code. I don't know if the given value or the automatical value will be stored in database.

dlt-connector/src/controller/Community.test.ts
```diff
@@ -0,0 +1,62 @@
...
+    it('get all topics', async () => {
+      expect(await getAllTopics()).toMatchObject([
+        '204ef6aed15fbf0f9da5819e88f8eea8e3adbe1e2c2d43280780a4b8c2d32b56',
+      ])
+    })
+
+    it('isExist with communityDraft', async () => {
+      const communityDraft = new CommunityDraft()
+      communityDraft.foreign = false
+      communityDraft.createdAt = '2022-05-01T17:00:12.128Z'
```
clauspeterhuebner, line 57: the column createdAt annotated with @CreateDateColumn will set its value automatically. So please do not set it manually in the code. I don't know if the given value or the automatical value will be stored in database.

clauspeterhuebner (changes requested):

dlt-connector/package.json
```diff
@@ -25,22 +25,26 @@
     "cors": "^2.8.5",
     "cross-env": "^7.0.3",
     "decimal.js-light": "^2.5.1",
+    "dlt-database": "file:../dlt-database",
```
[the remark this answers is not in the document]
einhornimmond: I don't know, I have it copied over from backend.
But I can say you what happen if it isn't there. On run yarn test, I get this error: 
```bash
DriverPackageNotInstalledError: Mysql package has not been found installed. Try to install it: npm install mysql --save
        at MysqlDriver.loadDependencies (~/code/gradido_apollo/dlt-connector/src/driver/mysql/MysqlDriver.ts:1197:23)
        at new MysqlDriver (~/code/gradido_apollo/dlt-connector/src/driver/mysql/MysqlDriver.ts:354:14)
        at DriverFactory.create (~/code/gradido_apollo/dlt-connector/src/driver/DriverFactory.ts:34:24)
        at createDriver (~/code/gradido_apollo/dlt-connector/node_modules/typeorm-extension/src/database/driver/utils/create.ts:20:26)
        at createMySQLDatabase (~/code/gradido_apollo/dlt-connector/node_modules/typeorm-extension/src/database/driver/mysql.ts:52:20)

```

dlt-connector/Dockerfile
```diff
@@ -90,9 +99,10 @@ RUN apk del rust cargo python3 make g++
 
 # Copy "binary"-files from build image
 COPY --from=build ${DOCKER_WORKDIR}/build ./build
-
+COPY --from=build ${DOCKER_WORKDIR}/../dlt-database/build ../dlt-database/build
 # We also copy the node_modules express and serve-static for the run script
 COPY --from=build ${DOCKER_WORKDIR}/node_modules ./node_modules
+COPY --from=build ${DOCKER_WORKDIR}/../dlt-database/node_modules ../dlt-database/node_modules
```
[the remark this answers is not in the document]
einhornimmond: No, it's a multistage Docker build.
Each stage has its own internal memory.
Everything needed to run in production is copied here from the build stage. It's like copying the data from one disk to another and then unmounting the src disk. It is not possible to use links at this point.

dlt-connector/src/graphql/resolver/CommunityResolver.ts
```diff
@@ -0,0 +1,78 @@
...
+      }
+
+      const queryRunner = getDataSource().createQueryRunner()
+      await queryRunner.connect()
+      await queryRunner.startTransaction()
+
+      try {
+        await queryRunner.manager.save(community)
+        await queryRunner.commitTransaction()
+        result = new TransactionResult()
```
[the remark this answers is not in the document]
einhornimmond: Because in case of home community in future there will be more than one transaction which should be saved together. 
But you right, in this state I don't need it and it is maybe better to use query runner and transaction only if I really need them, even if the code is than longer.

dlt-connector/src/graphql/validator/DateString.ts
```diff
@@ -0,0 +1,21 @@
...
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  return function (object: Object, propertyName: string) {
+    registerDecorator({
+      name: 'isValidDateString',
+      target: object.constructor,
+      propertyName,
+      options: validationOptions,
+      validator: {
+        validate(value: string): boolean {
+          return new Date(value).toString() !== 'Invalid Date'
```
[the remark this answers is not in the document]
einhornimmond: Oh that's good to know, I changed it in a hopefully platform independent way.

dlt-connector/src/utils.test.ts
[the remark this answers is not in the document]
einhornimmond: ok, then I will split it in typeConverter and for the key derivation helper which don't belong in this pr at all I look for a better place or a own file in this folder.

dlt-database/entity/index.ts
einhornimmond: Yeah I know, because the reason is first I used cakephp for the community an there the standard is using this type of name conventions. As we switched over to NodeJs where the name convention is using singular we keep using the plural names for tables but not for entities. Therefore we need to use the @Column({ name: 'plural_table_name' }) to tell typeorm how the table is named, else it would simply use the Entity name directly.

dlt-database/migrations/0002-refactor_add_community.ts
einhornimmond: I use typeORM only for the test database.
But maybe we can use typeORM and his migration capabilities like everyone does and don't move the entities in another folder. But first I need to know why this decision was made from the team at the first place. Do you know it? Can we discuss it at the next meeting?

dlt-connector/src/controller/Community.ts
einhornimmond: Oh ok, I thought I must use it together with default: TIMESTAMP.
I have it replaced with Column.

dlt-connector/src/controller/Community.test.ts
einhornimmond: Oh ok, I thought I must use it together with default: TIMESTAMP.
I have it replaced with Column.
default works for me as fallback if I don't set the value manually, but mostly I think I will.

dlt-connector/src/controller/Community.test.ts
einhornimmond: Oh ok, I thought I must use it together with default: TIMESTAMP.
I have it replaced with Column.
default works for me as fallback if I don't set the value manually, but mostly I think I will.

dlt-connector/src/utils.ts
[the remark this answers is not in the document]
einhornimmond: ok, then I will split it in typeConverter and for the key derivation helper which don't belong in this pr at all I look for a better place or a own file in this folder.

dlt-connector/Dockerfile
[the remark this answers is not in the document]
clauspeterhuebner: Ah ok.

dlt-connector/src/controller/Community.ts
```diff
@@ -0,0 +1,36 @@
...
+  const iotaTopic =
+    community instanceof CommunityDraft ? iotaTopicFromCommunityUUID(community.uuid) : community
+  const result = await getDataSource().manager.find(Community, {
+    where: { iotaTopic },
+  })
+  return result.length > 0
+}
+
+export const create = (community: CommunityDraft, topic?: string): Community => {
+  const communityEntity = new Community()
```
clauspeterhuebner, line 23: Isn't it necessary to use the internal EntityRepository with internal TX-Support to instanciate an entity by calling here Community.create() than the new-Constructor?
Then you can use for writing later the expression Community.save(comEntity), which handles the transaction and differ between insert and update

dlt-connector/src/graphql/resolver/CommunityResolver.ts
```diff
@@ -0,0 +1,69 @@
...
+      }
+      const community = createCommunity(communityDraft, topic)
+
+      let result: TransactionResult
+
+      if (!communityDraft.foreign) {
+        // TODO: CommunityRoot Transaction for blockchain
+      }
+      try {
+        await getDataSource().manager.save(community)
```
clauspeterhuebner, line 52: please use the entity-repository with the including TX-Handling.
See also the comment in Community-Controler line 23.

dlt-database/entity/0001-init_db/Community.ts
```diff
@@ -46,19 +37,15 @@ export class Community {
   @JoinColumn({ name: 'auf_account_id' })
   aufAccount?: Account
 
-  @CreateDateColumn({ name: 'created_at', type: 'datetime', default: () => 'CURRENT_TIMESTAMP(3)' })
+  @Column({ name: 'created_at', type: 'datetime', default: () => 'CURRENT_TIMESTAMP(3)' })
```
clauspeterhuebner, line 40: If the `createdAt` column will only contain the timestamp of the write access than the annotation `@CreateDateColumn` will be prefered.
But keep in mind:
the `@CreateDateColumn` and/or `@UpdateDateColumn` will not treat the ms-dimension properly

dlt-database/entity/0001-init_db/TransactionRecipe.ts
```diff
@@ -63,7 +55,7 @@ export class TransactionRecipe {
   @Column({ type: 'tinyint' })
   type: number
 
-  @CreateDateColumn({ name: 'created_at', type: 'datetime' })
+  @Column({ name: 'created_at', type: 'datetime' })
```
clauspeterhuebner, line 58: If the `createdAt` column will only contain the timestamp of the write access than the annotation `@CreateDateColumn` will be prefered.
But keep in mind:
the `@CreateDateColumn` and/or `@UpdateDateColumn` will not treat the ms-dimension properly

dlt-database/entity/0001-init_db/User.ts
```diff
@@ -26,7 +18,7 @@ export class User extends BaseEntity {
   @Column({ name: 'derive1_pubkey', type: 'binary', length: 32, unique: true })
   derive1Pubkey: Buffer
 
-  @CreateDateColumn({
+  @Column({
```
clauspeterhuebner, line 21: If the `createdAt` column will only contain the timestamp of the write access than the annotation `@CreateDateColumn` will be prefered.
But keep in mind:
the `@CreateDateColumn` and/or `@UpdateDateColumn` will not treat the ms-dimension properly

clauspeterhuebner (dismissed):
simply some remarks or hints.
feel free if you want to follow them...

dlt-connector/yarn.lock
```diff
@@ -1981,7 +2095,7 @@ cross-env@^7.0.3:
   dependencies:
     cross-spawn "^7.0.1"
 
-cross-spawn@^7.0.1, cross-spawn@^7.0.2, cross-spawn@^7.0.3:
+cross-spawn@^7.0.0, cross-spawn@^7.0.1, cross-spawn@^7.0.2, cross-spawn@^7.0.3:
```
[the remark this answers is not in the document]
einhornimmond: If you look closely at the line, you will see that cross-spawn@^7.0.0 is added and the rest remains as it is. I can't see any downgrade in it. So obviously one of the newly added packages or one of the dependencies needs the cross-spawn@^7.0.0 version and that is why it is added. This happens automatically so yarn takes care of it.

dlt-connector/yarn.lock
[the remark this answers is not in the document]
einhornimmond: I have added "sodium-native", "typeorm-extension",
"typeorm" and "dlt-database" added. See the package.json for this.
From sodium-native I need a hash function and the other three are for database binding.
What happens in the yarn.lock happens automatically, there all dependencies for the installed packages are installed. And since it is NodeJS and there is a separate package for every little thing, this can get a bit out of hand. One of the reasons why I don't like NodeJS.

dlt-database/entity/0002-refactor_add_community/Community.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  @OneToOne(() => Account)
+  @JoinColumn({ name: 'auf_account_id' })
+  aufAccount?: Account
+
+  @CreateDateColumn({ name: 'created_at', type: 'datetime', default: () => 'CURRENT_TIMESTAMP(3)' })
+  createdAt: Date
+
+  @Column({ name: 'confirmed_at', type: 'datetime', nullable: true })
+  confirmedAt?: Date
```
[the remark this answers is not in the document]
einhornimmond: I have revised the date accuracy now again and everywhere where it makes no sense it again taken back. 
- BalanceDate: Why do I need milliseconds here? The Decay function works only on seconds basis.
- ConfirmedAt: Why do I need milliseconds here? The Confirmation date is the timestamp of the Iota milestone and this is only in seconds.
I have added precision option in entities and a test in Community to test that ms precision work.

dlt-connector/src/graphql/resolver/CommunityResolver.ts
```diff
@@ -0,0 +1,69 @@
...
+  // Why a dummy function?
+  // to prevent this error by start:
+  //   GeneratingSchemaError: Some errors occurred while generating GraphQL schema:
+  //     Type Query must define one or more fields.
+  // it seems that at least one query must be defined
+  // https://github.com/ardatan/graphql-tools/issues/764
+  @Query(() => String)
+  version(): string {
+    return '0.1'
+  }
```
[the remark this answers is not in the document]
einhornimmond: The dummy query is necessary if there is no other query. No matter how many mutations there are. However, a query in one of the resolvers is sufficient, there does not need to be a query in every resolver. I just tried this, I can remove the dummy query from the community resolver because there is still one in the transaction resolver.

dlt-connector/src/graphql/resolver/CommunityResolver.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  @Mutation(() => TransactionResult)
+  async addCommunity(
+    @Arg('data')
```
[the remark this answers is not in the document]
einhornimmond: Less to write.
With @Arg('data'):
```grapqhl
mutation ($input: CommunityDraft!) { addCommunity(data: $input) {succeed} }
```
With @Args():
```graphql
query ($uuid: String) {
    isCommunityExist(uuid: $uuid)
  }
mutation ($uuid: String!, $foreign: Boolean!, $createdAt: String! ) { 
 addCommunity(uuid: $uuid, foreign: $foreign, createdAt: $createdAt) {
  succeed
} }
```

dlt-connector/src/controller/Community.ts
einhornimmond: If you want to use the Internal EntityRepository yes, also the entity classes must then be derived from BaseEntity. I have changed this now, originally I saved this optimization to be a bit faster.

dlt-connector/src/graphql/resolver/CommunityResolver.ts
einhornimmond: Okay, that's fine, I'll do :P

dlt-connector/src/graphql/resolver/CommunityResolver.ts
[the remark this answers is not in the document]
clauspeterhuebner: Good to know, many thanx...

dlt-connector/yarn.lock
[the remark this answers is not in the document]
clauspeterhuebner: everything's good, i still wondered about the amount of additional packages.
Thx for your detailed explanation...

dlt-database/entity/0002-refactor_add_community/Account.ts
```diff
@@ -0,0 +1,78 @@
...
+    transformer: DecimalTransformer,
+  })
+  balance: Decimal
+
+  @Column({
+    name: 'balance_date',
+    type: 'datetime',
+    default: () => 'CURRENT_TIMESTAMP()',
+  })
+  balanceDate: Date
```
clauspeterhuebner, line 64: in the `transactions`-table the `balance_date` is defined with `precision = 3`

dlt-database/migrations/0002-refactor_add_community.ts
```diff
@@ -0,0 +1,61 @@
...
+    `ALTER TABLE \`accounts\` MODIFY COLUMN \`balance_date\` datetime NOT NULL DEFAULT CURRENT_TIMESTAMP;`,
+  )
+  await queryFn(
+    `ALTER TABLE \`accounts_communities\` MODIFY COLUMN \`valid_from\` datetime NOT NULL;`,
+  )
+  await queryFn(
+    `ALTER TABLE \`accounts_communities\` MODIFY COLUMN \`valid_to\` datetime NULL DEFAULT NULL;`,
+  )
+  await queryFn(
+    `ALTER TABLE \`confirmed_transactions\` MODIFY COLUMN \`confirmed_at\` datetime NOT NULL;`,
```
clauspeterhuebner, line 32: as i know the precision for ms is important to be defined at first in the database.
If the entity will deliver ms, but the database column is not prepared, the ms will be lost or reset to `.000`

so i'am wondering about your solution, but if it works... good to know!

dlt-database/entity/0001-init_db/Account.ts
```diff
@@ -56,17 +59,14 @@ export class Account {
   @Column({
     name: 'balance_date',
     type: 'datetime',
-    default: () => 'CURRENT_TIMESTAMP(3)',
+    precision: 3,
+    default: () => 'CURRENT_TIMESTAMP()',
```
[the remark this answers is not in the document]
einhornimmond: Oh you right, I have fixed it.

dlt-database/entity/0002-refactor_add_community/Account.ts
einhornimmond: Take a closer look. This is for migration 0002 where I had set all balance_date and confirmation_date back to not use milliseconds.

clauspeterhuebner (approved):

PR #3224, opened 2023-10-30
feat(frontend): swap support urls with mailto: links

einhornimmond:
- change support link in wallet dashboard footer from hardcoded gradido url to support email from config
- change support email in email template to mailto: link


frontend/src/components/ContentFooter.vue
```diff
@@ -44,7 +44,7 @@
           >
             {{ $t('footer.whitepaper') }}
           </b-nav-item>
-          <b-nav-item :href="`https://gradido.net/${$i18n.locale}/contact/`" target="_blank">
+          <b-nav-item :href="`mailto:${supportEmail}`" target="_blank">
```
clauspeterhuebner, line 47: do we lost some i18n features with the changed code or is it not necessary or just not in use?

clauspeterhuebner (approved):
with one simple question...

frontend/src/components/ContentFooter.vue
einhornimmond: The task was to replace the support link that leads to a different subpage of the Gradido Academy main page for each language with a mailto: link with which you can send an e-mail to the support of the specific community server. This address is entered in the config. 
The link to which you are currently redirected only contains the imprint and an e-mail address to which you can send your questions. It is translated into different languages, but the support e-mail address always remains the same. 
So with my change we lose the different links per language, i.e. an i18n feature. 
As long as we don't have a config per language per community, I don't see how we could use the i18n feature for the direct mailto: support link. 


Translated with www.DeepL.com/Translator (free version)

PR #3230, opened 2023-11-01
fix(frontend): fix to less moderator/admins on information page

einhornimmond:
It seems default value in graphql/arg/Paginated overwrite default values which are used in various places in resolvers.
In searchAdminUsers default value is set to 25 was overwritten by 3 in Paginated. The frontend call searchAdminUsers without parameters so it gets only the default count of results.

TODO: Rethink pagination mechanism in light of larger communities. Not all places can handle more than one page results.

### DIFF
--- a/backend/src/graphql/arg/Paginated.ts
+++ b/backend/src/graphql/arg/Paginated.ts
@@ -6,15 +6,15 @@ import { Order } from '@enum/Order'
 
 @ArgsType()
 export class Paginated {
-  @Field(() => Int, { defaultValue: 1 })
+  @Field(() => Int, { nullable: true })
   @IsPositive()
-  currentPage: number
+  currentPage?: number
 
   @Field(() => Int, { nullable: true })
   @IsPositive()
   pageSize?: number
 
-  @Field(() => Order, { defaultValue: Order.DESC })
+  @Field(() => Order, { nullable: true })
   @IsEnum(Order)
-  order: Order
+  order?: Order
 }
--- a/backend/src/graphql/arg/Paginated.ts
+++ b/backend/src/graphql/arg/Paginated.ts
@@ -10,9 +10,9 @@ export class Paginated {
   @IsPositive()
   currentPage: number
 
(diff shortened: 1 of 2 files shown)


clauspeterhuebner (changes requested):
To remove the default initialization, but not to ensure for all usages an initialization could cause more problems than expected.
There are several usages of Paginated without value initialization:
- ContributionResolver.listContributions
- ContributionResolver.listAllContributions
- ContributionResolver.adminListContributions
- TransactionLinkResolver.listTransactionLinks
- TransactionLinkResolver.listTransactionLinksAdmin
and how is it ensured to have at least a default initialization in case the invoker will not set them?

einhornimmond:
- ContributionResolver.listContributions call findContributions with use default initializer (pageSize 3)
- ContributionResolver.listAllContributions call findContributions with use default initializer (pageSize 3)
- ContributionResolver.adminListContributions call findContributions with use default initializer (pageSize 3)
- TransactionLinkResolver.listTransactionLinks call transactionLinkList with use default initializer (pageSize 5)
- TransactionLinkResolver.listTransactionLinksAdmin call transactionLinkList with use default initializer (pageSize 5)

- ContributionLinkResolver.listContributionLinks use default initializer (pageSize 5)
- ContributionMessageResolver.listContributionMessages use default initializer (pageSize 5)
- ContributionMessageResolver.adminListContributionMessages use default initializer (pageSize 5)
- GdtResolver.listGDTEntries use default initializer (pageSize 5)
- TransactionResolver.transactionList use default initializer (pageSize 25)
- UserResolver.searchAdminUsers use default initializer (pageSize 25)
- UserResolver.searchUsers use default initializer (pageSize 25)

All this default initializer where overwritten with 3 from Paginated.
TypeScript make sure that the pageSize? value must be get a default initializer or must be checked if it was set.
But with default value set in Paginated all this different default initializer don't work any longer.

clauspeterhuebner (approved):

PR #3242, opened 2023-11-13
feat(workflow): fix backend and database build error on windows

einhornimmond:
Fix backend and database build error on windows.
It is triggered by the use of mkdir -p and cp which do not exist under Windows. The solution is to use the npm packages mkdirp and ncp which were created exactly for this purpose.

### DIFF
--- a/backend/package.json
+++ b/backend/package.json
@@ -8,7 +8,7 @@
   "license": "Apache-2.0",
   "private": false,
   "scripts": {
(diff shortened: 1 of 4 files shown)


clauspeterhuebner (commented):
as mentioned in the decription of this PR the solution base on additional npm-packages.
But actually the gradido project uses yarn instead of npm.
Isn't it necessary to switch first from yarn to npm over all modules before merging these changes of only two modules into the master?

einhornimmond:
yarn ist ein Paketmanager ähnlich wie npm. Aber die Pakete die wir durch yarn installieren, sind dieselben als wenn wir sie mit npm installieren würden. Das verwendete Repository ist dasselbe.

clauspeterhuebner (approved):

PR #3252, opened 2023-11-23
feat(admin): resubmission

einhornimmond:
## 🍰 Pullrequest

Select a date on which the contribution should be displayed again. Until then it will be hidden in default mode. But you can display it at any time by unchecking the checkbox for resubmission. 

- update findContributions query builder 
- add checkbox to admin underneath checkbox for hashtag
- add signal which will be emitted after adding comment to contribution which set resubmission date for updating contribution list
- add resubmission_at date to contribution_messages table and entity and ContributionMessageArgs
- add props to ContributionMessagesFormular for accessing checkbox for hiding resubmission to only update contribution list after adding resubmission date if resubmission should be hided 
- update frontend to show "moderator has edited contribution" in histories messages if messages was created from moderator

### DIFF
--- a/frontend/src/components/ContributionMessages/ContributionMessagesListItem.vue
+++ b/frontend/src/components/ContributionMessages/ContributionMessagesListItem.vue
@@ -4,8 +4,11 @@
       <b-row class="mb-3 border border-197 p-1">
         <b-col cols="10">
           <small>{{ $d(new Date(message.createdAt), 'short') }}</small>
-          <div class="font-weight-bold" data-test="username">
+          <div class="font-weight-bold" data-test="username" v-if="isNotModerator">
             {{ storeName.username }} {{ $t('contribution.isEdited') }}
+          </div>
+          <div class="font-weight-bold" data-test="moderator-name" v-else>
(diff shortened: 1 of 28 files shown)


frontend/src/components/ContributionMessages/ContributionMessagesListItem.vue
```diff
@@ -4,9 +4,12 @@
       <b-row class="mb-3 border border-197 p-1">
         <b-col cols="10">
           <small>{{ $d(new Date(message.createdAt), 'short') }}</small>
-          <div class="font-weight-bold" data-test="username">
+          <div class="font-weight-bold" data-test="username" v-if="isNotModerator">
```
clauspeterhuebner, line 7: does this include an Admin too?

clauspeterhuebner (approved):
good work! only one question...

frontend/src/components/ContributionMessages/ContributionMessagesListItem.vue
einhornimmond: Yes, the code checks that the logged in user first name and last name is identical with message user first name and user last name. I simply used something already existing. I personally would prefer to check the user id rather than first and last name.

PR #3256, opened 2023-11-30
feat(backend): 3237 feature connect gms api with a new rest client

clauspeterhuebner:
### Issues

- None

### Todo

- [X] None


backend/src/apis/gms/model/GmsEnums.ts
```diff
@@ -0,0 +1,36 @@
...
+export enum GmsLocationType {
+  GMS_LOCATION_TYPE_EXACT = 0,
+  GMS_LOCATION_TYPE_APPROXIMATE = 1,
+  GMS_LOCATION_TYPE_RANDOM = 2,
+}
+
+registerEnumType(GmsLocationType, {
+  name: 'GmsLocationType', // this one is mandatory
+  description: 'Type of location treatment in GMS', // this one is optional
+})
```
einhornimmond, line 36: Maybe you can put the enums in separate files in a separate enum folder?
registerEnumType you need only if you like to use the enum in a graphql object/arg/input.

backend/src/apis/gms/model/GmsUser.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  private getGmsPhone(user: dbUser): string | undefined {
+    if (
+      user.gmsAllowed &&
+      user.emailContact.gmsPublishPhone === GmsPublishPhoneType.GMS_PUBLISH_PHONE_FULL
+    ) {
+      return user.emailContact.phone
+    }
+  }
+}
```
einhornimmond, line 108: I think the function names are a bit irritating.
For example:
```
GmsUser.getGmsAlias
```
what you would expect simply from the function name what you get?
Only together with the input it makes sense. On DbUser this names would make more sense, but this were the wrong place for this api specific logic.
Maybe you can extend the function names, for example:
```
GmsUser.getGmsAliasFromDbUser
```
Or use a separate Object for this Converter or Logic, I'm not sure what is better.
```ts
class DbUserConverter {
public constructor(private user: DbUser) {}
public getGmsAlias(): string | undefined {
    if (
      this.user.gmsAllowed &&
      this.user.alias &&
      this.user.gmsPublishName === GmsPublishNameType.GMS_PUBLISH_NAME_ALIAS_OR_INITALS
    ) {
      return this.user.alias
    }
  }
```

database/migrations/0079-introduce_gms_registration.ts
```diff
@@ -0,0 +1,54 @@
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+export async function upgrade(queryFn: (query: string, values?: any[]) => Promise<Array<any>>) {
+  await queryFn(
+    'ALTER TABLE `users` MODIFY COLUMN `foreign` tinyint(1) NOT NULL DEFAULT 0 AFTER `id`;',
```
einhornimmond, line 6: is 
```AFTER `id` ``` 
necessary on modify column?

**Review by einhornimmond — CHANGES_REQUESTED** [review]
Yes, okay, I can't leave some things as they are.

File: backend/src/apis/gms/model/GmsEnums.ts (hunk 1 above)
Comment (clauspeterhuebner) [reply backend/src/apis/gms/model/GmsEnums.ts:36]: done

File: backend/src/apis/gms/model/GmsUser.ts (hunk 2 above)
Comment (clauspeterhuebner) [reply backend/src/apis/gms/model/GmsUser.ts:108]: I will agree with you if the functions are defined as public.
But here all they defined as privat functions and will in consequence be only invocable from inside the GmsUser-instance.
The GmsUser itself has no further public api than its own attributes.

As you write the private functions make sense in conjunction with the DbUser and this is only accessable in the constructor of the GmsUser.
The idea behind these functions is to hide the logic of the evaluation of the control settings and which value will be returned for the dedicated attribute like `getGmsAlias`, `getGmsFirstName` etc....

So i see no other or better context to shift them for a more efficient granuality.

File: database/migrations/0079-introduce_gms_registration.ts (hunk 3 above)
Comment (clauspeterhuebner) [reply database/migrations/0079-introduce_gms_registration.ts]: as i know yes...

File: backend/src/seeds/gmsUsers.ts [hunk 4]
```diff
@@ -0,0 +1,102 @@
...
+  if (items.length > 0) {
+    const ids = items.map((e: any) => e.id)
+    await entity.delete(ids)
+  }
+}
+
+const run = async () => {
+  const server = await createServer(context)
+  // const seedClient = createTestClient(server.apollo)
+  const { con } = server
```
einhornimmond, line 52: Why are you creating a complete apollo server here if you only need the database connection? 

```suggestion
  // open mysql connection
  const con = await Connection.getInstance()
  if (!con?.isConnected) {
    logger.fatal(`Couldn't open connection to database!`)
    throw new Error(`Fatal: Couldn't open connection to database`)
  }

  // check for correct database version
  const dbVersion = await checkDBVersion(CONFIG.DB_VERSION)
  if (!dbVersion) {
    logger.fatal('Fatal: Database Version incorrect')
    throw new Error('Fatal: Database Version incorrect')
  }
```

backend/src/seeds/gmsUserList.ts
```diff
@@ -0,0 +1,60 @@
...
+  if (items.length > 0) {
+    const ids = items.map((e: any) => e.id)
+    await entity.delete(ids)
+  }
+}
+
+const run = async () => {
+  const server = await createServer(context)
+  // const seedClient = createTestClient(server.apollo)
+  const { con } = server
```
einhornimmond, line 46: Why are you creating a complete apollo server here if you only need the database connection? 

```suggestion
  // open mysql connection
  const con = await Connection.getInstance()
  if (!con?.isConnected) {
    logger.fatal(`Couldn't open connection to database!`)
    throw new Error(`Fatal: Couldn't open connection to database`)
  }

  // check for correct database version
  const dbVersion = await checkDBVersion(CONFIG.DB_VERSION)
  if (!dbVersion) {
    logger.fatal('Fatal: Database Version incorrect')
    throw new Error('Fatal: Database Version incorrect')
  }
```

deployment/bare_metal/.env.dist
```diff
@@ -118,4 +118,13 @@ NGINX_SSL_INCLUDE=/etc/letsencrypt/options-ssl-nginx.conf
 # LEGACY
 NGINX_REWRITE_LEGACY_URLS=false
 DEFAULT_PUBLISHER_ID=2896
-WEBHOOK_ELOPAGE_SECRET=secret
\ No newline at end of file
+WEBHOOK_ELOPAGE_SECRET=secret
+
+WALLET_AUTH_URL=https://stage1.gradido.net/authenticate?token={token}
+WALLET_URL=https://stage1.gradido.net/login
```
einhornimmond, line 124: Why did you add these lines again? They are now further up and end in path and no longer in url.
```suggestion
```

einhornimmond (approved):
only some suggestions for performance reasons.

PR #3260, opened 2023-12-14
doc(feature): 3258 feature create gms usecase docu

clauspeterhuebner:
### Issues

- None

### Todo

- [X] None


BerndHueckstaedt:
Questions:
Is the exact geodata saved or the randomized data? In my opinion, the data should be randomized before and only then saved.

I think I read somewhere that the user gets the other users of his community displayed in his playground - according to his search criteria.
As far as I know, there is a filter for the community from which the other users are displayed. All communities are also possible. That's a good thing.

BerndHueckstaedt (approved):

clauspeterhuebner:
@BerndHueckstaedt:
Questions:
Is the exact geodata saved or the randomized data? In my opinion, the data should be randomized before and only then saved.

my answer:
The user will capture his location by entering his postal address data in the popup-window as exact as he wants. The result of this input will be returned to the gradido-system and stored there. The location-transfer to GMS will depend on the type the user grants: 0=exact, 1=approximate, 2=random (default).
The GMS itself will first interpret this location-type and depending on this value the given location from gradido will be stored - default is the randomized location around the community-location.


I think I read somewhere that the user gets the other users of his community displayed in his playground - according to his search criteria.
As far as I know, there is a filter for the community from which the other users are displayed. All communities are also possible. That's a good thing.

my answer:
the current playground dialog shows filter fields for "communities", "listing types - means offers, needs" and "radius".
The open question is about the default filter-settings when the user opens his playground?
Should the default filter settings be set on:
- the user's community or all communities - on my opinion all
- all listing types or a restricting value - on my opinion at the moment all
- radius should be set on the community's radius value, because the randomized user locations will depend on this value

PR #3286, opened 2024-02-14
feat(workflow): add videos to deployment readme

einhornimmond:
- add Videos to deployment readme
- update install.sh, branch is now needed always
- remove server suggestions because it is working with smallest server and can be easy rescaled


BerndHueckstaedt (approved):
Super 👍

PR #3288, opened 2024-02-15
fix(backend): prevent running validate communities continuous

einhornimmond:
## 🍰 Pullrequest
FEDERATION_VALIDATE_COMMUNITY_TIMER: 
    Number(process.env.FEDERATION_VALIDATE_COMMUNITY_TIMER) `??`  60000,
  
lead to FEDERATION_VALIDATE_COMMUNITY_TIMER = NaN if process.env.FEDERATION_VALIDATE_COMMUNITY_TIMER wasn't set. That lead to continuously running validate communities, put one CPU Core to 100% all time and fill the hdd with logs.

So to prevent this, changing back FEDERATION_VALIDATE_COMMUNITY_TIMER: 
    Number(process.env.FEDERATION_VALIDATE_COMMUNITY_TIMER) `||` 60000
and add additional check for positive number before running validate communities.

### DIFF
--- a/backend/src/config/index.ts
+++ b/backend/src/config/index.ts
@@ -128,7 +128,8 @@ if (
 
 const federation = {
   FEDERATION_BACKEND_SEND_ON_API: process.env.FEDERATION_BACKEND_SEND_ON_API ?? '1_0',
-  FEDERATION_VALIDATE_COMMUNITY_TIMER: // ?? operator don't work here as expected
+  // ?? operator don't work here as expected
+  FEDERATION_VALIDATE_COMMUNITY_TIMER:
     Number(process.env.FEDERATION_VALIDATE_COMMUNITY_TIMER) || 60000,
   FEDERATION_XCOM_SENDCOINS_ENABLED:
     process.env.FEDERATION_XCOM_SENDCOINS_ENABLED === 'true' ?? false,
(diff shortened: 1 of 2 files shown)


backend/src/config/index.ts
```diff
@@ -128,8 +128,9 @@ if (
 
 const federation = {
   FEDERATION_BACKEND_SEND_ON_API: process.env.FEDERATION_BACKEND_SEND_ON_API ?? '1_0',
+  // ?? operator don't work here as expected
   FEDERATION_VALIDATE_COMMUNITY_TIMER:
-    Number(process.env.FEDERATION_VALIDATE_COMMUNITY_TIMER) ?? 60000,
+    Number(process.env.FEDERATION_VALIDATE_COMMUNITY_TIMER) || 60000,
```
clauspeterhuebner, line 133:
```suggestion
    Number(process.env.FEDERATION_VALIDATE_COMMUNITY_TIMER ?? 60000) ,
```

perhaps the operator will have not the expected effect, because the property is evaluated inside the invocation of Number( ), right?

clauspeterhuebner (approved):
with one suggestion...

backend/src/config/index.ts
einhornimmond: you are right, yes ok.

PR #3289, opened 2024-02-15
refactor(backend): refactor community and add to dlt-connector

einhornimmond:
## 🍰 Pullrequest

- refactor community load code, reduce double code
- let dlt-connector request home community on startup
- let dlt-connector wait until backend is ready


backend/src/seeds/graphql/queries.ts
```diff
@@ -135,8 +135,8 @@ export const communitiesQuery = gql`
 `
 
 export const getCommunityByUuidQuery = gql`
-  query ($communityUuid: String!) {
-    community(communityUuid: $communityUuid) {
+  query ($communityIdentifier: String!) {
+    community(communityIdentifier: $communityIdentifier) {
```
clauspeterhuebner, line 139: how can you serve with this query the signature of the `CommuntityResolver.community`-method with a 2nd parameter `foreign`?
Yes the 2nd parameter is optional, but not able to set over this query. Is this intended?

backend/src/util/validate.ts
```diff
@@ -16,6 +17,14 @@ function isStringBoolean(value: string): boolean {
...
 }
 
+function isUUID4(value: string): boolean {
+  return validate(value) && version(value) === 4
+}
+
+function isEMail(value: string): boolean {
+  return /^.{2,}@.{2,}\..{2,}$/.exec(value) !== null
+}
+
```
clauspeterhuebner, line 27: very good idea!!!

dlt-connector/src/client/BackendClient.ts
```diff
@@ -0,0 +1,88 @@
...
+import { LogError } from '@/server/LogError'
+
+const communityByForeign = gql`
+  query ($foreign: Boolean) {
+    community(foreign: $foreign) {
+      uuid
+      foreign
+      creationDate
+    }
+  }
```
clauspeterhuebner, line 17: Ah i see here is you usecase to request `community` with only a boolean as parameter.
But i think it will be necessary to split the requests from frontend/admininterface and from dlt-connector in different queries/graphql-methods because of security reasons for different users with different roles and rights.

This will also solve the problem to avoid using the 1st parameter `communityIdentifier` as type boolean, right?

dlt-connector/src/client/BackendClient.ts
```diff
@@ -0,0 +1,88 @@
...
+    if (!BackendClient.instance) {
+      BackendClient.instance = new BackendClient()
+    }
+    if (!BackendClient.instance.client) {
+      try {
+        BackendClient.instance.client = new GraphQLClient(CONFIG.BACKEND_SERVER_URL, {
+          headers: {
+            'content-type': 'application/json',
+          },
+          method: 'POST',
```
clauspeterhuebner, line 60: Please use for read-access only the GET-method instead of POST

dlt-connector/src/client/BackendClient.ts
```diff
@@ -0,0 +1,88 @@
...
+    })
+    if (errors) {
+      throw new LogError('error getting home community from backend', errors)
+    }
+    const communityDraft = new CommunityDraft()
+    communityDraft.uuid = data.community.uuid
+    communityDraft.foreign = data.community.foreign
+    communityDraft.createdAt = data.community.creationDate
+    return communityDraft
+  }
```
clauspeterhuebner, line 87: As you see here in your dlt-code it is helpfull to name the methods more expressive like this example shows `homeCommunityUuid` than to generalize all to one `community` method with a very complex parameter treatment for all variants.

So on my opinion this method expresses the opposite of your refactorings in the `util.communities.ts`-file.

clauspeterhuebner (changes requested):

backend/src/graphql/resolver/CommunityResolver.ts
```diff
@@ -41,41 +42,30 @@ export class CommunityResolver {
     return dbCommunities.map((dbCom: DbCommunity) => new Community(dbCom))
   }
 
-  @Authorized([RIGHTS.COMMUNITY_BY_UUID])
+  @Authorized([RIGHTS.COMMUNITIES])
```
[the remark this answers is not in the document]
einhornimmond: So everyone can list all communities, but only admins can load a specific community entry?
Why?

backend/src/graphql/resolver/util/communities.ts
```diff
@@ -1,65 +1,116 @@
...
+  foreign?: boolean | null,
+): FindOptionsWhere<DbCommunity> {
+  if (communityIdentifier === undefined && foreign === undefined) {
+    throw new LogError('one of communityIdentifier or foreign must be set')
+  }
+
+  const where: FindOptionsWhere<DbCommunity> = {}
+  // != null cover !== null and !== undefined
+  if (communityIdentifier != null) {
+    if (typeof communityIdentifier === 'boolean') {
```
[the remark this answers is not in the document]
einhornimmond: ok, then I will implement the other solution I have thought about.

backend/src/graphql/resolver/CommunityResolver.ts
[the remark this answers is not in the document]
clauspeterhuebner: It is a defensive strategy to allow only as much as necessary and not vica versa to allow all till we run in problems...
The process for login- and registry including the community-selection is still not productiv, so it must not be allowed to search or select for community with a dedicated role and its associated rights.

The only role/person who is allowed to enter the apiKey for a dedicated community will be an ADMIN.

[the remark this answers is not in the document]
clauspeterhuebner: It is a defensive strategy to allow only as much as necessary and not vica versa to allow all till we run in problems...
The process for login- and registry including the community-selection is still not productiv, so it must not be allowed to search or select for community with a dedicated role and its associated rights.

The only role/person who is allowed to enter the apiKey for a dedicated community will be an ADMIN.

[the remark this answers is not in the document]
einhornimmond: Ok, I have created a dedicated role for dlt-connector and changed RIGHT COMMUNITIES to COMMUNITY_BY_IDENTIFIER and add this RIGHT to admin and dlt-connector role.

backend/src/graphql/resolver/util/communities.ts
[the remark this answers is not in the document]
einhornimmond: I now use CommunityArgs as input so it is possible to set only foreign { foreign: true } or only community identifier or both. It make also getCommunityFindOptions shorter and maybe easier to read.

backend/src/graphql/resolver/util/communities.ts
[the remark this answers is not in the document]
einhornimmond: Double? Could please specify how you count?
src/graphql/resolver/util/communities.ts
Before (master branch): 65 Lines of Code, without a single line of comment
Current version here in #3289: 101 Lines of Code, 39 are comments lines, so 62 Lines of Executable Code.

It is a bit more complex, but why it is harder to maintain?

misusage of communityIdentifier is removed

Why you find it quite complex? I'm puzzled.
basicly it checks the type of communityIdentifier and build a specific and shorter mysql query as before.
- if it is a url it will only compared with url in db
- if it is a uuid4 it will only compared with communityUuid in db
- if it is a none of the above it will only compared with name in db
- if foreign is set, it will be compared also with foreign field in db

So no longer things like this:
```ts
const homeCommunity = await DbCommunity.findOne({
    where: [
      { foreign: false, communityUuid: communityIdentifier },
      { foreign: false, name: communityIdentifier },
      { foreign: false, url: communityIdentifier },
    ],
  })
  ```
  which will lead to this mysql where query: 
  ```mysql
  WHERE (
    (foreign = false AND communityUuid = :communityIdentifier)
    OR
    (foreign = false AND name = :communityIdentifier)
    OR
    (foreign = false AND url = :communityIdentifier)
)
  ```

File: backend/src/seeds/graphql/queries.ts (hunk 1 above)
Comment (einhornimmond) [reply backend/src/seeds/graphql/queries.ts:139]: This query is in the seed folder so it is used only for tests.
It is named getCommunityByUuidQuery. I don't see why I should add foreign for a query which will from the name here only query by uuid.
But maybe you are right and I better update the tests.

File: dlt-connector/src/client/BackendClient.ts (hunk 3 above)
Comment (einhornimmond) [reply dlt-connector/src/client/BackendClient.ts:18]: > This will also solve the problem to avoid using the 1st parameter communityIdentifier as type boolean, right?

I have changed it

File: dlt-connector/src/client/BackendClient.ts (hunk 4 above)
Comment (einhornimmond) [reply dlt-connector/src/client/BackendClient.ts:61]: ok

File: dlt-connector/src/client/BackendClient.ts (hunk 5 above)
Comment (einhornimmond) [reply dlt-connector/src/client/BackendClient.ts]: The function names should show what the function is doing.
But a disadvantage is that is easy to forget to update the name, if the function changed. 
homeCommunityUUid isn't quite correct any longer here, because it returns not only the community uuid. I have updated it now. 

my refactor in community had also the goal to make the mysql request shorter. If I had not put that in one generalized function, It would increased the Lines of Codes quite a bit.
Without putting it into getCommunityFindOptions I would have add the if else structure to reduce mysql code in isHomeCommunity, getCommunityUrl, isCommunityAuthenticated and getCommunityName.

Comment (einhornimmond) [reply dlt-connector/src/client/BackendClient.ts]: The community grapqhl request had this name because graphql is designed for call for different datas via the same queries. I'd like to use graphQL as intended to make maximal use of the additional ressource cost using it instead of REST or jsonRPC. But even in popular REST Apis like from youtube they are normally not many different requests and mostly with easy and short names. 
- https://developers.google.com/youtube/v3/docs?hl=de
- https://www.apollographql.com/blog/8-free-to-use-graphql-apis-for-your-projects-and-demos
- https://docs.github.com/de/graphql/overview/public-schema

File: backend/src/graphql/resolver/util/communities.ts (hunk 7 above)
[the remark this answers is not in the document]
Comment (einhornimmond) [reply backend/src/graphql/resolver/util/communities.ts]: When I'm not allowed to use if I cannot write an universal endpoint so I use homeCommunity and communityByIdentifier as separat endpoints.

File: backend/src/graphql/resolver/util/communities.ts [hunk 8]
```diff
@@ -1,65 +1,88 @@
...
-      { name: communityIdentifier },
-      { url: communityIdentifier },
-    ],
-  })
-  return community.url
+  return (await DbCommunity.findOneOrFail(findWithCommunityIdentifier(communityIdentifier))).url
 }
 
+/**
+ * TODO: Check if it is needed, because currently it isn't used at all
```
clauspeterhuebner, line 52: yes that's right, it is prepared for future checks to only allow selecting an authenticated ommunity

clauspeterhuebner (approved):

backend/src/graphql/enum/RoleNames.ts
```diff
@@ -5,6 +5,7 @@ export enum RoleNames {
   USER = 'USER',
   MODERATOR = 'MODERATOR',
   ADMIN = 'ADMIN',
+  DLT_CONNECTOR_ROLE = 'DLT_CONNECTOR_ROLE',
```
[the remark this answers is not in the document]
einhornimmond: Or you are right, I'll change it.

PR #3292, opened 2024-02-22
build(other): add config for nvm

einhornimmond:
## 🍰 Pullrequest

- add .nvmrc file for every module containing same node version like in docker
- with installed nvm you need only to call `nvm use` in the module folder to use the correct NodeJS version
- you need to install yarn for every node version with `npm i -g yarn`


admin/.nvmrc
```diff
@@ -0,0 +1 @@
+v14.17.0
```
clauspeterhuebner, line 1: Does it make sense to upgrade the version as soon as possible or why using such an old version?

backend/.nvmrc
```diff
@@ -0,0 +1 @@
+v18.7.0
```
clauspeterhuebner, line 1: I know this modul will work with V20.0.0 too.
We should plan an upgrade for all moduls as soon as possible, right?

frontend/.nvmrc
```diff
@@ -0,0 +1 @@
+lts/gallium
```
clauspeterhuebner, line 1: Is this a correct entry?
I expected a version-number instaed of a path-string...

clauspeterhuebner (approved):
Some questions and remarks possibly for further actions/tickets, so i approved it...

backend/.nvmrc
einhornimmond: Yes right, but in another pr.

frontend/.nvmrc
einhornimmond: Like I write, I copy the version from docker. 
The Frontend Docker use node:16-alpine3.14 which is the latest version from 16 named gallium, so this string does the same. Tell nvm to use the latest release from gallium.

PR #3324, opened 2024-06-19
feat(backend): speedup listTransactions

einhornimmond:
## 🍰 Pullrequest

- make TransactionResolver.transactionList faster, improvement on stage2.gradido.net from ~350 ms to ~100 ms with production data set copy
- calling gdt server for gdt sum parallel to rest of transactionList logic
  - calling seems to be needing at least 100 ms, independent of gdt server speed (return balance in 15 ms)
- adding INDEX for transactions.user_id improve speed for getLastTransaction and getTransactionList dramatic
  - getLastTransaction from ~127 ms to 4 ms
  - getTransactionList ~300 ms to 14 ms
- reduce log output from decimal in BalanceResolver.balance with added DecayLoggingView and BalanceLoggingView

### DIFF
--- a/backend/src/config/index.ts
+++ b/backend/src/config/index.ts
@@ -12,7 +12,7 @@ Decimal.set({
 })
 
 const constants = {
-  DB_VERSION: '0084-introduce_humhub_registration',
+  DB_VERSION: '0085-add_index_transactions_user_id',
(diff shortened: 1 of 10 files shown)


clauspeterhuebner (approved):
very good...

PR #3340, opened 2024-08-01
fix(admin): fix i18n initial setup, cleanup unused code and console.logs

MateuszMichalowski:
## 🍰 Pullrequest
Fixed i18n locale setup on first load. Cleaned `console.log` remaining in codebase.


unnunhexium (approved):
looks good, one check needs to be fixed

PR #3365, opened 2024-09-06
feat(frontend): vue3 migration

MateuszMichalowski:
PR contains codebase upgrade from vue 2 to vue 3 for both admin and frontend applications. It also contains all tests updated to match current implementation. e2e tests were also updated.


einhornimmond (approved):
It seems to work.

PR #3376, opened 2024-10-17
feat(frontend): map feature in vue 3

MateuszMichalowski:
Implementation of map feature for user selection of their location. Vue 2 implementation available on branch [3262](https://github.com/gradido/gradido/tree/3262-feature-capture-user-location)


unnunhexium (approved):
looks neat!

PR #3392, opened 2024-11-20
feat(backend): speed up gdt request with keep alive connections

einhornimmond:
make sure backend keep connections to gdt server, to reduce response time and prevent creating so many new connections, especially ssl handshake cost most of the request time


backend/src/apis/HttpRequest.ts
```diff
@@ -2,15 +2,21 @@
...
 import { backendLogger as logger } from '@/server/logger'
 
+const httpAgent = new Agent({ keepAlive: true })
+const httpsAgent = new HttpsAgent({ keepAlive: true })
+
 export const apiPost = async (url: string, payload: unknown): Promise<any> => {
   logger.trace('POST', url, payload)
   try {
-    const result = await axios.post(url, payload)
+    const result = await axios.post(url, payload, { httpAgent, httpsAgent })
```
clauspeterhuebner, line 19: Isn't it much more lightweight to simply use a 'header-config' with switched on 'connection: keep-alive' like it is done in the GmsClient, than introducing additional and heavier Agents which simply take care for keep-alive connections?

GMS-Client:
```
...
    const config = {
      headers: {
        accept: 'application/json',
        language: 'en',
        timezone: 'UTC',
        connection: 'keep-alive',
        authorization: apiKey,
      },
    }
    try {
      const result = await axios.post(baseUrl.concat(service), user, config)
```

```

einhornimmond: When you only set keep-alive in the header, you're politely telling the server to keep the connection open. However, Axios will still go ahead and create a new connection for the next request. It's like you're using a URLConnection in Java, but without a connection pool—each request is treated as a fresh start. Axios, much like an old-school Java HTTP client, needs an agent to manage connection reuse. Think of the agent as the connection pool manager, ensuring that Axios doesn't have to reconnect every time like it's starting from scratch.

einhornimmond: Better I fix your code too?

clauspeterhuebner: Ok, then i will fix it for GmsClient on my own to understand from my side the correct handling with a connection pooll agent...
I will approve this PR now.

clauspeterhuebner (approved):
I saw you still have adapt the code of GmsClient. Its fine for me, thx...

PR #3426, opened 2025-01-23
feat(admin): resubmission date localized

einhornimmond:
- localize resubmission date in admin with user date
- move code for getting current local object in separate composable

### DIFF
--- a/admin/src/composables/useDateLocale.js
+++ b/admin/src/composables/useDateLocale.js
@@ -1,0 +1,10 @@
+import { useI18n } from 'vue-i18n'
+import { de, enUS as en, fr, es, nl } from 'date-fns/locale'
+
+const locales = { en, de, es, fr, nl }
+
+export function useDateLocale() {
+  const { locale } = useI18n()
+  const dateLocale = locales[locale.value] || en
+  return dateLocale
+}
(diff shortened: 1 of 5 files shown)


admin/src/composables/useDateLocale.js
```diff
@@ -0,0 +1,10 @@
+import { useI18n } from 'vue-i18n'
+import { de, enUS as en, fr, es, nl } from 'date-fns/locale'
+
+const locales = { en, de, es, fr, nl }
+
+export function useDateLocale() {
+  const { locale } = useI18n()
+  const dateLocale = locales[locale.value] || en
```
clauspeterhuebner, line 8: if attribute 'locales' is a hard coded array of locale-values and for the object 'locale' it is not sure to have a correct initialized value, which will match one of the 'locales' entries, then an index-out-of-bount exception will occur, right?
So i think a check if 'locales' contains an entry with 'locale.value' will be more stabil...

clauspeterhuebner (approved):
...with one open question to avoid unexpected exception...

admin/src/composables/useDateLocale.js
einhornimmond: Do you know how you usually check in js if an object contain an attribute?
```js
let dateLocale
if (locales[locale.value] !== undefined) {
  dateLocale = locales[locale.value]
  } else {
  dateLocale = en
}
```
Shorter is:
```js
const dateLocale = locales[locale.value] || en
```
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_OR

PR #3432, opened 2025-01-28
fix(frontend): fix meta url from env.template

einhornimmond:
Add https:// to COMMUNITY_HOST to make META_URL a valid url with scheme


frontend/.env.template
```diff
@@ -13,7 +13,7 @@ COMMUNITY_DESCRIPTION=$COMMUNITY_DESCRIPTION
 COMMUNITY_SUPPORT_MAIL=$COMMUNITY_SUPPORT_MAIL
 
 # Meta
-META_URL=$COMMUNITY_HOST
+META_URL=https://$COMMUNITY_HOST
```
clauspeterhuebner, line 16:
```suggestion
META_URL=$URL_PROTOCOL://$COMMUNITY_HOST
```

clauspeterhuebner (approved):
approve with one suggestion...

PR #3433, opened 2025-01-28
feat(backend): correct user data transfer to gms

clauspeterhuebner:
### Issues

- None

### Todo

- [X] None

### DIFF
--- a/admin/src/config/index.js
+++ b/admin/src/config/index.js
@@ -14,7 +14,7 @@ const constants = {
 
 const version = {
   ADMIN_MODULE_PROTOCOL: process.env.ADMIN_MODULE_PROTOCOL ?? 'http',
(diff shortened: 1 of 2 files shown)


einhornimmond (approved):
looks good for me, but you could also write:
```js
const fn = pnLogic.getFirstName(user.gmsPublishName)
this.firstName = fn !== '' ? fn : null
``` 
as
```js
this.firstName =  pnLogic.getFirstName(user.gmsPublishName) || null
``` 
The same for lastName because 
the || operator returns the first truthy value. Since an empty string ("") is falsy, null is assigned instead.
https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Operators/Logical_OR

einhornimmond (approved):

PR #3451, opened 2025-02-27
feat(backend): 3347 modify overview page with humhub- and usersearch-card 

clauspeterhuebner:
### Issues

- None

### Todo

- [X] None


backend/src/graphql/model/User.ts
```diff
@@ -37,6 +41,8 @@ export class User {
       this.gmsPublishName = user.gmsPublishName
       this.humhubPublishName = user.humhubPublishName
       this.gmsPublishLocation = user.gmsPublishLocation
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
```
einhornimmond, line 44:
```suggestion
```
not needed any longer, because your argument is safe, isn't it? When I remove it, I don't get linting errors.

frontend/src/pages/Login.vue
```diff
@@ -112,6 +113,7 @@ const onSubmit = handleSubmit(async (values) => {
     })
     const { login: loginResponse } = result.data
     await store.dispatch('login', loginResponse)
+    store.commit('gmsActive', CONFIG.GMS_ACTIVE)
```
einhornimmond, line 116: Why store it in local store, when it already is in the config?
So when we disable GMS on server side and the user was still logged in, his frontend will try to show gms things, but the backend will deliver errors for all gms related requests?

einhornimmond, line 116: Why you store the config value in the local store of the browser?
So when the user was still logged in, while we rebuild the frontend with disabled gms, the users frontend will still try to render the gms card and ask the backend for gms login key? But the backend will throw an error, if it was disabled.

frontend/src/pages/Overview.vue
```diff
@@ -1,15 +1,24 @@
...
   <div class="overview">
-    <community-news />
+    <BCol>
+      <BRow>
+        <card-circles />
+      </BRow>
+      <BRow>
+        <card-user-search />
+      </BRow>
+    </BCol>
```
einhornimmond, line 10: Please check here if gms and humhub are allowed, to prevent unnecessary loading of components, if not needed. And especially unnecessary triggering the graphql requests when the backend will surly answer with an error. I have tried it, if I disable both in config and open the frontend in browser, it will issue two queries for one for humhub authentication, one for gms authentication.

frontend/src/pages/Login.vue
clauspeterhuebner: Ok, yes that's right.
The previous used control-attribut "gmsAllowed" still exists in the store, but this is a user-dependend value and not equal the global config setting. That's the reason, why i tried to add the "gmsActive" in the store.

But you are right, i will change and remove it frome the store and read the control-attribute directly from config instead of the store

frontend/src/pages/Overview.vue
clauspeterhuebner: That's a good idea and now with the wellknown controlling it will be hopefully easier to check this here in the overview.vue...

einhornimmond (approved):
looks ok for me

PR #3452, opened 2025-03-06
feat(admin): optimize contribution list for ai usage

einhornimmond:
- text changes
- prevent copy of private data first name, last name and # in admin contribution table
- make all contributions toggle able 
- clear contribution message text field after submit
- add code for showing humhub profile link in admin


admin/src/components/ContributionMessages/ContributionMessagesList.vue
```diff
@@ -57,15 +78,36 @@ const props = defineProps({
   },
 })
 
-const emit = defineEmits(['update-status', 'reload-contribution', 'update-contributions'])
+const emit = defineEmits([
+  'update-status',
+  'reload-contribution',
+  'update-contributions',
+  'search-for-email',
+])
```
clauspeterhuebner, line 86: if you define the emit-string here, why do you not use them in the emit-statements below?
for example in line 134 and 138

admin/src/components/Tables/OpenCreationsTable.vue
```diff
@@ -168,14 +174,26 @@ export default {
...
     },
   },
-  emits: ['update-contributions', 'reload-contribution', 'update-status', 'show-overlay'],
+  emits: [
+    'update-contributions',
+    'reload-contribution',
+    'update-status',
+    'show-overlay',
+    'search-for-email',
+  ],
```
clauspeterhuebner, line 183: the same here for predefined emit-strings and line 219?

clauspeterhuebner (approved):
Approved but with two questions about predefined emit-stirng without usage...

admin/src/components/ContributionMessages/ContributionMessagesList.vue
einhornimmond: https://vuejs.org/guide/components/events.html

admin/src/components/Tables/OpenCreationsTable.vue
einhornimmond: https://vuejs.org/guide/components/events.html

PR #3456, opened 2025-03-14
feat(admin): add ai chat

einhornimmond:
- add ai chat in admin backend for easier use of chatgpt for contribution check
- update backend typescript version to get npm openai working
- update default nodejs version up to version 18.20.7 because needed by newer typescript version


clauspeterhuebner (approved):
I do not understand the logic to store an aithread-entry with userId and creationdate but without chat-message(s). We should talk about this topic in our next meeting...

PR #3458, opened 2025-03-15
feat(backend): upgrade typescript

einhornimmond:
- upgrade to higher typescript version
- upgrade to higher eslint version
- upgrade to higher ts-node version
- fix new eslint errors and warnings

### DIFF
--- a/backend/src/apis/gms/model/GmsUser.ts
+++ b/backend/src/apis/gms/model/GmsUser.ts
@@ -49,14 +49,16 @@ export class GmsUser {
     if (
       user.gmsAllowed &&
       user.alias &&
-      user.gmsPublishName === PublishNameType.PUBLISH_NAME_ALIAS_OR_INITALS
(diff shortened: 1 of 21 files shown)


backend/src/apis/gms/model/GmsUser.ts
```diff
@@ -49,14 +49,16 @@ export class GmsUser {
     if (
       user.gmsAllowed &&
       user.alias &&
-      user.gmsPublishName === PublishNameType.PUBLISH_NAME_ALIAS_OR_INITALS
+      (user.gmsPublishName as PublishNameType) === PublishNameType.PUBLISH_NAME_ALIAS_OR_INITALS
```
clauspeterhuebner, line 52: What has been changed to add this explicit "as PublishNameType"?
It seems to be necessary on every usage of enums in the following files...
It wasn't necessary before...

clauspeterhuebner (approved):
approved, but with the question about the changed enum-usage...
Is it caused by the version-upgrades or is an explicit change the reason?

PR #3467, opened 2025-04-11
feat(workflow): x-cross tx per link

clauspeterhuebner:
### Issues

- None

### Todo

- [X] None


backend/src/federation/client/1_0/logging/PublicCommunityInfoLogging.view.ts
```diff
@@ -11,7 +11,7 @@ export class PublicCommunityInfoLoggingView extends AbstractLoggingView {
     return {
       name: this.self.name,
       description: this.self.description,
-      creationDate: this.dateToString(this.self.creationDate),
+      creationDate: this.self.creationDate,
```
einhornimmond, line 14: The creationDate Attribute from PublicCommunityInfo is declared as Date so this function dateToString from AbstractLoggingView 
```ts
public dateToString(date: Date | undefined | null): string | undefined {
    if (date) {
      return date.toISOString()
    }
    return undefined
  }
  ```
  is used to make it pretty for logging.
  Or is it a bug in PublicCommunityInfo that it contains creationDate sometimes as string? That it is maybe better, to update the interface, ins't it?

File: frontend/src/locales/es.json [hunk 2]
```diff
@@ -198,6 +198,7 @@
     "decay-14-day": "Disminución gradual por 14 días",
     "delete-the-link": "Eliminar el enlace?",
     "deleted": "El enlace ha sido eliminado!",
+    "disbured": "Disbursement of the Link-Ammount initiated! The transfer of {n} GDD into your account will be completed shortly.",
```
einhornimmond, line 201: If you don't want to translate it, please don't add it, it will automatically fall back to English.

backend/src/graphql/resolver/TransactionLinkResolver.ts
```diff
@@ -364,6 +393,104 @@ export class TransactionLinkResolver {
...
+    @Arg('gradidoId') gradidoId: string,
+    @Arg('senderCommunityUuid') senderCommunityUuid: string,
+    @Arg('senderCommunityName') senderCommunityName: string,
+    @Arg('recipientCommunityUuid') recipientCommunityUuid: string,
+    @Arg('code') code: string,
+    @Arg('amount') amount: string,
+    @Arg('memo') memo: string,
+    @Arg('firstName', { nullable: true }) firstName?: string,
+    @Arg('alias', { nullable: true }) alias?: string,
+    @Arg('validUntil', { nullable: true }) validUntil?: string,
```
einhornimmond, line 408: Thanks for the work on this! One thing I'd suggest is to consider consolidating the many function parameters into a single DTO or input object. This will improve readability, reduce the risk of mistakes when passing so many arguments around, and make future maintenance easier.

Having so many individual parameters can get quite verbose and error-prone, especially as the codebase grows. Using an input type would also align better with GraphQL best practices and keep the resolver signature cleaner.

backend/src/graphql/resolver/TransactionLinkResolver.ts
```diff
@@ -364,6 +393,104 @@ export class TransactionLinkResolver {
...
+    @Arg('senderGradidoId') senderGradidoId: string,
+    @Arg('recipientCommunityUuid') recipientCommunityUuid: string,
+    @Arg('recipientCommunityName') recipientCommunityName: string,
+    @Arg('recipientGradidoId') recipientGradidoId: string,
+    @Arg('recipientFirstName') recipientFirstName: string,
+    @Arg('code') code: string,
+    @Arg('amount') amount: string,
+    @Arg('memo') memo: string,
+    @Arg('validUntil', { nullable: true }) validUntil?: string,
+    @Arg('recipientAlias', { nullable: true }) recipientAlias?: string,
```
einhornimmond, line 456: Thanks for the work on this! One thing I'd suggest is to consider consolidating the many function parameters into a single DTO or input object. This will improve readability, reduce the risk of mistakes when passing so many arguments around, and make future maintenance easier.

Having so many individual parameters can get quite verbose and error-prone, especially as the codebase grows. Using an input type would also align better with GraphQL best practices and keep the resolver signature cleaner.

frontend/src/components/LinkInformations/RedeemCommunitySelection.vue
```diff
@@ -0,0 +1,182 @@
+<template>
+  <div
+    :link-data="linkData"
+    :redeem-code="redeemCode"
+    :is-contribution-link="isContributionLink"
+    :is-redeem-jwt-link="isRedeemJwtLink"
+    class="redeem-community-selection"
+  >
```
einhornimmond, line 8: What is the function, the reason of that? Div cannot do any thing with the parameters, only showing in browser:
```html
<div link-data="[object Object]" redeem-code="6cf60204cd170196d2c8bdef" is-contribution-link="false" is-redeem-jwt-link="false" class="redeem-community-selection" is-transaction-link-loaded="true">
```

einhornimmond (approved):
I Approve it, it should work, but in my opinion it is really verbose.

PR #3472, opened 2025-04-26
refactor(other): use biome instead of eslint

einhornimmond:
### Replace ESLint and Prettier with Biome for Improved Performance and Maintenance
This pull request introduces the following changes to improve development performance and streamline code maintenance:

- Replaced ESLint and Prettier with Biome as the unified linter and formatter in all backend modules.
- Removed all existing ESLint-related comments and modules to avoid confusion and redundancy in all backend modules.
- Refactored code where necessary to adhere to Biome’s stricter validation rules.
- Added a biome.json configuration file at the project root to ensure seamless integration with Visual Studio Code.
- Configured Visual Studio Code to use Biome as the default linter and formatter for js and ts.
- Enabled order imports on save for consistent and clean code structure.

#### Rationale:
The primary motivation for this change is a substantial performance improvement.
Linting and formatting backend code that previously took approximately 22.95 seconds now complete in 0.17 seconds, representing a relative time saving of over 99%.
This enhancement significantly reduces waiting times during development, contributing to a more efficient and enjoyable workflow for all contributors.

In addition to performance improvements, Biome also offers a superior developer experience regarding error reporting.
Whereas ESLint typically only indicated the presence of an error in a specific line (often requiring additional investigation), Biome provides:

A detailed view of the problematic code section.

A concise explanation of the issue.

If possible, a clear and safe fix suggestion, making corrections significantly easier to understand.

 ```bash
 /home/gradido/code/gradido/gradido_partly_upgrade/federation/src/graphql/util/validateAlias.ts:30:3 lint/style/useBlockStatements  FIXABLE  ━━━━━━━━━━

  ✖ Block statements are preferred in this position.
  
    28 │   if (alias.length > 20) throw new LogError('Given alias is too long', alias)
    29 │   if (!alias.match(VALID_ALIAS_REGEX)) throw new LogError('Invalid characters in alias', alias)
  > 30 │   if (RESERVED_ALIAS.includes(alias.toLowerCase()))
       │   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  > 31 │     throw new LogError('Alias is not allowed', alias)
       │     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    32 │   const aliasInUse = await DbUser.find({
    33 │     where: { alias: Raw((a) => `LOWER(${a}) = "${alias.toLowerCase()}"`) },
  
  ℹ Unsafe fix: Wrap the statement with a `JsBlockStatement`
  
    28 28 │     if (alias.length > 20) throw new LogError('Given alias is too long', alias)
    29 29 │     if (!alias.match(VALID_ALIAS_REGEX)) throw new LogError('Invalid characters in alias', alias)
    30    │ - ··if·(RESERVED_ALIAS.includes(alias.toLowerCase()))
    31    │ - ····throw·new·LogError('Alias·is·not·allowed',·alias)
       30 │ + ··if·(RESERVED_ALIAS.includes(alias.toLowerCase()))·{
       31 │ + ····throw·new·LogError('Alias·is·not·allowed',·alias)
       32 │ + ··}
    32 33 │     const aliasInUse = await DbUser.find({
    33 34 │       where: { alias: Raw((a) => `LOWER(${a}) = "${alias.toLowerCase()}"`) },
```

### CODE REVIEW

File: backend/src/graphql/resolver/util/authenticateGmsUserPlayground.ts [hunk 1]
```diff
@@ -7,7 +7,7 @@ import { backendLogger as logger } from '@/server/logger'
 import { ensureUrlEndsWithSlash } from '@/util/utilities'
 
 export async function authenticateGmsUserPlayground(
-  apiKey: string,
+  _apiKey: string,
```
clauspeterhuebner, line 10: whats the motivation to set a leading underscore?
normally the apikey should be used as parameter for the gms-request, but currently the authentication is deactivated. That's the reason why the apiKey isn't in use here.

backend/src/util/virtualTransactions.ts
```diff
@@ -33,12 +32,12 @@ const defaultModelFunctions = {
...
   balance: Decimal,
   amount: Decimal,
-  holdAvailableAmount: Decimal,
+  _holdAvailableAmount: Decimal,
   decay: Decimal,
   createdAt: Date,
   validUntil: Date,
   user: User,
-  previousBalance: Decimal,
+  _previousBalance: Decimal,
```
clauspeterhuebner, line 40: whats the intention or rule to add a prefixed underscore?

clauspeterhuebner (approved):
i have not viewed all 436 files, but quite the half looks more convenient than the eslint format...
i made some remarks, but i will agree to use biome from now on.

backend/src/apis/humhub/compareHumhubUserDbUser.ts
```diff
@@ -1,23 +1,37 @@
-import { User } from '@entity/User'
+import type { User } from '@entity/User'
 
 import { Account } from './model/Account'
```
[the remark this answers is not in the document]
einhornimmond: > TypeScript allows specifying a type keyword on an import to indicate that the import doesn’t exist at runtime. This allows compilers to safely drop imports of types without looking for their definition. This also ensures that some modules are not loaded at runtime.

> The rule ensures that all imports used only as a type use a type-only import. It also groups inline type imports into a grouped import type.

> If you use the TypeScript Compiler (TSC) to compile your code into JavaScript, then you can disable this rule, as TSC can remove imports only used as types. However, for consistency and compatibility with other compilers, you may want to enable this rule. 

https://biomejs.dev/linter/rules/use-import-type/

I disable it for now, and remove all type imports, but we maybe will need it, when we change the slow tsc TypeScript compiler with bun or esbuild.

backend/src/graphql/resolver/util/authenticateGmsUserPlayground.ts
einhornimmond: > Disallow unused variables.

> There is an exception to this rule: variables that starts with underscore, e.g. let _something;.

> The pattern of having an underscore as prefix of a name of variable is a very diffuse pattern among programmers, and Biome decided to follow it.

https://biomejs.dev/linter/rules/no-unused-variables/
https://biomejs.dev/linter/rules/no-unused-function-parameters/

It is to mark the variable as intentional unused.

backend/src/util/virtualTransactions.ts
einhornimmond: It is to mark the variable as intentional unused.
But in this case we may remove them.. later in another refactoring round

PR #3479, opened 2025-05-01
refactor(other): use esbuild instead of tsc

einhornimmond:
After several intense days refining the Docker and build setup, things are finally falling into place.

based on #3478 
- optimize workflows 
- using esbuild for building all backend modules
- tried also bun, and write issues with bun into module readme
- use tsx for running typescript directly expect in backend and federation where it didn't work because of graphql, use ts-node with swc instead
- add build:bun and start:bun script entries in package.json where it is possible to use bun
- add missing types in database entities needed for esbuild
- update graphql schema to use modern import approach, include via path is deprecated in later versions of graphql
- clear double css generation call in frontend
- overhauled the Docker configuration using Bun, esbuild, and Turborepo, significantly enhancing speed and size especially in production
- wait until database is ready direct in backend modules, to remove need for external waiting code/scripts
- move lint with biome in seperat workflow to install biome once and use it on all modules
- refactor password worker to work with esbuild bundled build
- update deployment/bare_metal/start.sh to stop immediately on error


database/entity/0003-login_server_tables/LoginEmailOptIn.ts
```diff
@@ -6,21 +6,21 @@ export class LoginEmailOptIn extends BaseEntity {
...
 
   @Column({ name: 'verification_code', type: 'bigint', unsigned: true, unique: true })
   verificationCode: BigInt
 
-  @Column({ name: 'email_opt_in_type_id' })
+  @Column({ name: 'email_opt_in_type_id', type: 'int', unsigned: true, nullable: false })
   emailOptInTypeId: number
 
-  @Column({ name: 'created', default: () => 'CURRENT_TIMESTAMP' })
+  @Column({ name: 'created', type: 'datetime', default: () => 'CURRENT_TIMESTAMP' })
```
clauspeterhuebner, line 18: for a correct initialisation and setting of the ms, the type and initial value should be define with the accuracy of 3 like datetime(3) like it is described https://mariadb.com/kb/en/microseconds-in-mariadb
This is necessary and defined in gradido as standard for all created_at columns for correct sorting base on datetime columns especially to distingue entries created in the same second

database/entity/0038-add_contribution_links_table/ContributionLink.ts
```diff
@@ -77,10 +77,10 @@ export class ContributionLink extends BaseEntity {
   @Column({ name: 'created_at', type: 'datetime', default: () => 'CURRENT_TIMESTAMP' })
   createdAt: Date
 
-  @DeleteDateColumn({ name: 'deleted_at' })
+  @DeleteDateColumn({ name: 'deleted_at', type: 'datetime' })
   deletedAt: Date | null
```
clauspeterhuebner, line 82: awareness of milliseconds...

database/src/prepare.ts
```diff
@@ -10,19 +10,25 @@ export enum DatabaseState {
...
-      user: CONFIG.DB_USER,
-      password: CONFIG.DB_PASSWORD,
-    })
-  } catch (e) {
-    // biome-ignore lint/suspicious/noConsole: no logger present
-    console.log('could not connect to database server', e)
-    return null
+export async function connectToDatabaseServer(
+  maxRetries = 15,
+  delayMs = 500,
```
clauspeterhuebner, line 15: same hardcoded values like above. they should be defined once and used as constants.

dht-node/src/typeorm/DBVersion.ts
```diff
@@ -2,6 +2,33 @@ import { Migration } from 'database'
 
 import { logger } from '@/server/logger'
 
+import { CONFIG } from '@/config'
+import { Connection as DbConnection } from 'typeorm'
+import { connection as connectionFunc } from './connection'
+
+async function checkDBVersionUntil(maxRetries = 15, delayMs = 500): Promise<DbConnection> {
```
clauspeterhuebner, line 9: referre the default values from constant definitions...

federation/src/graphql/api/1_1/schema.ts
```diff
@@ -0,0 +1,9 @@
+import { NonEmptyArray } from 'type-graphql'
+import { AuthenticationResolver } from '../1_0/resolver/AuthenticationResolver'
+import { PublicCommunityInfoResolver } from '../1_0/resolver/PublicCommunityInfoResolver'
+import { SendCoinsResolver } from '../1_0/resolver/SendCoinsResolver'
```
clauspeterhuebner, line 4: are relative imports allowed now?
eslint often not allows them

federation/src/graphql/schema.ts
```diff
@@ -8,7 +8,7 @@ import { DecimalScalar } from './scalar/Decimal'
 
 export const schema = async (): Promise<GraphQLSchema> => {
   return await buildSchema({
-    resolvers: [getApiResolvers()],
+    resolvers: getApiResolvers(),
```
clauspeterhuebner, line 11: this will not allow several api versions in parallel, right?

frontend/Dockerfile
```diff
@@ -9,16 +9,20 @@ ENV DOCKER_WORKDIR="/app"
 ## We Cannot do `$(date -u +'%Y-%m-%dT%H:%M:%SZ')` here so we use unix timestamp=0
 ENV BUILD_DATE="1970-01-01T00:00:00.00Z"
 ## We cannot do $(npm run version).${BUILD_NUMBER} here so we default to 0.0.0.0
+# TODO: get the actually git commit hash into docker
 ARG BUILD_VERSION
-ENV BUILD_VERSION=${BUILD_VERSION}
+ENV BUILD_VERSION=${BUILD_VERSION:-'broken'}
 ARG BUILD_COMMIT
-ENV BUILD_COMMIT=${BUILD_COMMIT}
+ENV BUILD_COMMIT=${BUILD_COMMIT:-'deadbeefdeadbeefdeadbeefdeadbeefdeadbeef'}
```
clauspeterhuebner, line 16: confusing default settings...

clauspeterhuebner (changes requested):

federation/src/graphql/api/1_1/schema.ts
einhornimmond: In backend eslint was configured to disallow them, but not in federation. In Federation on multiple places relative imports are used.

clauspeterhuebner (approved):

PR #3483, opened 2025-05-13
refactor(frontend): contribution page and some graphql code in backend used by the page

einhornimmond:
- replace `deadbeef` with `decafcab`
- replace `status` with `contributionStatus` to have the same name in entity, graphql and frontend
- replace `creationDate` with `contributionDate` to have the same name in entity, graphql and frontend
- move user attributes in Contribution into separat user object
- move email into user.emailContact
- reduce code duplication in Contribution with extending UnconfirmedContribution and inherit attributes both share
- use user direct from dbContribution instead of separat parameter, auto load with Field-Resolver when missing and requested
- use naming schema with db as prefix for entity objects in Contribution, ContributionMessage and UnconfirmedContribution
- move currentPage, pageSize and order into Pagination object
- use small, specialized functions for listContributions and listAllContributions for getting data from db rather than extend the already complex findContributions further to support 
- remove tests and parameter in resolver for listContributions and listAllContributions with statusFilter, because it isn't used at all from frontend 
- move sort logic for listContributions from frontend into backend
- add field resolver for auto load messages if missing and requested
- change emits in frontend to make them more clear 
- refactor mixed collapse via bootstrap js and vue mechanics and favor to only use vue mechanics to prevent jumping and inconsistend behaviour
- split logic from ContributionForm into ContributionEdit, ContributionCreate and ContributionForm for better code understanding
- split logic from ContributionList into ContributionList and ContributionListAll also from ContributionListItem into ContributionListItem and ContributionListAllItem
- move paginator logic into separate component
- adapt test to changes
- fix warnings in frontend test which where changed
- remove not used props from ContributionListItem and ContributionListAllItem and also from there respective graphql queries
- move icon and variant picking logic for conritbution state into a composable, use only contributionState for deciding and not the date fields any more
- use explicit function to decide which NavCommunity Button is showed as active, to work with new optional page argument in query
- add Polling for ContributionList and ContributionListAll and a new config option AUTO_POLL_INTERVAL for that, default = 0, disabled
- move most logic from pages/Community in Components
- make Tabs on Community Page lazy so they reload the component on each tab change
- use cache for listContributions and listAllContributions so first data from cache will be used and trigger request to backend and update data if they changed
- make deploy script more robust against problems with restarting nginx after to many runs in short time
- adapt to path changes for orm logger path, the new root path is `<module>/build` instead of `<module>`


frontend/src/pages/Community.vue
clauspeterhuebner, line 1: Can we rename this file to express more that it has nothing to do with 'gradido communities' but with 'contributions inside the local community'?
I was very confused in the first run...

frontend/src/routes/routes.js
```diff
@@ -68,7 +68,7 @@ const routes = [
     },
   },
   {
-    path: '/community/:tab',
+    path: '/community/:tab/:page?',
```
clauspeterhuebner, line 71: i would prefer the same renaming for the path to express more that it has nothing to do with 'gradido communities' but with 'contributions inside the local community'?
I was very confused in the first run...

clauspeterhuebner (approved):
i approved it with 3 suggestions about a renaming refactory, because of confusions in the first run...

PR #3486, opened 2025-05-19
feat(frontend): disable cross community tx redeem in frontend

einhornimmond:
### Issues

- None

### Todo

- [X] None


frontend/src/pages/TransactionLink.vue
```diff
@@ -207,8 +212,12 @@ const itemType = computed(() => {
...
     }
   }
-  // console.log('TransactionLink.itemType...last return= REDEEM_SELECT_COMMUNITY')
-  return 'REDEEM_SELECT_COMMUNITY'
+  if (CONFIG.CROSS_TX_REDEEM_LINK_ACTIVE) {
+    // console.log('TransactionLink.itemType...last return= REDEEM_SELECT_COMMUNITY')
+    return 'REDEEM_SELECT_COMMUNITY'
+  } else {
+    return 'LOGGED_OUT'
+  }
```
clauspeterhuebner, line 220: this expression will only work on the sender-community, but on the recipient-community with an activ cross-tx-redeem the return-value VALID must be supported too.

clauspeterhuebner (changes requested):

frontend/src/pages/TransactionLink.vue
einhornimmond: I don't understand. I didn't touch the VALID return value.
And as I understand the code, it work basically the same as in the version before your update, except that it now ignore store.state.token and store.state.tokenTime

clauspeterhuebner: Ouh, ok sorry! It was my misstake...
i simple had a look on the changed statement without opening the whole code. The changed statement looks like it will only check for the config property and only decides between REDEEM_SELECT_COMMUNITY and LOGGED_OUT.
You're right the code supports all necessary constellations without any necessary changes after your adaption.

clauspeterhuebner (approved):

PR #3491, opened 2025-05-20
feat(frontend): use grass as faster alternative to sass

einhornimmond:
- use grass in frontend as faster alternative to sass, if installed 
- install grass in deployment script
- remove old stuff from frontend README and add install instructions for grass
- use lightningcss in admin and frontend instead of sass for css
- tested on linux and windows


frontend/README.md
clauspeterhuebner, line 1: Why removing all the detailed info about issues, testusers, pages, overviews etc. here?
Or are they shifted in an other documentation?

clauspeterhuebner (changes requested):

clauspeterhuebner (approved):

PR #3492, opened 2025-05-20
feat(other): log4js default config generator

einhornimmond:
The great benefit from log4js-config.json is that everyone working with the code, can write his own, without touching the code. So for example I can focus on apollo logs, where my college is focusing on backend logs.
Currently we have nearly the same copy from log4js-config.json all over the place. For me it is a bit confusing to finds options which aren't used from current module. For example federation define a backend logger, but don't use it, it's define additional a federation logger and use this instead. 
I have write a generator based on the log4js-config.json in backend to give a easy entry for generating a default config for each module and add log4js-config.json to .gitignore.
So if no log4js-config.json is present, a default one will be generated on first module run. 
So every one can have there one config, without fearing a overwrite by git pull.

[ ] use in all backend modules
[ ] check possibility to use one log4js-config.json file for all backend modules


clauspeterhuebner:
It seems there is a log4js-config problem for the federation log-file-names. The api-version and port is not replaced any more for the file creation, but it looks like the mechanism for zipping the files use the correct replacements...

grafik

einhornimmond:
Implemented in other pr

PR #3497, opened 2025-05-22
feat(other): disable index html caching, reenable limits

einhornimmond:
### Changes
- mark index.html as not cachable
- mark js and css as 1 year cachable, not changing.. so browser don't need to check ever
- mark all other assets as 1 week caching.. maybe we update a img some times
- enable and adjust limits 
- add limit status to nginx log so we can check if user will be blocked because of the limits
- fix some issues with deploy test and deploy script while searching for error I have made, it was a missing ; in gradido.conf.template
- add Readme.md to deployment/bare_metal with nginx rate limit information and notice for vserver
- reduce memory usage with starting node directly, without turbo or cross-env, ~150 MB RAM less

**TODO**: reduce double code in gradido.conf.template and gradido.conf.ssl.template to prevent errors which occure only with ssl or without  #3498

### 🔒 NGINX Rate & Connection Limits Overview

| Path                        | Zone     | Rate Limit     | Burst | Max Connections | Notes                          |
|----------------------------|----------|----------------|-------|------------------|--------------------------------|
| `/`                        | frontend | 15 requests/s  | 150   | 60               | Public frontend                |
| `/admin`                   | frontend | 15 requests/s  | 30    | 20               | Admin frontend                 |
| `/graphql`                 | backend  | 20 requests/s  | 40    | 20               | Main backend GraphQL API       |
| `/hook`                    | backend  | 20 requests/s  | 20    | 20               | Internal backend webhooks      |
| `/hooks/`                  | backend  | 20 requests/s  | 20    | 20               | Reverse proxy for webhooks     |
| `/api/<version>` | api   | 30 requests/s  | 60    | 30               | Federation GraphQL API            |

- `<version>`: placeholder for federation api version
- All zones use `$binary_remote_addr` for client identification.
- `nodelay` ensures burst requests are not delayed (they are either accepted or rejected).
- Global connection zone: `limit_conn_zone $binary_remote_addr zone=addr:10m;`

This setup helps protect public and internal interfaces from abuse, while ensuring smooth parallel access during high load periods (e.g., UI builds or cluster sync).

These limits work like a traffic cop at each route:
- **Rate limits** (`limit_req`) define how many requests per second a single client can send.
- **Burst values** allow short spikes without blocking – like a temporary buffer.
- **Connection limits** (`limit_conn`) cap how many concurrent connections a single IP can keep open.

Each route (frontend, backend, API, etc.) has its own configuration depending on its expected traffic pattern and sensitivity. For example:
- The public frontend allows higher bursts (many assets load at once).
- The GraphQL backend and admin interfaces are more tightly controlled.

This ensures fairness, avoids accidental DoS scenarios, and keeps overall latency low, even under high usage.

### DIFF
--- a/deployment/bare_metal/start.sh
+++ b/deployment/bare_metal/start.sh
@@ -305,7 +305,11 @@ pm2 start --name gradido-backend \
 
 pm2 save
 if [ ! -z $FEDERATION_DHT_TOPIC ]; then
-  pm2 start --name gradido-dht-node "turbo dht-node#start --env-mode=loose" -l $GRADIDO_LOG_PATH/pm2.dht-node.$TODAY.log --log-date-format 'YYYY-MM-DD HH:mm:ss.SSS'
+  pm2 start --name gradido-dht-node \
+    "env TZ=UTC NODE_ENV=production node ./build/index.js" \
+    --cwd $PROJECT_ROOT/dht-node \
+    -l $GRADIDO_LOG_PATH/pm2.dht-node.$TODAY.log \
+    --log-date-format 'YYYY-MM-DD HH:mm:ss.SSS'
   pm2 save
 else
   log_step "====================================================================="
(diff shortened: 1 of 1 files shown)


clauspeterhuebner (approved):
congratulations on solving this so quickly...
i suggest to copy the description details of this PR in a technical document of the gradido project, because as soon as the PR is closed the description will not present and accessable like the general documents are.

PR #3507, opened 2025-06-20
refactor(shared): add shared module

einhornimmond:
- add shared module
- use validateAlias from user as example
- move mysql query from validate alias to database
- test database queries with vitest 
- test shared code with bun test


shared/src/schema/user.schema.ts
```diff
@@ -0,0 +1,29 @@
...
+  'gast',
+  'gdd',
+  'gradido',
+  'guest',
+  'home',
+  'root',
+  'support',
+  'temp',
+  'tmp',
+  'tmp',
```
clauspeterhuebner, line 17: double tmp definition

shared/src/schema/user.schema.ts
```diff
@@ -0,0 +1,29 @@
...
+  'user',
+  'usr',
+  'var',
+]
+
+export const aliasSchema = string()
+  .min(3, 'Given alias is too short')
```
clauspeterhuebner, line 24: is "min(3..." equals <3 or <=3?
i simply ask to ensure the correct edges with the correct message.

shared/src/schema/user.schema.ts
```diff
@@ -0,0 +1,29 @@
...
+  .max(20, 'Given alias is too long')
```
clauspeterhuebner, line 25: is "max(20..." equals >20 or >=20?
i simply ask to ensure the correct edges with the correct message.

clauspeterhuebner (changes requested):

database/src/queries/communities.test.ts
```diff
@@ -0,0 +1,40 @@
...
+const db = AppDatabase.getInstance()
+
+beforeAll(async () => {
+  await db.init()
+})
+afterAll(async () => {
+  await db.destroy()
+})
+
+describe('community.queries', () => {
```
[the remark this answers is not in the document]
einhornimmond: Yes, my goal is to shift all queries which are complex enough to be tested into database and move the test them here also without using a apollo instance for this to speed up testing.
getHomeCommunity is a bit special because it is really simple, but it encode business-logic ( and is used more than one times ). When we maybe change the db structure it would be nice to need only change this and don't think where I use the assumption that the home Community is foreign: false 
Maybe we will host more than one community some time in future in one db/instance?

shared/src/schema/user.schema.ts
einhornimmond: I have added a test for this, this is a good use case for a test, to checkout the borders. 
the tests show: 3 character is ok, 2 is not enough
```ts
it('2 characters is not ok', () => {
      expect(() => aliasSchema.parse('Bi')).toThrowError()
    })
    it('3 characters is ok', () => {
      expect(() => aliasSchema.parse('Bib')).not.toThrowError()
    })    
```
I had removed the test, because I don't think we need test for used libs that they work like they should. 
But we can test that they work like we expect.. if the semantic isn't clear enough, good point.

clauspeterhuebner (approved):

PR #3508, opened 2025-06-27
refactor(federation): move code for checking pending transactions

einhornimmond:
- move code for checking pending transactions into own query + test


clauspeterhuebner (approved):
very good...

PR #3549, opened 2025-10-07
fix(backend): check for openai thread timeout

einhornimmond:
- add updatedAt field to openai_threads database table
- update updateAt field with each new messages send to openai
- check on resume openai thread if it is older (updatedAt) as 60 days and when so, delete and trigger new thread creation
- catch thread resume error on openai side and return empty message list to admin, so it will start a new thread

### DIFF
--- a/database/src/entity/OpenaiThreads.ts
+++ b/database/src/entity/OpenaiThreads.ts
@@ -5,10 +5,10 @@ export class OpenaiThreads extends BaseEntity {
   @PrimaryColumn({ type: 'char', length: 30 })
   id: string
 
-  @CreateDateColumn({ type: 'timestamp', default: () => 'CURRENT_TIMESTAMP' })
+  @CreateDateColumn({ type: 'timestamp' })
   createdAt: Date
 
-  @UpdateDateColumn({ type: 'timestamp', default: () => 'CURRENT_TIMESTAMP', onUpdate: 'CURRENT_TIMESTAMP' })
+  @UpdateDateColumn({ type: 'timestamp' })
   updatedAt: Date
 
   @Column({ name: 'user_id', type: 'int', unsigned: true })


database/migration/migrations/0094-openai_threads_add_updated_at.ts
```diff
@@ -0,0 +1,10 @@
+
+export async function upgrade(queryFn: (query: string, values?: any[]) => Promise<Array<any>>) {
+  await queryFn(
+    'ALTER TABLE `openai_threads` ADD COLUMN `updatedAt` timestamp NULL DEFAULT CURRENT_TIMESTAMP AFTER `createdAt`;'
```
clauspeterhuebner, line 4:
```suggestion
    'ALTER TABLE `openai_threads` ADD COLUMN `updatedAt` timestamp(3) NULL DEFAULT CURRENT_TIMESTAMP(3) AFTER `createdAt`;'
```

database/src/entity/OpenaiThreads.ts
```diff
@@ -8,6 +8,9 @@ export class OpenaiThreads extends BaseEntity {
   @CreateDateColumn({ type: 'timestamp', default: () => 'CURRENT_TIMESTAMP' })
   createdAt: Date
 
+  @UpdateDateColumn({ type: 'timestamp', default: () => 'CURRENT_TIMESTAMP', onUpdate: 'CURRENT_TIMESTAMP' })
+  updatedAt: Date
```
clauspeterhuebner, line 12:
```suggestion
  @UpdateDateColumn({ type: 'timestamp', default: () => 'CURRENT_TIMESTAMP(3)', onUpdate: 'CURRENT_TIMESTAMP(3)' })
  updatedAt: Date
```

clauspeterhuebner (approved):

PR #3554, opened 2025-10-12
feat(workflow): adjust for new bun version

einhornimmond:
- adjustments needed for new bun 1.3
- bun is more strict with node-modules used but not declared in own package.json, so add all direct used modules in package.json
- fix instable test with community id depending on exact ids, now reset auto-increment counter in db before CommunityResolver.test
- add bun config for enable old default behaviour hosting [Read more](https://bun.com/blog/bun-v1.3#isolated-installs-are-now-the-default-for-workspaces), else github workflows are breaking
- use explicit bun version in github workflows
- use explicit bun version in Dockerfile
- use explicit bun version in deployment

### DIFF
--- a/deployment/bare_metal/install-missing-deps.sh
+++ b/deployment/bare_metal/install-missing-deps.sh
@@ -28,7 +28,7 @@ fi
 
 # check for some tools and install them, when missing
 # bun https://bun.com/install, faster packet-manager as yarn
-BUN_VERSION_FILE="$SCRIPT_DIR/.bun-version"
+BUN_VERSION_FILE="$PROJECT_ROOT/.bun-version"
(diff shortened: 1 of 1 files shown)


backend/src/graphql/resolver/CommunityResolver.test.ts
```diff
@@ -48,6 +48,9 @@ beforeAll(async () => {
   query = testEnv.query
   con = testEnv.con
   await cleanDB()
```
clauspeterhuebner: Does it make sense to insert in cleanDB the reset id auto increment for all entities/tables than to explicitly clear only a few ones afterwards?

clauspeterhuebner (approved):

PR #3614, opened 2026-04-05
feat(shared): add shared-native workspace and use it

einhornimmond:
### Overview

This PR introduces a native module (`shared-native`) for decay calculation and adds versioning support via `decayCalculationType`.

### Motivation

The decay calculation previously existed in two separate implementations:
- C++ (Gradido Node)
- Node.js (using Decimal.js)

Despite significant effort, both implementations produced diverging results over time.  
Some differences were small, others unexpectedly large — and without a consistent pattern.

This made it impossible to reliably compensate for or correct the deviation mathematically.

For a system like Gradido, which is intended to run across different platforms and implementations, this is a critical issue:
decay calculations must be **deterministic and reproducible everywhere**.

### Solution

The decay logic has been reimplemented in C and exposed via a Node.js native module (`shared-native`).

Instead of relying purely on floating point arithmetic, the implementation:

- Uses integer operations where possible (e.g. bit shifting for full years)
- Minimizes floating point usage to the fractional remainder
- Ensures that the most critical ranges (e.g. full-year decay steps) are exact

This significantly reduces numerical instability while preserving performance.

C was chosen deliberately, as it can be integrated across all relevant environments and serves as a portable, low-level reference implementation.

#### shared-native
- Introduces a native C implementation of the decay calculation
- Exposed via Node.js bindings
- Acts as the single source of truth for decay logic

#### decayCalculationType
- Added to:
  - `Transaction`
  - `PendingTransaction`
- Allows distinguishing between different decay implementations

### Why versioning is necessary

With the introduction of a new deterministic decay calculation:
- existing transactions must retain their original behavior
- new transactions must use the corrected implementation

Without explicit versioning:
- historical data could change meaning
- inconsistencies would become invisible
- debugging would become extremely difficult

### GitHub Worker

- Add linting test for shared-native
- Add scope validation for PR titles
- Extend file-change triggers (backend, dht-node, database, federation) to include shared-native
- Run shared-native tests in `test_submodules`
- Build shared-native before running turbo test/typecheck (required for CI stability)

### shared

- Use native version of `calculateDecay` in `logic/decay.ts`

### shared-native

- C library with reimplemented Gradido blockchain core logic
- Manually written C++ N-API wrapper
- Build environment-independent via zig-build (downloads and uses Zig as C/C++ compiler)
- Includes workarounds for GitHub Actions and Windows compatibility

### backend | dht-node | federation

- Update build setup to integrate shared-native
- Update Dockerfiles to include shared-native

### Final note

Floating point arithmetic was not removed — but constrained to a well-defined, minimal role.

The result is a decay calculation that behaves consistently across platforms,
with exact results in the most relevant ranges and controlled precision elsewhere.


shared-native/build.ts
```diff
@@ -0,0 +1,184 @@
...
+import AdmZip from 'adm-zip'
+import { family, MUSL } from 'detect-libc'
+import headers from 'node-api-headers'
+import { build, Target, type TargetTriple } from 'zig-build'
+
+async function isMusl(): Promise<boolean> {
+  return (await family()) === MUSL
+}
+
+const ZIG_VERSION = '0.10.1'
```
clauspeterhuebner, line 15: Should this version-number not be read from environment or package.json?

shared-native/package.json
```diff
@@ -0,0 +1,42 @@
...
+    "@biomejs/biome": "2.0.0",
+    "@types/adm-zip": "^0.5.8",
+    "adm-zip": "^0.5.16",
+    "node-api-headers": "^1.8.0",
+    "typescript": "^4.9.5"
+  },
+  "dependencies": {
+    "detect-libc": "^2.1.2",
+    "node-addon-api": "^8.7.0",
+    "zig-build": "0.3.0"
```
clauspeterhuebner, line 37: what's the difference between this zig-build version and the above mentioned zig-version in the build.ts file?

clauspeterhuebner (approved):
with two small questions

shared-native/build.ts
einhornimmond: This is for workarounds for zig-build which had the zig-version 0.10.1 hardcoded. It is only compatible with this zig version. zig-build work only on my linux out of the box. On windows and github worker the used exec command to unpack the downloaded tar containg zig compiler doesn't work. So I need to download it myself so that the zig-build code find it already there and didn't need to download.

shared-native/package.json
einhornimmond: the ZIG_VERSION is the version of the zig binary (https://ziglang.org/download/)
the version in package.json is the version of the npm packet. (https://www.npmjs.com/package/zig-build)

PR #3616, opened 2026-04-13
refactor(other): improve GradidoUnit, replace many Decimal with GradidoUnit

einhornimmond:
## Replace Decimal with GradidoUnit in backend, intern and graphql
Replace where it makes sense (Where it represent a Gradido Balance)
Skip Updating Entities Models for now

- update graphql queries
- update logic and data classes and functions
- add as GradidoUnit Scalar and validation for graphql
- add helper classes for easy handling of gradido balances (GradidoUnit, TemporalGradidoUnit, Duration)
- move Decay graphql Model from shared back into backend, put Decay initalizing logic into static functions in Decay instead of Transaction Constructor


einhornimmond (commented):
Read through changes, move some updates from following branches direct into this.

PR #3619, opened 2026-04-24
refactor(shared-native): update decay algorithmus

einhornimmond:
- update decay algorithm for using fixed point arithmetic
- restructure files to match c-library flair (now using include, src, bindings and third_party folders)
- export more functions to use in type script
- use custom from- and to- string functions written for special case of integer number represent a fixed point decimal number, optimized for accuracy and speed
- c files are now a direct copy from https://github.com/gradido/gradido-blockchain-core/tree/be9cb0d827089b35d4273b7befc291745812ea78


einhornimmond (commented):
Read through changes and add more comments and test shared-native with bun and also with nodejs on windows and linux runner.

PR #3620, opened 2026-04-25
feat(shared): add GradidoUnit and helper classes

einhornimmond:
Add classes for more easy Decay Management.
- GradidoUnit: Represent a Gradido Balance, home of new decay function
- TemporalGradidoUnit: GradidoUnit with date, for simpler calculation with Gradido Units, will auto decay before add or subtract other TemporalGradidoUnit and throw a error, if other TemporalGradidoUnit is earlier
- Duration: Simple Helper for managing duration, stored as seconds, but has static create functions for minutes, hours, days and date diff and a toString Function which will choose a good unit (seconds or minutes or hours or days)

- Update typescript target to ES2020 to make usage of bigint possible


einhornimmond (commented):
Read through changes and add more comments, move some updates from following branches direct into this.

PR #3670, opened 2026-07-04
fix(frontend): make the session-timeout modal renew and dismiss reliably

BerndHueckstaedt:
## Problem
When the session-timeout warning appears (75 s before the JWT expires), clicking **Angemeldet bleiben** (extend) does not reliably keep the user signed in: the modal briefly closes and immediately pops up again, and the user still ends up logged out. It could also stay stuck showing `00`. On top of that the modal showed a stray **OK / Cancel** footer that isn't ours.

Affects both production (1 h session) and staging (10 min).

## Cause
- The renew action used a **cache-first** `useLazyQuery` and only called `load()`, which runs once. It often skipped the network, so no fresh `token` header came back and the expiry never moved — the timer kept re-opening the modal.
- Modal visibility was set to `true` on every 1 s tick and **never reset**, so it reappeared right after the imperative hide (and stayed visible at `00`).
- The footer used the Vue 2 slot name `#modal-footer`, which bootstrap-vue-next ignores, so the component's **default OK/Cancel** footer rendered instead of the intended buttons.

## Fix
- Renew via a **`network-only` refetch** so a fresh token is issued and the Apollo afterware slides the expiry forward; recompute the remaining time immediately so the modal closes at once.
- Drive the modal **purely from the remaining time** (visible only inside the warning window), which also clears the "stuck at `00`" case.
- Use the correct **`#footer` slot** with exactly two buttons — extend (primary, gold) and logout — matching the app's other popups.

## Testing
Verified on the ki-playground staging instance (with a temporarily shortened warning window): the modal appears, **Angemeldet bleiben** keeps the session and the modal stays closed, **Abmelden** logs out, and no stray buttons appear.


frontend/src/components/SessionLogoutTimeout.vue
```diff
@@ -47,36 +39,39 @@ import { ref, computed, watch, onMounted, onUnmounted } from 'vue'
...
-  const diff = tokenExpirationTime.value - now
+  const diff = tokenExpirationTime.value - new Date()
   remainingTime.value = Math.max(0, Math.floor(diff / 1000))
-  // Show modal if remaining time is 75 seconds or less
-  if (remainingTime.value <= 75) {
-    sessionModalModel.value = true
-  }
-  // Clear interval if time expired
-  if (remainingTime.value <= 0) {
+  // Single source of truth: the modal is visible only inside the warning window.
```
einhornimmond: Which window? I don't understand. Please clarify the comment and make sure it is easy to understand that it refers to a time span.

frontend/src/components/SessionLogoutTimeout.vue
```diff
@@ -85,37 +80,38 @@ const formatTime = (seconds) => {
...
+const handleOk = async () => {
+  try {
+    const result = await renewSession()
+    if (!result) throw new Error('session renewal returned no result')
+    // Recompute at once so the modal closes as soon as the new expiry is known.
+    calculateRemainingTime()
+  } catch {
+    emit('logout')
   }
+}
```
einhornimmond, line 115: This code throw an error and catch it shortly after? What is the reason for throw an error at all, when no one will be ever seen it?
Please make sure the user get's at least a message (toast) which is understandable by non programmer.

einhornimmond (changes requested):
Looks code except for the two things I have commented.

frontend/src/components/SessionLogoutTimeout.vue
BerndHueckstaedt: Reworded so it clearly refers to the time span: the modal shows only while the remaining time is inside the warning span (the final `MODAL_WARNING_SECONDS` before expiry). Thanks!

frontend/src/components/SessionLogoutTimeout.vue
BerndHueckstaedt: Removed the throw-then-catch: `renewSession()` is now checked directly, no artificial Error. On failure the user gets a plain-language toast (`error.session-renewal-failed`, added to all 10 locales) before the logout, instead of a silent redirect. Added a spec assertion that the toast fires on failed renewal.

einhornimmond (approved):
ok

PR #3671, opened 2026-07-07
fix(frontend): boot-render fixes and brand colours for toasts and buttons

BerndHueckstaedt:
Both-mode changes pulled out of the dark-mode work so they can be reviewed and land on their own, ahead of the (purely additive) dark-mode PR that follows. Every change here affects light mode too.

## 1. Boot rendering (bug fixes)

- **Login-page flash:** the app mounted before the router's initial navigation had resolved, so the first render ran on the empty start location and `App.vue` picked the `AuthLayout` — flashing the login page (image slider, community name, coin) for a moment on every reload of an authenticated route, and when returning from the admin interface. Fixed by awaiting `router.isReady()` before mounting.
- **Boot language:** `vuex-persistedstate` rehydrates `state.language` without firing the `language` mutation that syncs the i18n locale — so without the incidental login-layout language switcher (removed by the fix above), the app showed the default language on reload until Settings was opened. The locale is now restored explicitly on boot, mirroring how the theme is applied.

## 2. Brand colours (both modes)

- Toasts use solid Gradido colours instead of the pale Bootstrap pastels: green success, red error, gold info (white text on green/red, dark text on gold, inverted close x).
- The success and link/QR-result confirmation close buttons use the gold `gradido` button variant, matching the other confirmation actions.

Verified against the ki-playground staging build.

### DIFF
--- a/frontend/src/assets/scss/gradido-template.scss
+++ b/frontend/src/assets/scss/gradido-template.scss
@@ -142,40 +142,35 @@ html > body {
   border-color: #f5c6cb;
 }
 
-.b-toast-danger .toast .toast-header {
-  color: #721c24;
-  background-color: rgb(248 215 218 / 85%);
-  border-bottom-color: rgb(245 198 203 / 85%);
-}
-
+// Toasts use the Gradido colours as solid fills in both light and dark mode,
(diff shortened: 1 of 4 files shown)


einhornimmond (approved):
Looks good and seems to work.

PR #3674, opened 2026-07-08
feat(frontend): device-local dark mode (system / light / dark)

BerndHueckstaedt:
## Device-local dark mode (system / light / dark)

Adds an opt-in dark theme, selectable in **Settings -> Appearance**:
- **Follow system setting** (default) - tracks the OS/browser light/dark preference live via `matchMedia`
- **Light**
- **Dark**

The choice is stored **per device** (localStorage, no backend/DB) and applied **before the first paint**, so there is no flash of the wrong theme on load.

### How it works
- **Store**: `themeMode` + `applyTheme` in the Vuex store; `main.js` dispatches `applyTheme` before mount. `App.vue` mirrors the `.dark-mode` class onto `html`/`body` (so teleported modals/toasts inherit it) and follows OS changes live.
- **No FOUC**: an inline script in `index.html` sets the dark class before the app mounts.
- **Colours**: semantic tokens in `_design-tokens.scss`; the dark theme is delivered as an **additive override sheet** `gradido-template-dark.scss` (`#app.dark-mode ...`). Light mode is left untouched.
- **i18n**: theme-switcher strings (`settings.theme.*`) in all 10 locales.

### Scope / how this was assembled
This is **PR 2** of the dark-mode work; it follows #3671 (both-modes render/colour fixes, already merged) and is **purely additive and light-mode preserving**.

The feature was developed on the `ki-playground` staging branch, which sits on top of the (not-yet-merged) GMS branch. To keep this PR clean it was **reconstructed onto master by intent** rather than raw cherry-picked - deliberately **excluding all GMS and Crea code** that live on adjacent branches. A few selectors in the dark sheet target the GMS "Matching" page that is not in main yet; they are inert no-ops until GMS lands.

### Testing
- Local CI parity green: stylelint, eslint, locales, vitest (675 passed), compile-scss, build.
- Live-tested on `ki-playground` across desktop + mobile, light + dark (pages, modals, toasts, login, mobile drawer).


frontend/index.html
```diff
@@ -52,6 +52,27 @@
...
+        if (parsed && parsed.themeMode) mode = parsed.themeMode
+      }
+      var osDark =
+        window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').matches
+      if (mode === 'dark' || (mode === 'system' && osDark)) {
+        document.body.classList.add('dark-mode')
+      }
+    } catch (e) {}
+  })()
+</script>
```
einhornimmond, line 78: @Claude Maybe you can speed this up, because it is in the hot-path and will block.
You parse the whole ~ 1KB json from local storage, maybe you can use a separate LocalStorage Key for themeMode to skip JSON parsing entirely.
And call matchMedia after it is clear that mode is system.

BerndHueckstaedt: Thanks, both good catches - fixed in 59b4c1ed8.

**No JSON parsing on the boot path any more.** `applyTheme` now mirrors `themeMode` into a dedicated key (`THEME_MODE_STORAGE_KEY = 'gradido-theme-mode'`), so the pre-paint script reads it with a single `getItem`. I put the write into `applyTheme` rather than the mutation on purpose: it runs on boot (`main.js`), on every theme change (Settings, logout) and on OS changes (`App.vue`), so the key always stays in sync while the mutations stay free of side effects.

**`matchMedia` is short-circuited** and only evaluated when the mode really is `'system'`.

```js
var mode = window.localStorage.getItem('gradido-theme-mode') || 'system'
if (
  mode === 'dark' ||
  (mode === 'system' &&
    window.matchMedia &&
    window.matchMedia('(prefers-color-scheme: dark)').matches)
) {
  document.body.classList.add('dark-mode')
}
```

No migration needed: dark mode is not released yet, so no one has a stored `themeMode`; and on the very first boot `applyTheme` seeds the key from the hydrated state anyway.

I also updated the branch with master. Local CI parity is green (eslint, stylelint, locales, vitest 676, build).

einhornimmond (approved):
As Step to get DarkMode without cleanup it is ok.

PR #3680, opened 2026-07-13
refactor(frontend): import Bootstrap once instead of components + full package

BerndHueckstaedt:
## What this does

`gradido.scss` imported the individual Bootstrap components (`alert`, `card`,
`modal`, …) **and then** the full `bootstrap` package, which already re-includes
all of them. Every Bootstrap rule was therefore emitted two or three times.

This PR drops the individual component imports plus `bootstrap-grid` and
`bootstrap-reboot` (all subsumed by the full package) and imports
`bootstrap/scss/bootstrap` once. It also fixes the stale "Bootstrap (4.5.3)"
header comments -- the project runs 5.3.3.

## Effect

- Compiled `gradido.css`: **537 KB -> 303 KB (-43 %)**.
- Shipped minified main CSS: **~415 KB -> 338 KB raw / ~56 KB -> 47 KB gzip**.

## Verified functionally identical

Compiling before/after and comparing the rule sets with selector order
normalised, the only differences are:

- one `box-sizing: border-box` declaration on `.row > *` that the standalone
  `bootstrap-grid` bundle added and that the global reboot rule
  `*, *::before, *::after { box-sizing: border-box }` already covers, and
- two duplicate license banners.

No rule is lost; rendering is unchanged in light and dark. `bun run build`,
stylelint and the full vitest suite pass.

## Note on merge order

This touches `gradido.scss`, which #3679 (native color mode) also touches in a
different place. Whichever merges second will need a trivial rebase.


einhornimmond (approved):
nice, 87.28 KByte less (duplicates) CSS (10.11 KByte compressed)

PR #3685, opened 2026-07-15
feat(backend): report the running commit at GET /api/version

BerndHueckstaedt:
## What

Adds a public GraphQL query `version { commit source }` that reports the commit the running backend was built from.

## Why

Backend-only deploys are invisible from the outside: the frontend/admin bundle hashes only change when *those* bundles rebuild, so there's no way to confirm from outside that a backend-only change actually went live. A plain `POST /graphql { version { commit source } }` now answers that.

## How

- `commit` is resolved once and cached: `git rev-parse HEAD` when the backend runs from a git checkout (bare-metal deploy), falling back to a build-time `BUILD_COMMIT` env var, then to `"unknown"`.
- `source` reports which of those was used (`git | env | unknown`).
- The query is public (no `@Authorized`) and exposes only the commit SHA — the repo is open source.


einhornimmond (changes requested):
@claude 
In the long run I like to shift to use graphql only where it make sense or nothing at all, so new especially static routes are added to backend/src/server/createServer.
Please add a /api/version route to backend/src/server/createServer.ts instead of the graphql Resolver VersionResolver.ts.
Add also /api to deployment/bare_metal/nginx/sites-available/gradido.conf.ssl.template and gradido.conf.template. Use /realms/gradido route as example.

esbuild is used to bundle the backend server code, please use this to bundle the commit version into this, for example like it is done in database/ with latestDbVersion instead of calling exec in backend.

Check also how docker-compose and docker handle the env forwarding in frontend for BUILD_COMMIT and BUILD_VERSION and mirror this for backend.

BerndHueckstaedt:
Reworked along your four points:

1. **Static route, not GraphQL** — `GET /api/version` is now a plain Express route in `server/createServer.ts`; `VersionResolver`, `BackendVersion` and the schema registration are gone.
2. **Baked in at build time** — an esbuild plugin (`backend/esbuild.config.ts`) rewrites the `./buildInfo` import in `server/version.ts` into a literal `const BUILD_INFO = { commit, source }`, resolved once at build time from `BUILD_COMMIT` (env) else `git rev-parse HEAD` else `"unknown"` — mirroring `database/` `latestDbVersion`. No `execSync` per request.
3. **nginx** — `location /api` added to `gradido.conf.template` and `gradido.conf.ssl.template`, modeled on `/realms/gradido`.
4. **Docker** — `backend/Dockerfile` takes `ARG BUILD_COMMIT` (mirrors `frontend/Dockerfile`) and `docker-compose.yml` forwards it as a build-arg to the backend service, so esbuild sees the running commit inside the image build.

The endpoint returns `{ commit, source }` (source = git | env | unknown, so a reader can tell how trustworthy the value is).

One caveat, same as the frontend: `BUILD_COMMIT` is not a declared input in `turbo.json`, so a cached `backend#build` could in principle bake a stale commit when only non-backend files changed. In practice a new commit changes source files, so the cache misses and it rebuilds. I left `turbo.json` untouched rather than change a repo-wide file for this — happy to add `BUILD_COMMIT` to the build task's env inputs if you'd prefer.

Merged current master in too. Ready for another look.

einhornimmond (approved):
nice work, I have only reduced unnecessary complexity.

PR #3688, opened 2026-07-16
feat(backend): add the Crea moderation assistant

BerndHueckstaedt:
Crea is an admin-side moderation assistant that helps moderators review community
contributions. For a contribution it gives a decision recommendation (confirm /
inquire) and drafts a warm reply, powered by the Claude API. The moderator always
decides — Crea only advises.

## Backend
- A new Anthropic client and a Crea service under `apis/anthropic/` — a rules-as-data
  ruleset (cached system prefix), deterministic post-processing (hour arithmetic,
  local salutation so the recipient's name never reaches the API), pseudonymous
  evaluation records, and runtime model/effort settings.
- `CreaResolver` (evaluate / rewrite / batch / settings), guarded by the existing
  `AI_SEND_MESSAGE` and `COMMUNITY_UPDATE` rights.
- Config `ANTHROPIC_ACTIVE` / `ANTHROPIC_API_KEY` / `ANTHROPIC_MODEL` and an opt-in
  `CREA_STUB` preview (off by default in production).
- New dependency `@anthropic-ai/sdk`.

## Database
- Two additive migrations: `crea_records` (pseudonymous per-activity rows that become
  the participant history) and `crea_settings` (single-row model/effort).
- Nullable `users.gender` / `salutation` / `crea_signature` columns (moderator-curated
  salutation/signature, E-013). Additive, no existing data touched.

## Admin
- A Crea evaluation modal, opened from a per-row button in the creation lists; a Crea
  settings page (model / effort, admin-only).
- One-click insertion of Crea's reply draft into the message field, and a public memo
  supplement; a Cmd/Ctrl+B bold shortcut for moderator messages.
- `crea.*` locale keys in de/en, with an English fallback in the other eight locales
  (a native-speaker translation pass follows separately).

## Rollout / safety
- Crea stays **dormant** until `ANTHROPIC_API_KEY` is set and `ANTHROPIC_ACTIVE=true`;
  without a key the resolver reports "not enabled" and the UI shows a calm hint.
- Admin-only. No frontend/wallet changes.

## Verification
- Backend `tsc` + `biome` green locally; jest + docker build run on CI.
- Admin `eslint` + `stylelint` + `locales` + `vitest` (329 tests) + `build` green locally.
- The two migrations run against the test DB on CI.

---

## Hinweis für Dario (Koordination, kein Review-Blocker)

Hallo Dario,

das ist der dritte und letzte Crea-PR — dieselbe Crea, die seit Tagen auf dem
ki-playground-Staging läuft und von uns abgenommen ist, hier sauber aus `master`
rekonstruiert (kein Fork-Cherry-Pick). **Das proprietäre GMS ist bewusst komplett
draußen** — geteilte Kern-Dateien (`User`-Entity, `schema.ts`, `entity/index.ts`,
`UserResolver`, `RIGHTS`, `config`) sind hunk-genau nur um die Crea-Teile ergänzt.

Drei Punkte für die Infra, alle unkritisch:
- **Neue Dependency** `@anthropic-ai/sdk` (nur `bun.lock` + `backend/package.json`).
- **Zwei Migrationen** (`crea_records`, `crea_settings`) laufen beim Deploy automatisch
  gegen die Prod-DB — nur zur Vorwarnung, rein additiv.
- **Prod-`ANTHROPIC_API_KEY`** setzen (+ `ANTHROPIC_ACTIVE=true`), damit Crea aktiv wird.
  Der Key **existiert schon** (Du hast ihn auf ki-playground) — denselben wiederverwenden
  oder einen neuen holen. Ohne Key bleibt Crea schlicht schlafend. **Die Kosten trägt die
  Gradido-Akademie / Bernd** — nichts abzuwägen.

**Merge-Reihenfolge:** dieser PR teilt Dateien mit #3685 (`schema.ts`) und #3686
(`CreationConfirm.vue`, `OpenCreationsTable.vue`). Bitte **#3685 → #3686 → diesen PR**
mergen. Ich habe lokal einen Probe-Merge auf `master + #3685 + #3686` gemacht: alles
auto-merged bis auf **zwei triviale „beide-behalten"-Stellen** (je ein `emits`-Eintrag in
`OpenCreationsTable.vue` und `CreationConfirm.vue`). Sobald die zwei drin sind, aktualisiere
ich den Branch und löse die zwei Stellen auf.

Falls Du den PR lieber kleiner hättest, kann ich Backend und Admin trennen — sag Bescheid.


einhornimmond (changes requested):
Please wait until I have enough time to check this.

BerndHueckstaedt:
Status update — no action needed, just so this does not sit blocked behind the wrong thing.

**#3686 is merged**, and I have merged master into this branch, so it is up to date and CI is green.

The two places where Crea and the bulk resubmission feature touch the same lines are resolved by keeping both: the `crea-evaluate` / `resubmission-saved` emits in `OpenCreationsTable.vue`, and in `CreationConfirm.vue` the matching handler plus the Crea modal alongside the bulk modal. The two features are independent and now coexist — the admin suite exercises them together (337 tests).

**On the ordering:** this PR only ever depended on #3685 through `backend/src/graphql/schema.ts`, where both registered a resolver. Since your review asks for `/api/version` as a static route in `server/createServer.ts` instead of the GraphQL `VersionResolver`, #3685 will not touch `schema.ts` any more — so this PR is no longer coupled to it and can be reviewed and merged on its own, whenever you find the time.

I am doing the #3685 rework along your four points separately.

backend/src/apis/anthropic/crea/deterministics.ts
```diff
@@ -0,0 +1,146 @@
...
+  if (gender === 'male') {
+    return { salutation: `Lieber ${name}`, uncertain: false }
+  }
+  if (gender === 'female') {
+    return { salutation: `Liebe ${name}`, uncertain: false }
+  }
+  // Name known, gender not: keep the name, default to the neutral "Hallo", flag
+  // for review — never a risky "Liebe/Lieber" guess (E-014).
+  return { salutation: `Hallo ${name}`, uncertain: true }
+}
```
einhornimmond, line 95: Only german hardcoded salutation?
Please use i18n for that.

einhornimmond (changes requested):
Think (and discuss with Bernd) about using i18n for system prompt injections or db so admin can edit it freely in admin interface

Please add design documents in docu/crea (currently not exist) which contain all the Es like E-020 and othere short references in comments

For next time:
Please use english variable names, english in comments and english for flags.
Please write new Code with DrizzleOrm instead of TypeOrm as long you don't need the existing TypeOrm Entities.

BerndHueckstaedt:
Thanks — docs added and your two changes merged in.

**Design documents:** `docu/crea/README.md` covers the architecture (the three layers: behaviour, creation rules, deterministic code), the privacy design and a code map; `docu/crea/decisions.md` resolves every `E-xxx` a comment references, plus the `DO-x` build steps.

**On i18n vs. db for the prompt (your first point):** this is the direction the design was already built for. The rules are deliberately kept as a named data document behind a seam from Crea's behaviour (E-002), precisely so that "admins edit the rules in the UI" becomes wiring rather than a rewrite.

I would split your two options along that seam:
- the **creation rules** (what counts as a contribution to the common good) belong in the **db** — they are per-community *content*, not translations of one canonical text, so i18n would be the wrong shape;
- the **behaviour spec** (voice, format, procedure) should stay in code and *not* be admin-editable, otherwise an admin editing rules could break the output format. That layer is where your i18n TODO fits: translating it would stop non-German moderators' models from translating on the fly.

**One adjustment to your commit — please say if you disagree.** I kept the dedicated `AI_SETTINGS` right (clearly better than borrowing `COMMUNITY_UPDATE`), but moved it from `MODERATOR_AI_RIGHTS` to `ADMIN_RIGHTS`. Switching the model and effort applies globally to every moderator at once and drives both API cost and answer quality, so we would like it to stay with the administrator — moderators inherit the effect, not the control. That was the intent behind E-028, and the academy carries the API cost.

**On your last two points ("for next time"):** understood, and agreed — the German flags and identifiers do violate our own convention. One honest caveat: `anrede_unsicher`, `beitragRef` and `stunden_geschaetzt` sit in the prompt/output-schema contract and in persisted records, so renaming them is a behaviour-version bump rather than a rename. Happy to do it as its own change. Can this PR merge on that basis?

BerndHueckstaedt:
One more thought on the editable-rules direction, so it does not get wired in by accident.

When the creation rules become editable, it would be good if the editing right were **its own right**, not the admin right. The two are different in kind:

- switching the model/effort is a **budget** decision with a global blast radius — that is what `AI_SETTINGS` now guards, and an admin is the right home for it;
- editing the creation rules is a **domain** judgement: whether a rule matches how contributions are actually assessed in practice. An administrator is usually not the person who can judge that — the moderator doing the daily work is.

With per-community rules this stops being cosmetic: every community needs someone who tends its own rules, and by construction that is not a global administrator.

So the cheap seam would be a second right (say `AI_RULES`) beside `AI_SETTINGS`. Who holds it can then be decided separately — an existing role, a dedicated "Crea curator" tier, or per person — without pinning the architecture down now. Nothing to do for this PR; just flagging it before the rules move into the database.

It also fits the `// TODO - load from database dynamically` on the role resolution: once roles come from the db, adding such a tier is configuration rather than code.

einhornimmond (approved):
seems to be ok now.

PR #3689, opened 2026-07-16
fix(workflow): make clear more platform independent

einhornimmond:
- use npm modul rimraf instead or rm -rf for platform independent clear workspace
- add node_modules from root folder to be cleared by turbo clear / bun clear
- use bun for calling sass instead of node to circumvent problems with chokidar


clauspeterhuebner (approved):

clauspeterhuebner:
ja der Build funktioniert wieder, super danke für deine schnelle Hilfe...

PR #3690, opened 2026-07-22
refactor(frontend): draw the separators as a line and re-enable no-missing-keys

BerndHueckstaedt:
Five places asked the translation function for a `"|"` to print a separator. There is no such entry in the language files, so vue-i18n fell back to printing the key — it produced the pipe **by accident**. It was written that way to get past `no-raw-text`, and it is the reason `no-missing-keys` had to be switched off for `.vue` files.

Two rules were standing in each other's way, and the one that lost is the one that catches a real class of bug: a `$t()` pointing at a key that does not exist renders the raw key name to the user.

## What changes

The separator is decoration, so it is drawn as one: a border on the entry that follows, via a `.separator-start` class.

- `currentColor` makes it take the colour of whatever text surrounds it, so **one** class fits the light footer and the dark font-size popover with no value to keep in sync. Bootstrap's own `border-start` was not enough — its `#e9ecef` is close to invisible on white.
- A line stays **out of the accessibility tree**. The old pipe was announced between every menu entry; a border is not. This is also why it is a border and not generated content — screen readers do read `::before` content.
- It **cannot dangle**: it belongs to the entry it precedes, so the hand-managed visibility classes on the separator span in the auth layout fall away.

With the five calls gone, `no-missing-keys` reports nothing and goes back to `'error'`.

## Checked

- The rule was verified in both directions: clean on the current tree, and **failing** against a deliberately broken key reference — so it is a net, not a formality.
- `frontend/src/assets/css/gradido.css` is generated and not versioned. Confirmed that `frontend#build` depends on `compile-scss` in `turbo.json`, and that `.separator-start` really lands in the built stylesheet — without that step the separators would have disappeared entirely, including on the login screen.
- Local: eslint clean (0 problems), stylelint clean, locales sorted, vitest 686 passed, build green.

This has been running on the Akademie staging instance since this morning.

@einhornimmond — as agreed.

### DIFF
--- a/frontend/.eslintrc.js
+++ b/frontend/.eslintrc.js
@@ -59,7 +59,7 @@ module.exports = {
     'vue/v-on-event-hyphenation': 0, // TODO remove at the end of migration and fix
     'vue/require-default-prop': 0, // TODO remove at the end of migration and fix
     'vue/no-computed-properties-in-data': 0, // TODO remove at the end of migration and fix
-    '@intlify/vue-i18n/no-missing-keys': 0, // TODO remove at the end of migration and fix
+    '@intlify/vue-i18n/no-missing-keys': 'error',
     '@intlify/vue-i18n/no-unused-keys': [
       'error',
       {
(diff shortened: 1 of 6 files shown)


einhornimmond (approved):
Thank you, at last it was fixed.

PR #3693, opened 2026-07-25
feat(backend): offer Claude Opus 5 and a fast-mode switch for Crea

BerndHueckstaedt:
Two changes to the Crea moderation assistant's runtime settings.

## Claude Opus 5 in the model presets

Replaces Opus 4.8 with Opus 5 in the dropdown, and Opus 4.7 with Opus 4.8. The
model field itself stays free text, so any model id keeps working and a new model
needs no release — the dropdown only fills the field.

## A fast-mode switch

Fast mode runs the same model with faster output at premium pricing. New admin
checkbox, stored in `crea_settings`, off by default.

Requesting it needs three things together: the beta messages endpoint, the beta
flag `fast-mode-2026-02-01`, and `speed: "fast"` as a top-level parameter. It is
only available on some models (Opus tier), and it has its own rate limit separate
from the standard one.

Rather than pin a model list in code — which would defeat the free-text field —
the client asks for fast mode and retries the same request at normal speed when
the API rejects it: a 400 naming speed/fast, or a 429 from the separate fast-mode
rate limit. Every other error still surfaces, so a real failure is never hidden
behind a silent second call.

The "test model" button reports a downgrade instead of hiding it, so an admin
learns before saving that a model answers normally but not in fast mode. The
second commit makes that report honest: it previously claimed "not available for
this model" for every refusal, including a plain rate limit, which would send an
admin off changing the model to fix something the model is not responsible for.
It now names a 429 as "busy right now" and otherwise passes through the API's own
wording. That fix came straight out of testing — the first refusal we saw on
staging was the rate limit, not the model.

## Notes

- One additive migration, `0107-crea_settings_add_fast_mode`: a single
  `tinyint(1) NOT NULL DEFAULT 0` column. Nothing changes until an admin ticks
  the box.
- Fast mode roughly doubles the token price while it is on. At Crea's fractions
  of a cent per answer that stays small, but it is a deliberate choice, which is
  why it is a switch rather than a default.
- Switching speed invalidates the prompt cache once — Crea's rule prefix is
  rebuilt on the first answer after the change, then behaves as before.
- Verified on our staging instance: the settings page, the model test including
  the fast-mode probe, and the fallback path.


backend/src/apis/anthropic/AnthropicClient.ts
```diff
@@ -27,8 +27,47 @@ const logger = getLogger(`${LOG4JS_BASE_CATEGORY_NAME}.apis.anthropic.AnthropicC
...
+ * model to fix something the model is not responsible for.
+ */
+function fastModeRefusalNote(error: unknown): string {
+  if (error instanceof Anthropic.RateLimitError) {
+    return ' (schneller Modus gerade ausgelastet - Crea antwortet normal)'
+  }
+  const detail = error instanceof Error ? error.message : String(error)
+  return ` (schneller Modus abgelehnt: ${detail} - Crea antwortet normal)`
+}
+
```
einhornimmond: use english or locals (in core) for string error messages

einhornimmond (changes requested):

einhornimmond (approved):

PR #3695, opened 2026-07-25
feat(backend): group tags — submission field, moderator scope and the groups page

BerndHueckstaedt:
# Group tags

Contributions can belong to a group, and a moderator can be scoped to the groups
they look after. This replaces the fragile "inline #hashtag in the memo"
convention with a structured, administrator-managed list.

The feature has been running on our staging system (ki-playground) for a week and
has been used there to set up real groups.

## What it does

**For a member (wallet)**
- the submission form carries a group field, pre-filled from their own last
  choice, derived from their history
- both contribution tabs filter by group and show the group of a contribution
- the community info page lists the moderators grouped by the groups they look
  after

**For a moderator (admin)**
- a "Groups" page to create and rename groups. The name is what is shown; the tag
  is a short slug suggested from the name
- the contribution list filters by group, and the group of a contribution can be
  changed while it is still open or in progress
- a moderator can be assigned a visibility scope: they then see and act on only
  the contributions of their groups. This holds for `MODERATOR` and
  `MODERATOR_AI` alike

## How it is built

Four commits, one per package, each self-contained:

| Commit | Package |
|---|---|
| `feat(database)` | tables, entities, migrations 0107 and 0108 |
| `feat(backend)` | resolvers, rights, the scope, the filter |
| `feat(admin)` | Groups page, filter, role tab |
| `feat(frontend)` | submission field, lists, info page |

**Design decisions worth knowing about**

- **The scope is a real boundary, not a filter.** One predicate
  (`buildModeratorScopePredicate`) is shared by the contribution list and by
  `assertContributionInModeratorScope`, which guards every action taken by
  contribution id. Enforcement and display therefore cannot drift apart.
- **An assigned group overrides any hashtag in the memo**, decided by
  `contributions.group_tags_set_at` (migration 0108). A contribution from before
  the group field keeps its inline tag; once the field has been used — even to
  say "no group" — a `#word` in the text is ordinary text again.
- **"No group" asks the canonical list.** Only a `#` that names an existing group
  counts, so a stray `#thanks` does not hide a contribution from every list.
- **The wallet keeps its own filter argument type.** The admin search matches the
  e-mail address; the wallet search must never do that, and a test asserts zero
  hits when searching for an address.
- **An unassigned scope means "sees everything".** An empty scope is stored as
  `NULL` and means no restriction; there is no "sees nothing" state.
- **Both migrations are additive** and touch no existing data.

## Two things we know about and left as they are

**The legacy inline-hashtag rule is approximate, on purpose.** Before the group
field existed, a `#word` in the memo was the only way to name a group, and that
stock is still resolved for contributions the field was never used on. The
resolution is not exact: the SQL side matches the hashtag as a substring while
the display side tokenises it, and the two compare the memo in different
collations. So with two groups where one slug is a prefix of the other, a legacy
contribution can answer to the wrong group's filter; an accented slug is resolved
by SQL but not by the display path; and renaming a group drops its legacy
memberships.

We decided not to chase this. Those contributions are history — already confirmed
or denied, and the wallet's community list only reaches back six months anyway,
so most of them are no longer on display at all. A legacy contribution showing up
under "no group" instead of its group costs nothing. What we did fix is the one
path that kept *producing* new contributions in that state: a contribution filed
by a moderator through the admin now carries the "no group" stamp, so a `#word`
in the moderator's own wording cannot choose a group.

**No event records on the group writes.** Three of the new writes follow an
existing pattern but omit its event:

- `setModeratorGroupScope` writes the scope into another user's role record; its
  model `setUserRole` writes `EVENT_ADMIN_USER_ROLE_SET`
- `assignContributionGroupTags` changes the group of another member's
  contribution; its sibling `adminUpdateContribution` writes
  `ADMIN_CONTRIBUTION_UPDATE`
- `createGroupTag` / `updateGroupTag` have their own right but no event

Adding them means three new event types plus wiring, which is behaviour that has
not been through our staging system. Happy to add it here if you would rather
have it in one go.

## Checks run locally

- `database`: typecheck, lint
- `backend`: typecheck, lint
- `admin`: lint, stylelint, locales, 383 tests, build
- `frontend`: lint, stylelint, locales, 701 tests, build

The eight new backend test files all need the test database, so they run in CI.

10 languages in the locale files. The non-DE/EN values are our own translations
and would benefit from a native-speaker pass, which is pending for the locale
files generally.


BerndHueckstaedt:
Hi Dario,

heads-up before you review: **migration 0109 does nothing on a real upgrade, so we are taking it back out.**

0108 creates `group_tags` empty, 0109 runs right after it, finds no canonical tags and returns. A migration runs once, so the conversion of the existing inline hashtags never happens — and since the same release removes the inline resolution path, those memos stay ordinary text for good. It cannot be repaired inside a migration either: the table itself only comes into existence with this release, so no group can exist at the moment 0109 runs.

A follow-up commit will drop 0109 with its two tests and correct the comments that refer to it. Nothing else changes — the feature does not depend on the conversion. So there is no need to review 0109 as it stands.

The replacement comes as its own PR: on a group in the admin, show how many older contributions mention its tag and let an admin adopt them deliberately. Same guard the migration had (only contributions with no link row and no `group_tags_set_at`) and the same display rule, so nothing moves into a group it was never shown in. That is also the first moment the canonical list exists at all.

— Claude

BerndHueckstaedt:
Following up on my note above: rather than patching the migration out of this branch, we took your earlier suggestion and set the whole thing up again from scratch.

**Superseded by #3699.** Please review that one instead.

The reason is the history here, not the content. This branch had grown into a
build-it-then-take-it-back-out arc — four commits building the feature with the inline
hashtag path, then a commit converting it, one ripping the read path out, and one
removing the migration again — plus a master merge in the middle. Since PRs here land
as merge commits, all of that would have stayed on master commit by commit, and a
later reader would have had to work out which state actually won.

#3699 is four commits on current master, one per package, describing what is there
rather than how we got there.

**The content is unchanged, and that is checkable rather than a claim:** both branches
point at tree `4a163c772ce58f987d11ec9d78c135a22b9c7cd2`, so `git diff group_tags
group_tags_v2` is empty. CI agrees — 27 pass / 14 skipping on both.

Nothing was reviewed here yet, so nothing is lost by closing it. Sorry for the extra
notification.

— Claude

PR #3711, opened 2026-08-06
feat(backend): switch optional modules on and off from the admin panel

BerndHueckstaedt:
A community can switch an optional module on or off itself, in the admin panel, without
a deploy and without anyone touching the server config. First module: matching, whose
backend and wallet follow in their own deliveries. It ships **off**.

The switch lives in a new single-row `module_settings` table (Drizzle, migration `0111`),
is read and written through `database/src/queries/moduleSettings.ts`, and is flipped on a
new admin page under **Modules** — which also absorbed the Crea settings page, so
`/creaSettings` now redirects there.

## The switch subtracts rights from the role, rather than being a second check

The obvious shape would be a check inside the authorization directive. That covers
`@Authorized` and nothing else, and it is not where this backend asks about rights: all
eight field resolvers call `context.role.hasRight()` directly, and the salutation guard
presents itself as the pattern to copy. A matching field hung off the shared `User` type
the same way would have answered yes with the module off.

So a switched-off module has its rights taken **off the role** once the role is resolved.
Every caller inherits it — the directive's own check, every `hasRight()` in a resolver,
and anything written later. Administrators included, since `ROLE_ADMIN` inherits
`USER_RIGHTS`: off means off, not off-for-most.

The cost is one read per request, memoized on the context. Not cached beyond the request,
so a flip is visible on the very next one.

**No row means every module is off.** The migration deliberately writes none, so a fresh
database and an upgraded one start from the same safe state and the first run has nothing
it could get wrong. The admin panel writes the row on the first save.

## Four matching rights arrive with no resolver behind them

They come with the gate rather than with the matching backend: the gate can neither be
written nor tested without them. They sit in `USER_RIGHTS` because every member may keep
an entry of their own — and while the module is off they are withdrawn from every role,
so nothing can reach them. `LIST_ACTIVE_MODULES` is separate on purpose and belongs to no
module: a switched-off module has to be able to report that it is off.

## Where the gate ends

It governs the backend service on the `@Authorized` path and every `hasRight()` behind it.
Federation builds its own schema with no auth checker; webhooks, the standalone export
scripts and dlt-connector never resolve a role at all. A module with a surface outside
those needs its own check there, and `data/Module.logic.ts` says so where a reader will
find it.

The old GMS people search is a separate feature this switch deliberately leaves alone —
what leaves for gms.gradido.net is server configuration and a different decision. Since a
page called "Modules" reads as if it governed both, the GMS section now says so.

## What the review changed

The branch was reviewed twice. The first round rewrote the design (the withdrawal above
replaced a separate gate); the second round, against the rewritten code, produced fifteen
findings, of which fourteen were real. The ones worth naming here:

- **The module read sat between assigning the role and checking it.** One context serves
  every field of an operation, so across that await the un-narrowed role was readable by
  whatever else was resolving — the very thing the withdrawal exists to prevent. The read
  now happens before any role is published, and the withdrawal is synchronous again.
- **That read had no guard**, while the user lookup twenty lines above has one. An
  exhausted pool sent mysql2's own message, failing statement included, to the client for
  every authorized request. It now refuses with our own message.
- **The memo cached the value, not the promise**, so the root fields of one operation all
  missed it and each issued its own query — and could disagree with each other across a
  flip.
- **`dbUpsertModuleSettings` decided success from `result.at(0)`**, which for a drizzle
  insert is always the truthy `ResultSetHeader`. The failure branch was unreachable and a
  real write failure escaped the `Result` contract as raw SQL in an admin toast.
- **The Crea half of the merged page could save a form it had not read.** The module half
  had that guard; the older Crea half, now thirty lines below it, did not.
- **Two comments stated the opposite of the mechanism** — one claiming `hasRight()` was
  not covered, one mislabelling `LIST_ACTIVE_MODULES` as a matching right, which if
  followed would refuse the wallet the answer exactly when matching is off.

## Tests

`isAuthorized.test.ts` had a real gap: every case ran as a plain member with an empty
`userRoles`, so the role switch never executed and the claim this design rests on — off
means off for administrators too — had no coverage at all. "Switched off" was also only
ever fed as a *missing* row, never as the stored `0` every instance holds once an admin
has unchecked the box.

Both are fixed, and each new test was measured rather than assumed: the four faults were
put back in one at a time, and each fells exactly the test written for it. The stored-zero
fault fells three that the old fixture could not have seen.

`database/src/queries/moduleSettings.test.ts` needs a real database and runs for the first
time in CI.

## Left open on purpose

- No event is written when a switch is flipped, so it cannot be traced later.
  `setCreaSettings` writes none either, but neither of those withdraws a right from every
  member.
- `crea_settings` stays on TypeORM while `module_settings` is Drizzle. Two ORMs side by
  side is the state `AGENTS.md` describes, not a deviation from it — but the two settings
  tables are near-identical singletons and will want merging once the migration reaches
  the older one.
- One flag per module means one **column** per module, so a second module needs a
  migration. Fine for the second; worth revisiting before the third.

## Summary by CodeRabbit

- **New Features**
  - Added a Modules administration page for enabling or disabling matching functionality.
  - Added read-only GMS availability status and localized save, loading, error, and unavailable-state messages.
  - Added backend support for module availability and permission enforcement.
  - Existing Crea settings links now redirect to the Modules page.

- **Localization**
  - Added translations for the Modules section across supported languages.

- **Documentation**
  - Updated the code map to reference the new Modules settings page.


BerndHueckstaedt:
Closing this deliberately, not parking it. The requirement it answers turned out to be
much smaller than what it grew into, and the reasoning is worth leaving behind.

**What was actually needed:** matching should be mergeable into master while it is
unfinished, and invisible to users until it is done — then switched on once, for everyone.

**What this PR became:** a per-instance module system — a settings table, a migration, an
admin page, a rights-withdrawal mechanism in the authorization directive, and a
process-level cache to keep that mechanism off the request path.

The step that produced the gap was a finding phrased as a security hole: the current
switch covers only the frontend, and the four matching rights sit with every logged-in
member, so anyone could call the resolvers with the switch off. That is true. The question
nobody asked is **what they would reach** — and the answer is: `listMatchingEntries`
filters on `userId: user.id`, so a crafted request reaches nothing but the caller's own
entry in an unfinished feature. No foreign data, no leak. The exposure never justified the
mechanism.

**What replaces it, and it is smaller than this PR's first commit:** matching is not on
master at all today, so there is nothing to hide yet. The rights arrive with the matching
backend. The wallet delivery brings its pages behind the build-time flag that already
exists, defaulting to off and properly substitutable — the one line currently pinned to
`true` is the whole fix. Switching it on is one environment variable and one deploy.

A community that does not want matching disables it the same way.

**Not lost:** this branch stays. If per-instance module switching is ever wanted as a
feature communities choose, the shape here is a starting point — with the caveat already
recorded in the description, that one column per module carries the second module and not
the third.

**Carried over separately:** the admin Crea settings page can save a form it has not read
yet — a bug that exists on master independently of any of this. It follows in its own
pull request.

One thing this PR did earn: the branch was CI-red for eight hours before anyone noticed,
because the module read put a second connection pool on the authorization path and three
timer-faking tests hung on it. That trap is now written down where it will be read.

